PR #4, opened 2016-02-21
FIX: Check for fillOpacity when fill is undefined

jagan-veeraraghavan:
Fix for condition where both fill and fillOpacity properties are
undefined


magicismight:
Thanks for this PR. :+1:

PR #15, opened 2016-04-11
get rid of getChildren arrow function

gilesbradshaw:
this doesn't work the same in arrow functions - I getting this.props = undefined


gilesbradshaw:
Ignore me - i needed babel stage 0 reset..

PR #21, opened 2016-04-23
Bring back missing patternReg.js and refactor all regular expressions…

jcsmesquita:
Missing patternReg.js and refactor patterns.js to import from there as well.


magicismight:
Thx for this PR
:+1:

PR #22, opened 2016-04-23
Switches RNSVGPath.m and RNSVGText.m strokeDash to new strokeDasharray

duhduhdan:
A few iOS files weren't switched over to the new `strokeDasharray`


duhduhdan:
Also possible the Android equivalents weren't addressed either, but I haven't checked.

magicismight:
:+1: 
Thanks for this PR

PR #26, opened 2016-04-27
Remove the postinstall script that runs rnpm.

gijoehosaphat:
In our build scripts, `rnpm` does not 100% accurately insert packages properly. The combination of other dependencies confuses `rnpm` which ultimately causes conflicts and build errors. This is the case for our application, but I suspect it also applies to some other people as well.

This PR is just to remove the postinstall script so the developer can install manually, or via running rnpm themselves.

### DIFF
--- a/package.json
+++ b/package.json
@@ -22,9 +22,6 @@
         "svgpath": "^2.1.5",
         "lodash": "^4.0.0"
     },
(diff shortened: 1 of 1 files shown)


magicismight:
Fair enough
Thanks for this PR.

Issue #35, opened 2016-05-09
React-native 0.25 is available, could the code be updated to use this version


pietgk:
Hello,

how is the migration to newer released (like 0.25) planned?

magicismight:
Done.

pietgk:
I am using it and the warnings on react / react-native or gone for me.
as far as i can tell its working for me.
thx for the update.

magicismight:
I think this have been fixed in 2.0.0
Everthing works fine on 0.25.1 and 0.26.0.
So I`m closing this.

PR #39, opened 2016-05-12
React API must be now required from react package

chunghe:
#35 with the pull request, should be able to upgrade react-native to version 0.25.0. Tested on both iOS and Android.


magicismight:
Thx for this PR
:+1:

PR #45, opened 2016-05-19
Fix createReactNativeComponentClass import for RN 0.26.0

marcshilling:
This broke again in React Native 0.26.0.

`createReactNativeComponentClass` is no longer under `react-native/Libraries/ReactNative/`, it's under `react/lib/`.

### DIFF
--- a/elements/G.js
+++ b/elements/G.js
@@ -1,5 +1,5 @@
 import React, {Component, PropTypes} from 'react';
-import createReactNativeComponentClass from 'react-native/Libraries/ReactNative/createReactNativeComponentClass';
+import createReactNativeComponentClass from 'react/lib/createReactNativeComponentClass';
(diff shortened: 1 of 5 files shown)


magicismight:
I have fixed this in the next release.
No longer use createReactNativeComponentClass any more.
`import createReactNativeComponentClass from react/lib/createReactNativeComponentClass`
won`t work in 0.25.1 and 0.24.0.
Got anthor workaround to do this instead.

magicismight:
I`m closing this because the react-native-svg@2.0.0 is released just now.
I have found another way to work this out, no longer use createReactNativeComponentClass anymore.
The workaround in the latest code is not a pretty one, but will work both on react-native@0.26.0 and below.

magicismight:
Thanks for this PR anyway.
:+1:

PR #86, opened 2016-07-24
SVG images double rendering on Android

silverspace:
Fixing RNSVGImageShadowNode to no longer dangerously store a reference
to a DataSource result, since it's only valid for the lifetime of the
onNewResultImpl() method. Instead, the images are rendered from the
Fresco image bitmap cache.

Updating RNSVGVirtualNode to call getScreenDisplayMetrics() instead of
getWindowDisplayMetrics(), since the latter is deprecated.

Synchronizing access to RNSVGSvgViewShadowNode.drawChildren(), since
it can be called by multiple threads simultaneously (the native UI
thread and the asynchronous Image loader callback).

### DIFF
--- a/android/src/main/java/com/horcrux/svg/RNSVGImageShadowNode.java
+++ b/android/src/main/java/com/horcrux/svg/RNSVGImageShadowNode.java
@@ -16,21 +16,22 @@ import android.graphics.Paint;
 import android.graphics.PorterDuff;
 import android.graphics.Rect;
 import android.net.Uri;
-import android.util.Log;
 
-import com.facebook.common.executors.CallerThreadExecutor;
+import com.facebook.common.executors.UiThreadImmediateExecutorService;
+import com.facebook.common.logging.FLog;
 import com.facebook.common.references.CloseableReference;
 import com.facebook.datasource.DataSource;
-import com.facebook.common.logging.FLog;
 import com.facebook.drawee.backends.pipeline.Fresco;
-import com.facebook.imagepipeline.core.ImagePipeline;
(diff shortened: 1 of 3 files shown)


magicismight:
Thanks fo this PR!
:+1: 
I`ll get this shiped within next version

magicismight:
But there is a problem with this PR.
The first time to render the Image soemtimes it won`t work.
It seems like the `markUpdateSeen()` didn`t work properly.
qq 20160726181354

silverspace:
Thanks for letting me know, I'll take a look tonight. I suspect this is related to bitmap caching behavior in Fresco. 

One possibility is that there are caching differences between Android 4.x and 5.0:  http://frescolib.org/docs/caching.html

----
"On Android 4.x and lower, the bitmap cache’s data lives in the ashmem heap, not in the Java heap. This means that images don’t force extra runs of the garbage collector, slowing down your app.

Android 5.0 has much improved memory management than earlier versions, so it is safer to leave the bitmap cache on the Java heap."
----

Did you see this blank image behavior on an Android 4.x system? Or after backgrounding the app and re-launching?

magicismight:
I`m running it on Android 4.4.4 API19
I got it compiled and launched from android studio sometimes the Image wont`t render.
Unless I have triggered the state change of the virtual DOM in react.

magicismight:
Reloading the project from the react-native tool works fine.
It only happened when the APP relaunched.

silverspace:
FYI: this should be fixed under https://github.com/magicismight/react-native-svg/pull/94

PR #94, opened 2016-07-29
Android: SVG Images not rendering the first time a view is displayed

silverspace:
Since SVG images are loaded asynchronously, we need to invalidate the
view after drawing to the canvas so that the view will be re-rendered.

Storing a reference to the actual SvgView in ViewManager, and passing a
reference to the ViewManager to each shadow node, so that shadow nodes
can invalidate the root SVG view.

(Also fixing a NPE caused by getDefinedBrush() returning null).

### DIFF
--- a/android/src/main/java/com/horcrux/svg/RNSVGImageShadowNode.java
+++ b/android/src/main/java/com/horcrux/svg/RNSVGImageShadowNode.java
@@ -16,12 +16,10 @@ import android.graphics.Paint;
 import android.graphics.PorterDuff;
 import android.graphics.Rect;
 import android.net.Uri;
-import android.util.Log;
 import com.facebook.common.executors.UiThreadImmediateExecutorService;
 import com.facebook.common.logging.FLog;
 import com.facebook.common.references.CloseableReference;
 import com.facebook.datasource.DataSource;
-import com.facebook.common.logging.FLog;
 import com.facebook.drawee.backends.pipeline.Fresco;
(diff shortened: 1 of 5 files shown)


magicismight:
I\`m afraid the Image issue haven\`t been fixed in this PR.
This is a screenshoot from example project.
image

silverspace:
Thanks for the feedback! 

I am unfortunately unable to reproduce after killing and restarting the Example app on both 4.4.2 (API 19) and 5.x (API 23) in the simulator. I also verified that the issue exists without my fix, but is fixed with my patch. 

image_fixed

If you have some cycles to debug this, could you please instrument your com.horcrux.svg.RNSVGSvgViewShadowNode#invalidateView method with a few log messages? 

In particular:
 1.  I suspect the problem is that svgView.getRootView() returns null. I'm just not sure how it got into this situation or how to correct from the situation. (Could this be related to the race condition TODO in RNSVGSvgViewManager?)  
  * One potential workaround would be to call svgView.invalidate() if rootView is null. However, I'm skeptical that this will work since that didn't fix the original issue in my testing. That's why I am calling rootView.invalidate() rather than svgView.invalidate(). 

 2.  A smaller possibility is that "dirtyView" has the wrong coordinates for the rootView to invalidate. Maybe try changing "rootView.invalidate(dirtyRect)" => "rootView.invalidate()" to see if it fixes your device?

magicismight:
Thanks for the detailed information.
It`s the second problem.
Use `rootView.invalidate()` works fine now

PR #99, opened 2016-08-03
translate then scale - fixes #98

### LINKED ISSUE #98: Incorrect ViewBox translation (Android only)
I'm using a ViewBox that is scaling a large portrait graphic down to fit the mobile screen, with default `preserveAspectRatio` - ie. `xMidYMid`. 

The aspect ratio of the graphic is narrower than the screen, so there should be margins to left and right. 

On iOS this renders correctly with equal margins on both left and right. 

On Android the graphic is offset to the right such that there is a large margin to the left and part of the graphic is off-screen to the right.

steveliles:
Lucky for me you left a comment in the code pointing to [the draft spec algorithm](https://svgwg.org/svg2-draft/coords.html#ComputingAViewportsTransform). I noticed the following mismatch between the spec and the Android implementation:

Spec:
> The transform applied to content contained by the element is given by translate(translate-x, translate-y) scale(scale-x, scale-y).

Implementation:
```
mMatrix.postScale(scaleX, scaleY);
mMatrix.postTranslate(-translateX * (mFromSymbol ? scaleX : 1), -translateY * (mFromSymbol ? scaleY : 1));
```

Switching the order to match the spec (translate then scale) resolves #98

### DIFF
--- a/android/src/main/java/com/horcrux/svg/RNSVGViewBoxShadowNode.java
+++ b/android/src/main/java/com/horcrux/svg/RNSVGViewBoxShadowNode.java
@@ -159,8 +159,11 @@ public class RNSVGViewBoxShadowNode extends RNSVGGroupShadowNode {
 
         }
 
-        mMatrix.postScale(scaleX, scaleY);
+        // The transform applied to content contained by the element is given by
+        // translate(translate-x, translate-y) scale(scale-x, scale-y).
         mMatrix.postTranslate(-translateX * (mFromSymbol ? scaleX : 1), -translateY * (mFromSymbol ? scaleY : 1));
+        mMatrix.postScale(scaleX, scaleY);
+        
         super.draw(canvas, paint, opacity);
     }
 


magicismight:
Thanks for this PR.
But I don\`t this is the right order.
The document is said how to implement transform  on `viewport` of SVG.not how to transform the children inside viewBox,it is all opposited.
And some conditions are not included in the document.
I have compared the viewBox inside react-native-svg with the web svg.It all works the same.

But I`m not so sure about this.I`ll test it later.

steveliles:
OK sure - its good to check these things ;) 

It seems there must be something wrong though - the output from the same SVG code is different between Android and iOS.

To replicate, just render this:

```
        <View style={{flex:1}}>
          <Svg style={{flex:1}} viewBox='0 0 500 1000'>
            <Rect x={0} y={0} width={500} height={1000} fill='#f00'/>
          </Svg>
        </View>
```

On iOS this will render correctly with the red rectangle centred horizontally. 



On Android it will render the rectangle aligned to the right of the screen (and some of it is off-screen to the right).



With the matrix operations reversed, Android also renders correctly.

magicismight:
You are right,it should translate first.
That is wierd,
I thought I have this tested,It works all fine by then.

magicismight:
Thanks for this  PR
:+1:

normanzb:
is this a regression issue? the latest version does postTranslation then preScale:
https://github.com/react-native-svg/react-native-svg/blob/4123a46/android/src/main/java/com/horcrux/svg/ViewBox.java#L100

which is different from what is merged here.
@magicismight

PR #100, opened 2016-08-03
fixes #96 for Android only - viewBox or nested G elements stifle events

### LINKED ISSUE #96: Touch events are not mapped to the viewBox
In 4.0.0-rc1, if I have an SVG with no viewBox attribute the touch events work correctly - touching the visible shape generates an event, touching outside the shape does not.

If I add a viewBox attribute the touch events aren't translated/scaled into the new viewBox, so touching the visible elements doesn't generate an event, but touching the place where they would appear without the viewBox attribute does generate an event.

steveliles:
This appears to fix #96 on Android ... I'm not much use with iOS although I'm prepared to have a look at it maybe tomorrow. I suspect the same logic problem is tripping up the iOS version too ...

### DIFF
--- a/android/src/main/java/com/horcrux/svg/RNSVGGroupShadowNode.java
+++ b/android/src/main/java/com/horcrux/svg/RNSVGGroupShadowNode.java
@@ -91,9 +91,10 @@ public class RNSVGGroupShadowNode extends RNSVGPathShadowNode {
 
             RNSVGVirtualNode node = (RNSVGVirtualNode) child;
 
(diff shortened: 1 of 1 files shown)


magicismight:
Thanks for this PR.
I`ll have this tested and get merged later,maybe tomorrow.

magicismight:
Thanks for those awesome PRs.

magicismight:
Thanks for this PR
:+1: 
Awesome!

PR #102, opened 2016-08-04
fixes #96 for iOS only - viewBox or nested G elements stifle events

### LINKED ISSUE #96: Touch events are not mapped to the viewBox

steveliles:
iOS was suffering a similar logic problem for #96 as Android - slightly worse I think, since G elements would _always_ steal the event on iOS, whereas it needed a double-nested G to steal the event on Android. 

First ObjectiveC code I've ever written :)

### DIFF
--- a/ios/Elements/RNSVGGroup.m
+++ b/ios/Elements/RNSVGGroup.m
@@ -11,15 +11,15 @@
 @implementation RNSVGGroup
 
 - (void)renderLayerTo:(CGContextRef)context
-{    
+{
     RNSVGSvgView* svg = [self getSvgView];
     [self clip:context];
-    
+
     for (RNSVGNode *node in self.subviews) {
         if ([node isKindOfClass:[RNSVGNode class]]) {
             [node mergeProperties:self mergeList:self.propList inherited:YES];
             [node renderTo:context];
-            
+
             if (node.responsible && !svg.responsible) {
                 self.responsible = YES;
             }
@@ -46,7 +46,11 @@
         if ([node isKindOfClass:[RNSVGNode class]]) {
             UIView *view = [node hitTest: point withEvent:event];
             if (view) {
-                return self;
+                if (node.responsible || node != view) {
+                    return view;
+                } else {
+                    return self;
+                }
             }
         }
     }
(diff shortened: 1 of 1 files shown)


steveliles:
This PR does fix a part of the issue: that events were stifled by nested G elements.

It doesn't completely fix the viewBox touch mapping though.

With this PR applied I do start receiving the touch events again, but the touches are received by the wrong elements if the viewBox has moved them significantly (see the original description of #96).

I spent a while looking into that problem too.

From what I can see, the CGPoint of the touch event is the pixel position of the touch on the enclosing view. Without a viewBox, or with an identity viewBox, this works and the touch events are received by the correct elements, but once you have a viewBox that transforms the elements  significantly I think maybe the CGPoint needs to be transformed too? 

I tried transforming the CGPoint using the viewBox matrix but it didn't seem to quite work, even though I can see that the transformed point is then within the bounds of the transformed shape. I'm guessing I'm doing something else wrong due to my inexperience with ObjectiveC and iOS in general.

magicismight:
I have been fixing the Path transform bug those days
But haven`t got any progress.
I will try another way to fix this

steveliles:
OK, cool. I do think it is worth applying this PR though, otherwise events still won't propagate through a G at all ;)

magicismight:
Yes.I`ll get this merged after I fix the transform bug

steveliles:
So I tracked down and fixed the transform problem before I noticed you'd already committed a fix that was cleaner than mine anyway :) 

With the addition of this PR to the current master the touch events get passed through G elements and map correctly to the element you are trying to touch.

magicismight:
Thanks for this PR.
:+1:

magicismight:
I think I have fixed it.
But not sure if it's fixed in the right way.
The hitTest will always called 3 times, and the last time called with a null event, can't figure out what happened.
But everthing works fine now on the Example project.

PR #105, opened 2016-08-04
Fixes #104 - re-draws on Android apply viewbox transforms repeatedly

### LINKED ISSUE #104: ViewBox transforms are multiplied each time the image is redrawn (Android only)
This is an Android-only bug.

Reproduction is easy:

* Create an SVG with a viewBox that differs from identity
* Do something that triggers a redraw, e.g. in `componentDidMount`, cause another shape to be added to the SVG

Each time a re-draw is triggered, the resulting image scales and translates further and further. In my case the image gets smaller and moves up and to the left on each re-draw, but it will depend on the viewBox settings.

steveliles:
fixes #104 by resetting the matrix at each draw to prevent applying the transforms on top of the previous transforms


magicismight:
Thanks for this PR.
:+1:

PR #124, opened 2016-08-22
Make it work with RN@0.33: Use RCTFont instead of RCTConvert

ptmt:


magicismight:
Thanks for this PR.
I'll get this merged after 0.33 is released
:+1:

neiker:
RN 0.33 is out

PR #133, opened 2016-08-31
Handle touches when root is a ViewRootImpl

slycoder:
Sometimes (maybe in modals?) the root is not a ReactRootView, it’s an
object which is not castable to a View.  Check for these cases before
trying to cast.


magicismight:
Thanks for this PR
:+1:

PR #136, opened 2016-09-06
fix passing numbers to Text element

gorangajic:
Currently, if you pass `number` to `Text` your app will be broken.


magicismight:
Thanks for this PR
:+1:

gorangajic:
Thank you

PR #149, opened 2016-09-26
replace original heavy hitTest method bitmap -> region.contains point

weixingsun:


gorangajic:
This does not work when `viewBox` is set. Is there any way to fix that?

weixingsun:
@gorangajic sorry, I am not familiar with viewBox, is that some way your define the border of a shape or a template?

gorangajic:
@weixingsun https://developer.mozilla.org/en/docs/Web/SVG/Attribute/viewBox

I have figured how to respect viewBox property, would you like to include that in your commit ?

```
    @Override
    public int hitTest(Point point, View view, @Nullable Matrix matrix) {
        RectF rectF = new RectF();
        Path path = new Path(mPath);
        if (matrix != null) {
            path.transform(matrix);
        }
        path.transform(mMatrix);
        path.computeBounds(rectF, true);
        Region region = new Region();
        region.setPath(path, new Region((int) rectF.left, (int) rectF.top, (int) rectF.right, (int) rectF.bottom));
        if (region.contains(point.x, point.y)) {
            return view.getId();
        }

        return -1;
    }
```

gorangajic:
btw `<Text />` still uses canvas to draw text and then figure out if it's hit. Any idea how we can calculate text `Region` or do something more performant?

weixingsun:
@gorangajic thanks for your info, I merged your codes into this commit, it works for me for no viewBox props, 
please let me know if your viewBox does not work.
For Text, I saw RNSVGTextShadowNode extends from RNSVGPathShadowNode, 
so what if we just remove the two hitTest methods? ping me if you got an idea.

weixingsun:
@gorangajic I am trying to reuse the RectF in method RNSVGTextShadowNode.draw(Canvas canvas, Paint paint, float opacity) (as global rectBorder), but I cannot make it work
    @Override
    public int hitTest(Point point, View view, @Nullable Matrix matrix) {
        RectF rect = new RectF();
        if (matrix != null) {
            matrix.mapRect(rect,rectBorder);
        }
        mMatrix.mapRect(rect,rectBorder);
        FLog.w(ReactConstants.TAG, "rect="+rect.toShortString()
            +"\n  x="+point.x+"  y="+point.y);
        if (rect.contains(point.x, point.y)) {
            return view.getId();
        }
        return -1;
}
=======================================
rect=[619.0,415.46667][628.0,436.46667]
x=599  y=454
the touching point is always below the rect, I don't know why. please help @magicismight

magicismight:
I'm sorry I cannot help with this right now, I'm quite busy those days. I'll try to get this reviewed next week.

gorangajic:
@weixingsun I have tried using `Matrix.mapRect` and it didn't work so I went with `Path.transform`. For now, I have disabled touch events for `Text` elements. 

@magicismight do not worry when you can. Thank you for the work you already put in the lib 😄 👍

gorangajic:
@magicismight do you know maybe why touch events will be buggy on galaxy s6 and everything works fine in simulator and nexus 5?

gorangajic:
This test is fine, something else is the problem will create a new issue

magicismight:
Thanks for this PR.
I have refactored the whole system related to touch events, and using the region.contains instead of Bitmap.Thanks for this great idea, but I have to close this now.
You can find the related PR [here](https://github.com/react-native-community/react-native-svg/pull/226/files)

PR #152, opened 2016-10-04
Use regular class methods instead of static class properties

ghengeveld:
This fixes a `this` binding issue causing _this.onReset in SerializablePath to throw 'undefined is not a function'. This happens when not using React Native's default Babel config, which is a requirement for some projects [such as Relay](https://facebook.github.io/relay/docs/guides-babel-plugin.html).

### DIFF
--- a/lib/SerializablePath.js
+++ b/lib/SerializablePath.js
@@ -15,7 +15,7 @@ export default class SerializablePath {
     }
 
     /* parser */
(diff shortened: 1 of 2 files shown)


magicismight:
Thanks for this PR 👍 
But we had moved the `d` prop serialization to the native side.
I'm closing this now.

Issue #115, opened 2016-08-15
Text not updating


josephchang:
When I use state to change the content of `<Text>`, it doesn't update. As you can see in the screenshot, the first number is using React Native's Text and the second using RNSvg's, both with the same state value, but the latter is not updating.

<img width="222" alt="screen shot 2016-08-15 at 10 37 08" src="https://cloud.githubusercontent.com/assets/1324815/17660985/45ae0d3a-62d4-11e6-8cd4-565ed3bfdda1.png">

DominicTobias:
I'm having this issue too :(

josephchang:
I have another app using 4.0.0-rc of this, and the text updates based on state, I think the combination of `react-native` 1.5.2 and `react-native-svg` 4.1.1 is problematic...or something.

gorangajic:
For me it's not working on android only, can someone check that? @magicismight

Char-Ten:
Similar issue but it's not updating with G element's props in Android.
the RN vision is 0.38
the RNSVG vision is 4.40
when I change the props to rotated G element,it doesn't work in Android
```xml
<G 
    rotate={`${this.props.value}`}
    width={SIZE}
    height={SIZE}
    origin={`${HALFSIZE},${HALFSIZE}`}
    id="hand">
     <!--other elements-->
</G>
```

PR #153, opened 2016-10-06
Fix android text update #115

gorangajic:
after text is changed on android it's not updated #115


magicismight:
Thanks for this PR
👍

Issue #122, opened 2016-08-20
Trouble with custom fonts


dubert:
I added a custom font to my app- and I can use it with the RN `<Text>` component. (The font in question is: AbadiMT-CondensedExtraBold)

However, if I try to use it in the SVG's `<Text>` component, it does not show up. I'm using the SVG one because I want to be able to add a stroke to the text, and RN does not offer a way to do this.

So as an example:
```js
<View>
  <Text style={{
      color: '#fff',
      fontSize: 32,
      fontFamily: 'AbadiMT-CondensedExtraBold',
   }}>
     {score}
  </Text>
</View>
```
works fine and shows the correct font.

But
```js
<Svg width={100} height={50}>
  <Text
    fill='#fff'
    stroke='#000'
    fontSize={32}
    fontFamily='AbadiMT-CondensedExtraBold'
    x={0}
    y={0}
  >
     {score}
  </Text>
</Svg>
```
shows up with the default font (looks like Helvetica).

Note: I'm aware of the namespace conflict, but I'm sure I didn't switch up the text components. They're in different files.

lewisbarber:
Is there a solution to this???

chengyin:
The way the Android drawing environment works requires specifying not just the font family name, but also the font file path. I suspect iOS has something similar, since it is also out of UIKit.

rgoldiez:
@chengyin - Your fix/PR works for me!  Thank you!!

chengyin:
Note that my PR only addresses iOS. Android is trickier since you can just specify a custom font name. It has to be a custom font path. I am not sure where to hide that complexity.

pengfeiyan:
I got the same issue.

How to use custom fonts for text on Android device?

kopax:
Same question here, I use react native and I don't know how I can use my app font family within svg.

CaioMars:
Same here

AdamGerthel:
Still no solution to this?

Josema:
> 

How did you fix it? I'm having the same problem.

Jobaer-hassan-orangetoolz:
same problem

VitaliiBlack:
Same issue

PR #163, opened 2016-10-19
Fixes issue with custom fonts #122

lewisbarber:


chengyin:
#166 is addressing the same issue with a safer approach.

PR #166, opened 2016-10-26
Better font family check on iOS to support custom fonts

### LINKED ISSUE #122: Trouble with custom fonts

chengyin:
Font names can be specified as something like `Helvetica-Bold`. This is specially common when using custom fonts. This fix makes sure those supported custom fonts are not accidentally removed.

Fix #122

### DIFF
--- a/ios/Utils/RCTConvert+RNSVG.m
+++ b/ios/Utils/RCTConvert+RNSVG.m
@@ -102,9 +102,22 @@ RCT_ENUM_CONVERTER(RNSVGVBMOS, (@{
     NSDictionary *fontDict = dict[@"font"];
     NSString *fontFamily = fontDict[@"fontFamily"];
     
-    if (![[UIFont familyNames] containsObject:fontFamily]) {
-        fontFamily = nil;
+    BOOL fontFound = NO;
+    NSArray *supportedFontFamilyNames = [UIFont familyNames];
+
+    if ([supportedFontFamilyNames containsObject:fontFamily]) {
+      fontFound = YES;
+    } else {
+      for (NSString *fontFamilyName in supportedFontFamilyNames) {
+        if ([[UIFont fontNamesForFamilyName: fontFamilyName] containsObject:fontFamily]) {
+          fontFound = YES;
+          break;
+        }
+      }
     }
+
+    fontFamily = fontFound ? fontFamily : nil;
+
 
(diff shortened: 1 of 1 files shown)


magicismight:
Thanks. 👍 
I'll get this merged later

brentvatne:
@magicismight - why is this check necessary at all? can't we just pass the fontFamily directly into `RCTFont updateFont:` and let it do the work for us?

magicismight:
@brentvatne Because if you have specified a fontFamily which is not supported, `RCTFont updateFont:` will throw an error. But I prefer it to fallback to the default fontFamily instead.

brentvatne:
Ah I see @magicismight. This seems like a behaviour that should live in react-native itself rather than react-native-svg, no? Otherwise fontFamily behaves differently on <Text> and <Svg>

rgoldiez:
This patch does not seem to work (or have made it in?) to `react-native-svg@5.0`

chengyin:
@rgoldiez this patch only works for iOS. Custom font is harder with Android since it is always file based, and I don't know a way to register a new font that can be referred with font names.

rgoldiez:
@chengyin - Got it.  I had this patch working fine on iOS using `react-native-svg@4.4.1` and my custom font (that I use elsewhere in my app).  I just went through the process of upgrading from RN 38 to RN 41, and to `react-native-svg@5.1.5` the charts in my app aren't showing the font (possibly a problem with the chart component, not this one).

PR #184, opened 2016-11-22
Mark all descendants of Defs as seen after a draw

AlbertBrand:
When adding clipping paths in a def and changing properties of that clipping path (or its descendants), redraw is triggered only once as updates are never marked as seen.
This change makes sure to call `markUpdateSeen` on all descendants of the Defs node on a `draw`.

### DIFF
--- a/android/src/main/java/com/horcrux/svg/RNSVGDefsShadowNode.java
+++ b/android/src/main/java/com/horcrux/svg/RNSVGDefsShadowNode.java
@@ -28,7 +28,7 @@ public class RNSVGDefsShadowNode extends RNSVGDefinitionShadowNode {
         traverseChildren(new NodeRunnable() {
             public boolean run(RNSVGVirtualNode node) {
                 node.markUpdateSeen();
-                node.traverseChildren(this);
+                node.traverseChildren(NodeRunnable.this);
                 return true;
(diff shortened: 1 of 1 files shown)


android/src/main/java/com/horcrux/svg/RNSVGDefsShadowNode.java
```diff
@@ -25,5 +25,12 @@ public boolean run(RNSVGVirtualNode node) {
                 return true;
             }
         });
+        traverseChildren(new NodeRunnable() {
+            public boolean run(RNSVGVirtualNode node) {
+                node.markUpdateSeen();
+                node.traverseChildren(this);
```
magicismight: LGTM.
But can you assign the NodeRunnable to a variable?
The `this` is a little confused.

AlbertBrand: I could make it a `NodeRunner.this` for more clarification.

AlbertBrand:
My previous fix didn't work (never compiled it, lol). So I've given the NodeRunnable a nice name now.

AlbertBrand:
Any news on this? Can I do anything to improve it further?

PR #191, opened 2016-11-24
Remove dependency on lodash in ViewBox.js

admmasters:
We can just use the array.prototype native methods for some and reduce here rather than using lodash. Also fixed usage of let to const where the params are not re-assigned.


magicismight:
Thanks for this PR
👍

PR #238, opened 2017-01-22
Fixes/5.1.4

magicismight:


SudoPlz:
What about the ones that are on react-native 0.39.2 what can we do to use those changes?

We cannot use 5.1.4, is there something like 4.4.1.4 ?

magicismight:
@SudoPlz  You have to upgrade react-native to use those changes. sorry about that

PR #243, opened 2017-01-26
illustrator compatibility for polyline and polygon elements

clozr:
The svg generated from illustrator has a flat list of points without any commas separating the points. This change takes care of the parsing for illustrator generated format, reducing the need for manual change of the point specification.


magicismight:
Thanks for this PR 👍 
But I have to close this now.
This had been fixed in https://github.com/react-native-community/react-native-svg/pull/298/commits/e3b2f4c71add751c29eaa572bb8f0555aca305fb

PR #253, opened 2017-02-10
Replace RNSVGBrushConverter type with RNSVGPainter

EuAndreh:
This modification was made in cfaca8b3cb166207357b7b945151762db81d3053, but was missing in a private method signature inside RNSVGPattern.m.


gitim:
cc @magicismight

magicismight:
Thanks for this PR

PR #268, opened 2017-03-02
Fix TextShadowNode cast to fix Victory Native issue https://github.com/FormidableLabs/victory-native/issues/50

guitoof:


sibelius:
LGTM

AlanFoster:
Confirmed that this branch works locally; Also cross referencing the related issue:

https://github.com/FormidableLabs/victory-native/issues/50

browniefed:
@magicismight any chance of this getting merged? Would be really great to unblock a lot of people and fix `victory-native`

magicismight:
This is not a bug from react-native-svg.
https://github.com/FormidableLabs/victory-native/issues/50
The error mentioned ReactVirtualTextShadowNode cannot be cast to TextShadowNode.
[ReactVirtualTextShadowNode](https://github.com/facebook/react-native/blob/master/ReactAndroid/src/main/java/com/facebook/react/views/text/ReactVirtualTextShadowNode.java) is a component from ReactNative.Text, it cannot be rendered in Svg component.

It seems like Victory Native imported Text from 'react-native' instead of 'react-native-svg', and used it inside a Svg component.

magicismight:
I'll add a type check to prevent a crash like this.
But this bug should inform the collaborators of Victory Native

browniefed:
@magicismight Ah thank you for the response and for pointing out the probably cause. I'll see about tracking down where it gets imported from mistakenly.

Was a curious issue as it doesn't throw on iOS but throws on Android

browniefed:
I looked all through `victory-native-*` and could not see any place that imports `Text` from `react-native` instead of `react-native-svg`.

I'll dig deeper, try and create the bare minimum of repro and see if I can find out what's going on with Android.

I assumed if this was the case then it would also break iOS?

PR #274, opened 2017-03-05
5.1.6

magicismight:


gitim:
@magicismight is something stopping from publishing this version to npm?

Issue #79, opened 2016-07-19
Error - No component found for view with the name "RNSVGRect"


StanDZ:
I'm just starting to use the library and started by simply pasting the bit of code given in the example, but although I've installed and imported all the elements as indicated, I get this error message.. Am I missing something obvious ?

The bit of code : 
 
       <Svg height="100" width="100">
          <Rect x="0" y="0" width="100" height="100" fill="black" />
          <Circle cx="50" cy="50" r="30" fill="yellow" />
          <Circle cx="40" cy="40" r="4" fill="black" />
          <Circle cx="60" cy="40" r="4" fill="black" />
          <Path d="M 40 60 A 10 10 0 0 0 60 60" stroke="black" />
        </Svg>;

The imports:
`import {
    Svg,
    Circle,
    Line,
    Rect,
    Path,
} from 'react-native-svg';
`

And the error message : 

<img width="375" alt="screen shot 2016-07-19 at 16 04 52" src="https://cloud.githubusercontent.com/assets/20460067/16952811/1e68bdc4-4dcb-11e6-895f-9bd9a832e2a9.png">

magicismight:
It seems like you haven`t link the native code for the project.
Run `react-native link react-native-svg` or `rnpm link react-native-svg`

gHashTag:
The same thing, Run react-native link react-native-svg or rnpm link react-native-svn, did not help

fdnhkj:
After doing `rpm link`, you need to re-run the app by executing `react-native run-ios` or `react-native run-android` depending on the platform you're using.

julesmoretti:
On my end it sends me to this:

```
own source file : 	at com.android.dx.command.dexer.Main.run(Main.java:277)
Unknown source file : 	at com.android.dx.command.dexer.Main.main(Main.java:245)
Unknown source file : 	at com.android.dx.command.Main.main(Main.java:106)

:app:dexDebug FAILED

FAILURE: Build failed with an exception.

* What went wrong:
Execution failed for task ':app:dexDebug'.
> com.android.ide.common.process.ProcessException: org.gradle.process.internal.ExecException: Process 'command '/Library/Java/JavaVirtualMachines/jdk1.8.0_112.jdk/Contents/Home/bin/java'' finished with non-zero exit value 2

* Try:
Run with --stacktrace option to get the stack trace. Run with --info or --debug option to get more log output.

BUILD FAILED

Total time: 15.994 secs
[...]

PR #282, opened 2017-03-15
Appending restart instruction to readme

bellentuck:
In response to issue #79, regarding the "No component found for view with the name [insert SVG shape here]" error, @fdnhkj mentions in the comments that "After doing rpm link, you need to re-run the app by executing react-native run-ios or react-native run-android depending on the platform you're using." This approach also worked for me in response to this problem. So, I propose adding this instruction explicitly into the react-native-svg documentation to give future noobs like myself that little bit more peace of mind. While re-launching the app may seem an obvious tactic to try in retrospect, it did take me coming across that comment to do so!
Thanks for all your work on bringing SVG to React Native. :)

### DIFF
--- a/README.md
+++ b/README.md
@@ -23,15 +23,15 @@
     ```bash
     npm install react-native-svg --save
     ```
-    
+
     # NOTICE:
-    
+
     - react-native-svg >= 3.2.0 only supports react-native >= 0.29.0
     - react-native-svg >= 4.2.0 only supports react-native >= 0.32.0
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -44,7 +44,8 @@
     ```bash
     rnpm link react-native-svg
     ```
-    
+3. Re-run the app: re-execute `react-native run-ios`/`react-native run-android`.
```
dustinsavery: This is pretty self explanatory for react-native.

PR #321, opened 2017-04-27
Fix text position on Y-axis, we should take fontSize in mind to layout the Text/TSpan shapes

ssssssssssss:
SVG coord tes case https://www.w3.org/TR/SVG/images/coords/OrigCoordSys.svg


ssssssssssss:
Close as it's been picked into 6.0.0 branch.

PR #328, opened 2017-05-04
Fix setNativeProps for all elements that inherit from Shape.  

udfalkso:
The native props being set were not passing through extractProps, leading to various mismatches in the values. 

More discussion here:
https://github.com/react-native-community/react-native-svg/issues/180 

And here:
https://github.com/react-native-community/react-native-svg/issues/326

### DIFF
--- a/elements/Shape.js
+++ b/elements/Shape.js
@@ -1,5 +1,6 @@
 import {Component} from 'react';
 import SvgTouchableMixin from '../lib/SvgTouchableMixin';
+import extractProps from '../lib/extract/extractProps';
 import _ from 'lodash';
 
 class Shape extends Component {
@@ -10,6 +11,17 @@ class Shape extends Component {
         });
         this.state = this.touchableGetInitialState();
     }
+
+    setNativeProps = (args) => {
+        var extracted = extractProps(args, this)
+        for(var key in args) {
+          if (extracted[key]) {
+            args[key] = extracted[key]
+          }
+        }
+        this.root.setNativeProps(args);
+    };
+
 }
 
 export default Shape;
--- a/elements/Circle.js
+++ b/elements/Circle.js
@@ -21,10 +21,6 @@ export default class extends Shape {
         r: 0
     };
 
-    setNativeProps = (...args) => {
-        this.root.setNativeProps(...args);
-    };
-
     render() {
         let props = this.props;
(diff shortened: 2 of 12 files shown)


elements/Shape.js
```diff
@@ -10,6 +11,17 @@ class Shape extends Component {
         });
         this.state = this.touchableGetInitialState();
     }
+
+    setNativeProps = (args) => {
+        var extracted = extractProps(args, this)
```
yic03685: This however doesn't work for transform matrices. Ex. I want to specify a translation, if I specify in {transform: {translateX: 12}} the for loop afterwards will filter out the matrix extracted.

udfalkso: Thanks for your comment.  The for loop is just limiting the values that get changed to the ones returned by extractProps().  

So if extractProps() does what you want, then that value will persist.  Anything else is left alone. It looks like extractProps() does support transform matrices.  So I think that any transform you use without setNativeProps should also keep working here, no?

yic03685: Correct me if I'm wrong. `extractProps` does not expect matrices. It expects transform objects. As a result, if I pass a matrix directly, it will modify it to the identity matrix. And ofc, the for loop will let the matrix pass. 

Namely, I think the original design is to pass transform objects and the `extractProps ` will translate it into matrix. With your changes, it cannot do this anymore.

udfalkso: I see.  

I'm not sure the goal of setNativeProps is to act as a bypass like this.  I believe it's meant to be used to skip the react rendering flow for efficiency reasons alone.  If that's the case (please correct me if I'm wrong), then you would typically hope to see the same rendering behavior with regular props as well as setNativeProps.  

The problem that I was trying to solve for my use case is that initial rendering worked great, but then i could not change the value via setNativeProps (in my case for animating the fill color).  This PR fixes that for me and makes things consistent between the initial render and setNativeProps changes.  

It seems that you were working around some logic in this extractProps() function, when maybe the correct fix would be for us to fix that logic or add an explicit bypass for matrices there in addition to my changes?  Can we somehow change extractProps() to accept a matrix in the format you want?  

I'm not sure how many other people are working with your same expectation however.  It might break many people's code?

dustinsavery:
This does not appear to be needed. Closing this PR.

PR #357, opened 2017-06-07
Implement setGradientTransform

msand:
Compatible with
react-native >=0.40.0 <0.43.0
react >=15.4.0 <15.5.0
Still early work, but seems to work correctly in android at least, I personally don't have access to any Apple products to test with right now...


msand:
I've added support for Image transform as well now.

PR #350, opened 2017-05-29
Replace yoga's YogaNodeAPI with YogaNode

willdawsonme:
Fixes https://github.com/react-native-community/react-native-svg/issues/340. I imagine this will be a breaking change for react-native versions < 0.45.0.

PR #359, opened 2017-06-08
replace yoganode with yoganodeapi

Tooyz:
Fix "method does not override or implement a method from a supertype" in SvgViewManager class for version 5.2.0.


itsmepetrov:
These breaks build for react-native@0.45, related to #350

PR #384, opened 2017-06-23
Remove react-native from dependencies (only should be peerDep)

ide:
RN libraries should almost never depend on react-native and just specify it as a peerDep.

### DIFF
--- a/package.json
+++ b/package.json
@@ -28,8 +28,7 @@
     },
     "dependencies": {
         "color": "^0.11.1",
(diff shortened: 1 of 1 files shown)


Igor1201:
we should actually move it to `devDependencies`, so it's  easier to development

pietropizzi:
In any case, could this be merged?

ide:
Closing since this is effectively in master.

Issue #313, opened 2017-04-23
Image cropped itself when y over 100


isener:
I don't understand what happened. 
When I place Text property to svg, it works like charm,
But when I place Image same logic, it brokes image.

<img width="464" alt="screenshot at apr 23 19-43-57" src="https://cloud.githubusercontent.com/assets/10687714/25315425/3dd94d1e-285d-11e7-895a-278ba1319d35.png">

lschmierer:
Seems to be related to #312

isener:
It's not actual solution.there is something wrong on rendering image

LRNZ09:
This issue is still happening on iOS with version `6.3.1`

msand:
Please try with the latest version, or provide a full replication. Closing until then.

Issue #312, opened 2017-04-21
Image iOS render wrong origin


lschmierer:
Image rendering on iOS seems to be not quite right.

Following code
```
<Svg width="100" height="100">
  <Image
    x="0"
    y="0"
    width="200"
    height="200"
    href={this.props.image}
  />
</Svg>
```
produces
<img width="487" alt="screen shot 2017-04-21 at 10 07 38" src="https://cloud.githubusercontent.com/assets/2204708/25268799/6d9a3502-267a-11e7-9c13-c1cf9f99994e.png">

while following is expected:
<img width="487" alt="screen shot 2017-04-21 at 10 07 53" src="https://cloud.githubusercontent.com/assets/2204708/25268818/77acf020-267a-11e7-8435-080644cde517.png">

Setting the `y` property to `-50`, however, makes it render correctly:
```
<Svg width="100" height="100">
  <Image
    x="0"
    y="-50"
    width="200"
    height="200"
    href={this.props.image}
  />
</Svg>
```

I did not quite understand yet what is the rational of the number `-50` here.
The issue seems to be with the origin, the image is rendere, somehow.

On Android, everything seems to be correct.

msand:
@lschmierer Is this still in issue with the latest version?

lschmierer:
@msand I have no idea... switched to Flutter long time ago...

msand:
@LRNZ09 Can you try with the latest v7? Or even better, this branch: https://github.com/react-native-community/react-native-svg/compare/SVGLength

msand:
v8 has been released, this issue has like been fixed in https://github.com/react-native-community/react-native-svg/pull/763 which is available since v7 up.

PR #404, opened 2017-07-17
Fix image layout iOS [#312 #313]

ScrimpyCat:
Fixes the issue with image translation on iOS, as brought up in issues #312 and #313 

This fix has worked so far on examples I've tested it with, though I'm unclear as to what was the original intention for the `CGContextClipToRect(context, rect);` on line 136 in `RNSVGImage.m`. I removed it because it's causing the image to be clipped to its starting coordinate and size (so if the image is translated or scaled it will end up being cropped). But as I'm not sure why it was there originally, this change might cause some other issue? If it was there to handle some case, please let me know and I can fix that up.

### DIFF
--- a/ios/Elements/RNSVGImage.m
+++ b/ios/Elements/RNSVGImage.m
@@ -126,6 +126,7 @@
     
     CGRect vbRect = CGRectMake(0, 0, CGRectGetWidth(renderRect), CGRectGetHeight(renderRect));
     CGRect eRect = CGRectMake([self getContextLeft], [self getContextTop], rectWidth, rectHeight);
+    CGContextTranslateCTM(context, CGRectGetMinX(eRect), 0.0f);
     
     CGAffineTransform transform = [RNSVGViewBox getTransform:vbRect eRect:eRect align:self.align meetOrSlice:self.meetOrSlice fromSymbol:NO];
     
--- a/ios/Elements/RNSVGImage.m
+++ b/ios/Elements/RNSVGImage.m
@@ -133,7 +133,6 @@
     renderRect = CGRectApplyAffineTransform(renderRect, CGAffineTransformMakeTranslation(rectX, rectY));
     
     [self clip:context];
-    CGContextClipToRect(context, rect);
  
     CGContextDrawImage(context, renderRect, _image);
     CGContextRestoreGState(context);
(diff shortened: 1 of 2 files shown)


ScrimpyCat:
Actually found another issue with the translation, it isn't translated by the same amount relative to other content.

e.g.

```svg
<G x="100" y="50">
    <Rect x="0" y="0" width="50" height="50" fill="#fff"/>
    <Image
        x="0"
        y="0"
        width="50"
        height="50"
        href={image}
    />
</G>
```

The image will be on the correct y-axis but double the x-axis. I'll fix this other issue first.

ScrimpyCat:
Added the other changes. The align operations weren't updated to the current svg2 spec, so changed those. I also was unsure about what to do with the `none` meetOrSlice branch, since it's using the old equation (which is susceptible to divide by 0 errors), but I couldn't find any reference to it in the spec? So I left that unchanged.

Lastly the parent view bounds are already applied to the current context matrix, but the `+getTransform:eRect:align:meetOrSlice:fromSymbol:` method doesn't account for this and applies it again. So I just reverted that x-axis in the context matrix to avoid this issue.

msand:
Looks great, I've made similar fixes to android inspired by your pull request. And some more fixes to viewbox and image translation / scaling / transformation. https://github.com/react-native-community/react-native-svg/pull/406/commits/8ffa2f3373e2fd9c9227459733a28b2b4ea1ae59

ScrimpyCat:
Forgot to commit some of the additional needed changes, but now everything should work.

@msand Cool, I'll have a read through and see if there's anything else I can update on this one.

msand:
Great. I have a test suite in https://github.com/msand/RNSvgTextBug for comparing chrome and android, haven't started work on porting anything to ios yet. Just read my first objective-c syntax summary today and will start looking at that part soon. Meanwhile, if you're interested, you could adapt it to ios; to have more reference material.

msand:
@ScrimpyCat I've now ported almost all the changes from android to ios and we're tracking it here: https://github.com/react-native-community/react-native-svg/pull/430
https://github.com/msand/react-native-svg/tree/6.0.0
Would you like to review it? So we get it good enough for merging into master and published to npm asap.

magicismight:
This has been fixed in 6.0.0.
I'm closing this PR, thanks for this PR anyway.

msand:
@michaelgobbers Can you try with the latest version? Or did you resolve the issue?

LRNZ09:
This issue is still happening on iOS with version `6.3.1`

PR #406, opened 2017-07-17
5.3.0 conformance

msand:
I've done quite a bit of work to improve the svg specification conformance of the implementation, using the svg 1.1, tiny, and 2.0 specs, and chrome as reference implementation, using godadddy/svgs to compare results of various test cases. I haven't made any of the corresponding fixes for ios yet, but think it might be a good time to merge these into master. Or what do you think?

### DIFF
--- a/.npmignore
+++ b/.npmignore
@@ -3,3 +3,4 @@ node_modules/
 Example/
 experimental/
 android/build/
+screenShoots/
--- a/README.md
+++ b/README.md
@@ -75,7 +75,7 @@
 
 Here's a simple example. To render output like this:
 
-![SVG example](./screenShoots/svg.png)
+![SVG example](https://raw.githubusercontent.com/react-native-community/react-native-svg/master/screenShoots/svg.png)
 
 Use the following code:
 
@@ -192,7 +192,7 @@ The <Rect> element is used to create a rectangle and variations of a rectangle s
 </Svg>
 ```
 
-![Rect](./screenShoots/rect.png)
+![Rect](https://raw.githubusercontent.com/react-native-community/react-native-svg/master/screenShoots/rect.png)
 
   Code explanation:
 
@@ -218,7 +218,7 @@ The <Circle> element is used to create a circle:
 </Svg>
 ```
 
-![Rect](./screenShoots/circle.png)
+![Rect](https://raw.githubusercontent.com/react-native-community/react-native-svg/master/screenShoots/circle.png)
 
   Code explanation:
 
@@ -247,7 +247,7 @@ An ellipse is closely related to a circle. The difference is that an ellipse has
     />
 </Svg>
 ```
-![Rect](./screenShoots/ellipse.png)
+![Rect](https://raw.githubusercontent.com/react-native-community/react-native-svg/master/screenShoots/ellipse.png)
 
 Code explanation:
 
@@ -276,7 +276,7 @@ The <Line> element is an SVG basic shape, used to create a line connecting two p
 </Svg>
 ```
 
-![Rect](./screenShoots/line.png)
+![Rect](https://raw.githubusercontent.com/react-native-community/react-native-svg/master/screenShoots/line.png)
 
 Code explanation:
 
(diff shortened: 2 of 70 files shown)


magicismight:
Cool! Thanks for this PR.
I'll check out this PR in the week.

msand:
I'm currently working on improving/optimizing the glyphcontext memory use and computation. And, I got to borrow a mac yesterday, so I'm now learning objective-c to be able to port the fixes to ios.

msand:
Propagation, of multiple x or y values expressed as a comma or white-space separated string, over multiple nested tspans, is not supported yet. And other units than viewbox relative units and fontsize relative (em) units are not supported either. Same goes for right-to-left and vertical text. Otherwise I think everything from the text related examples in https://www.w3.org/TR/SVG11/ work correctly.

msand:
Optimally, the SVG 2.0 text layout algorithm should be implemented. It is quite explicit and clear worded, and supports bi-directional, horizontal/vertical text and text on a path, and even text layout content wrapping areas and shape-subtract: https://www.w3.org/TR/SVG2/text.html#TextLayoutContentArea
https://www.w3.org/TR/SVG2/text.html#TextLayoutAlgorithm

msand:
I think I've fixed all the remaining issues I listed today, except bidi and rtl text.
I realized, svg inside svg (as in the PreserveAspectRatio example) doesn't work on android, would anyone have any clues why?

msand:
ViewBox, Text, Image, Gradient and Path rendering, and generic matrix transforms, in react-native-svg android; now conforms with chrome, firefox, opera, and, edge; on desktop windows 10. And, on mobile chrome and firefox on android.

The main remaining issue on android is the inclusion of svg inside svg. Any clues about this would be very helpful. It is possible to include jsx from other components, but the ViewBox transforms and PreserveAspectRatio would be quite useful to have as well.

msand:
I've made an alternative GlpyhContext implementation, which should have better memory characteristics and computational performance. The hot-spots (methods which are called for every character / glyph) 
 should be quite optimized in that version now. 

https://github.com/msand/react-native-svg/tree/OptimizedGlpyhContext

I haven't made any proper tests to compare performance yet, but at least in theory it should be significantly better on any text heavy svg view. Depending on various kinds of nesting and what attributes are set by how many ancestors, and how many values are set for one attribute in one node.

msand:
I merged the optimized version into this pull branch, as it seems clearly superior and closer to the svg 2.0 text layout algorithm as well.

msand:
I noticed units in the x, y, dx, dy and font-size attributes didn't work correctly, so I've fixed that as well. Now font-size in % or em relative to parent font size should work as well. This probably requires an api change in the ios / objective-c code as well, to send all of these attributes as SVGLengthList, i.e., a ReadableArray containing only String elements, in the js -> java case.

https://www.w3.org/TR/SVG/types.html#InterfaceSVGLengthList

msand:
The implementation is good to go for porting to obj-c/ios now I think. There are some new units in css 3 / svg 2, and the named absolute and relative sizes aren't implemented either, but I think both of them are relatively rare in the wild still. So they can be implemented later on and ported across, probably without too much effort.

msand:
I just realized there might be potential for a large optimization still, by caching the character glyphs, such that if the same character is rendered more than once then it can just clone the Path corresponding to that glyph, instead of calling Paint.getTextPath every time.

GeeWee:
I don't have anything to do with the project - I just want to say this looks like really awesome work.

msand:
I intend to implement this still before starting/meanwhile porting to obj-c:
https://drafts.csswg.org/css-fonts-3/#font-rend-desc

msand:
Alright, now Low-level font feature settings control: the font-feature-settings property is done, http://dev.w3.org/csswg/css-fonts/#propdef-font-feature-settings

Ready for porting to ios, it is essentially feature complete at this point.

msand:
@magicismight Have you had any time to look at this?

msand:
@GeeWee Check out https://github.com/react-native-community/react-native-svg/pull/430 
It has almost all the fixes from android ported to ios now.

PR #421, opened 2017-08-03
support 0.47.0

magicismight:


cooperka:
FYI you can maintain backwards compatibility if you just remove the `@Override` annotation but keep the method. For example: https://github.com/rebeccahughes/react-native-device-info/pull/191/files.

PR #426, opened 2017-08-03
Podspec support for Apple TV

dlowder-salesforce:
One line change to allow the RNSVG podspec to be used for Apple TV projects.


magicismight:
Thanks for this PR.

PR #430, opened 2017-08-10
iOS SVG spec conformance work

msand:
Initial work on porting fixes from Android to iOS, continuation of https://github.com/react-native-community/react-native-svg/pull/406 and https://github.com/react-native-community/react-native-svg/pull/428


msand:
@magicismight @fxwan @ssssssssssss I've now ported the new glyph context and font data, and text rendering logic. some minor issues with kerning and ligatures exist on iOS, and there is some potential for optimisations. Would you like to have a look, and if you find anything you consider in need of / worth fixing, just make a comment/review. And then lets try to get it merged into master and published as the new major version. Quite a few people are anticipating its release.

ios/RNSVGNode.m
```diff
@@ -9,14 +9,21 @@
...
 {
+    RNSVGGroup *_textRoot;
+    RNSVGGlyphContext *glyphContext;
     BOOL _transparent;
     CGPathRef _cachedClipPath;
     RNSVGSvgView *_svgView;
 }
 
+CGFloat const M_SQRT1_2l = 0.707106781186547524400844362104849039;
+CGFloat const DEFAULT_FONT_SIZE = 12;
```
ssssssssssss: No success to find the default font-size in spec. Any hints?

ssssssssssss:
I've looked through the commits, seems like a direct port of Android version. I do not have solid knowledge about detailed SVG spec, but other parts look good. Are you running SVG tests on iOS against Chrome browser?

msand:
I've mainly used the android version as reference, as I've already improved the conformance of that beyond chrome, firefox and safari, to similar levels as edge. It is very much a straight port, I've used https://developers.google.com/j2objc/ to make an initial conversion of the glyphcontext, fontdata and enums to obj-c, but had to remove almost all of it and rewrite/tailor make the rest anyway to fit closer to maintainable obj-c patterns and be readable at all. It contains two libraries for the moment as well, which can be removed, by rewriting to a tailor-made bezier curve length integration / numerical approximation (with adjustable tolerance) and position/tangent calculator. I just feel a bit of urgency to get something working out the door asap.

msand:
I don't have any iPhone at hand to test iOS Chrome at the moment, but will try Safari in the iPhone simulator soon.

ios/RNSVGNode.m
msand: It is defined here: https://www.w3.org/TR/SVG11/text.html#FontSizeProperty
And the normative document is https://www.w3.org/TR/2008/REC-CSS2-20080411/fonts.html#propdef-font-size
and has this to say about it:
Initial value: medium
An <absolute-size> keyword refers to an entry in a table of font sizes computed and kept by the user agent. Possible values are:
[ xx-small | x-small | small | medium | large | x-large | xx-large ]

On a computer screen a scaling factor of 1.2 is suggested between adjacent indexes; if the 'medium' font is 12pt, the 'large' font could be 14.4pt. Different media may need different scaling factors. Also, the user agent should take the quality and availability of fonts into account when computing the table. The table may be different from one font family to another.

So, it should be read from the font data actually.

msand:
@magicismight Could you have a look asap? And merge or comment on anything that needs improvement before wider adoption/testing.

msand:
screen shot 2017-08-27 at 16 38 48

msand:
screen shot 2017-08-27 at 16 39 58

msand:
@magicismight The code is now defect free as far as I know. Ligatures, kerning, fontFeatureSettings, and, baseline alignment/shift; are missing from iOS, for it to be feature complete. But none of those are really mission critical, and can be filled in later on.

msand:
Ligatures and kerning are now working in iOS.
Don't know how to implement fontFeatureSettings, so will skip that.

AlignmentBaseline and BaselineShift behave strange in iOS, they are always nil, works correctly in android. I'm unsure how to proceed, I must be missing something simple/fundamental.

Otherwise, it is now essentially feature-complete and defect free in all the test-cases I have at the moment.

Any suggestions @magicismight ?

msand:
@magicismight Hey, have you had any time to look at this? :) Would love to get this to master and npm <3

magicismight:
Thanks for this awesome PR!
Merged into 6.0.0.
But there are some crashes need to be fixed.
I'll release a beta version after I fixed those crashed.

msand:
@magicismight I've done some work on improving cocoapods integration, but not sure if the changes make complete sense, could you check it out? https://github.com/msand/react-native-svg/tree/cocoapods

magicismight:
@msand I have removed PerformanceBezier code, added it as a submodule, and it works fine.
I'm not sure if add it as a cocoapods dependency will make the installation more complicated.
What do you think?

msand:
@magicismight I tried installing the latest commit using yarn, and it doesn't build in xcode, is there any way of getting npm/yarn to install git submodules?

msand:
The commit from my cocoapods branch should work without cocoapods as well, it just allows installing/including the lib in a xcode project using cocoapods as well, which makes it easier to set up continuous integration using e.g. fastlane and circleci. But when I changed how PerformanceBezier was included (directly instead of as a framwork, as I didn't get the original approach to work with cocoapods), it started complaining on some lines of code being incompatible with the ARC (automatic reference counting), and so I googled around and applied the suggested fixes (essentially just remove the lines of code), and thus I'm not certain if the memory is handled correctly anymore, or if all of it was even handled correctly according to obj-c best practices and if it actually frees all the relevant memory. Have you looked at the memory behaviour of the current code?

rh389:
@msand I think `alignmentBaseline` and `baselineShift` are nil on iOS because they ought to be coming from the `GlyphContext` but they're not made available. If I break on [this line](https://github.com/react-native-community/react-native-svg/blob/04887af278c49a0fa3b7956d2992fb2e71efd279/ios/Text/RNSVGText.m#L101) I can see that they're set (on `self`) but obviously not pushed in to the context.

Does that seem right to you? It's the first time I've dived in to this library but if that sounds like the correct solution I'll work on a PR.

msand:
@rh389 Yeah, they could be stored in the FontData structure by passing them to the glyphcontext, currently on android there are two helpers (missing from ios I notice now) handling the getting instead: https://github.com/react-native-community/react-native-svg/blob/04887af278c49a0fa3b7956d2992fb2e71efd279/android/src/main/java/com/horcrux/svg/TextShadowNode.java#L139-L176 
The spec claims the property is not inherited: https://www.w3.org/TR/css-inline-3/#propdef-alignment-baseline

The baseline-shift can have a value of length and percentage as well, and thus need to take the correct coordinate system transformations into consideration, thus cannot have simple enum parsing.
https://github.com/react-native-community/react-native-svg/blob/04887af278c49a0fa3b7956d2992fb2e71efd279/lib/props.js#L224-L262

alignment-baseline is parsed as an enum on android at least https://github.com/react-native-community/react-native-svg/blob/04887af278c49a0fa3b7956d2992fb2e71efd279/android/src/main/java/com/horcrux/svg/TextShadowNode.java#L52

The glyphcontext has a set of stacks, one for each property (stored as an array of input values, as they are glyph specific) which are set for the current element and any ancestors. 
It also has a set of indices which keeps track of which input array is used in what glyph context (dom element), and another array which keeps track of what indices into these arrays are used (the spec says it should increment these for all ancestors, not only the current element, for each glyph/character)

You wont need to touch those, just pass along the alignment values to the fontdata and store them there, then in tspan getLinePath you get it from the fontdata instead of self.

msand:
@rh389 btw, the java version of glyphcontext actually has comments: https://github.com/react-native-community/react-native-svg/blob/04887af278c49a0fa3b7956d2992fb2e71efd279/android/src/main/java/com/horcrux/svg/GlyphContext.java

msand:
@rh389 I made a pull request with some fixes, including baseline alignment in ios and a commented and cleaned up glyphcontext (to align with android), can you try it out? https://github.com/react-native-community/react-native-svg/pull/527

rh389:
@msand wow, I was just about to start work on this thanks to your advice, but it looks like there's no need! Your PR is working nicely for me, using `alignmentBaseline` on a bezier. I don't have anything set up to test exhaustively but it looks good to me (good idea porting those comments across to iOS too - `GlyphContext.m` looked a bit impenetrable before!).

msand:
@rh389 BaselineShift wasn't working correctly, so I amended the pull request with another commit. Now everything seems to be working fine.
https://github.com/react-native-community/react-native-svg/pull/527/commits/c8a2922ce7b81d257f7d0a76bd074ba1bf6c11a1

Issue #401, opened 2017-07-14
Error: Can't resolve Rect.js ( in react-native-svg ) 


ev1stensberg:
I'm getting watchman error when trying to use one of the modules. Running on Google Pixel XL remote. Any suggestions on why this is happening? Everything is properly set up!

lionel95200x:
Same error @ev1stensberg Have you found a solution ?

ev1stensberg:
I can try to debug it if its an pressing issue?

lionel95200x:
Yes if you can

dguillamot:
Also seeing this issue atm

dguillamot:
It looks like it is from importing createReactNativeComponentClass

`import createReactNativeComponentClass from 'react-native/Libraries/Renderer/shims/createReactNativeComponentClass';`

dguillamot:
Ah, I believe it is an issue with react-native-svg-uri using older react-native-svg.

mikearche:
Go to `node_modules/react-native-svg-uri/node_modules/react-native-svg/elements/Rect.js`

Edit line 3 to
`import createReactNativeComponentClass from 'react-native/Libraries/Renderer/src/renderers/native/createReactNativeComponentClass';`

I had to restart the packager afterwards and run
`npm start --reset-cache`

SupriyaKalghatgi:
It is not just with `Rect.js`
`createReactNativeComponentClass` error is thrown in all files of `elements` folder of react-native-svg

SupriyaKalghatgi:
Replace the `createReactNativeComponentClass` import statement to 
`import createReactNativeComponentClass from 'react-native/Libraries/Renderer/shims/createReactNativeComponentClass.js';`

magicismight:
```
npm ls react-native
npm ls react
npm ls react-native-svg
```

Make sure module versions are matched to [this table](https://github.com/react-native-community/react-native-svg#notice).
If not, remove the `^` before the version name in package.json and install.

Like this:
```
	"dependencies": {
		"react": "^16.0.0-alpha.12",
		"react-native": "^0.46.4",
		"react-native-svg": "^5.3.0"
	},
```

====> 
```
	"dependencies": {
		"react": "16.0.0-alpha.12",
		"react-native": "0.46.4",
		"react-native-svg": "5.3.0"
	},
```

developer239:
[...]

PR #431, opened 2017-08-14
Fix Error: Can't resolve Rect.js ( in react-native-svg ) #401

agungjk:
I changed 

> import createReactNativeComponentClass from 'react-native/Libraries/Renderer/shims/createReactNativeComponentClass.js';
to
> import createReactNativeComponentClass from 'react-native/Libraries/Renderer/src/renderers/native/createReactNativeComponentClass.js';


robertpcontreras:
What version of react-native are you using? Updating to the latest version should fix the error you were having. After updating to 0.47.2 I no longer got this error. This PR is not required.

dustinsavery:
Confirmed, using the latest react-native doesn't require this fix. Closing this PR, will re-open if needed.

PR #436, opened 2017-08-17
fix Image render postion bug on iOS

soulwu:
if render a Image with x or y set to a nonzero value and imageRatio equal to rectRatio, the Image will be rendered at position of [2x, 2y]


dustinsavery:
@soulwu Is this still needed?

dustinsavery:
Closing this PR for now since it appears to be inactive. Feel free to re-open if you want to work on it more

PR #450, opened 2017-09-08
Added documentation for fillRule property

kkjdaniel:
Very useful property that I didn't even know existed until I went digging for it, so might as well save some people the headache and just document it in the README.


magicismight:
Thanks for this PR.

PR #493, opened 2017-11-01
Add TypeScript type definitions

aeirola:
TypeScript definitions constructed from element prop types.

### DIFF
--- a/index.d.ts
+++ b/index.d.ts
@@ -1,0 +1,232 @@
+import * as React from 'react';
+import * as ReactNative from 'react-native';
+
+// Common props
+type NumberProp = string | number;
+
+interface TouchableProps {
+  disabled?: boolean,
+  onPress?: (event: any) => any,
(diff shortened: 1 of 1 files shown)


dustinsavery:
Why not add this to https://github.com/DefinitelyTyped/DefinitelyTyped?

aeirola:
@dustinsavery Of course possible to have the TypeScript definitions separately in DefinitelyTyped, and many packages do this. But this makes the type definitions a bit harder to maintained, since they are removed from the actual code. DefinitelyTyped also [recommends](https://github.com/DefinitelyTyped/DefinitelyTyped#create-a-new-package) including the type definitions within the library.

That being said, if you prefer not to include the type definitions within this library, I'll gladly add the to DefinitelyTyped instead.

dustinsavery:
@aeirola wasn't aware of their recommendation. Thanks for letting me know 👏

PR #497, opened 2017-11-07
6.0.0

magicismight:
@msand I had removed Quartz and PerformanceBezier from this repo.
Will run a postinstall script to download them from Github.
And finally this will be released as `6.0.0`.


magicismight:
@msand You can try to install this in 6.0.0-rc10

PR #501, opened 2017-11-09
Update "color" dep to 2.x to fix RN minifier bug

ide:
There's a bug with the version of UglifyJS that RN/Metro uses which causes an error about "e.length" inside of "color" that was fixed in 2.x. One change to "color" is that `rgbaArray()` was removed and now the call is `rgb().array()`.


magicismight:
Thanks for this PR.

PR #507, opened 2017-11-14
Implement correct transform parser

msand:
Support for transform lists with repeated occurrences of one type of transform.
Support for optional arguments.

https://www.w3.org/TR/SVG/coords.html#TransformAttribute
https://developer.mozilla.org/en-US/docs/Web/SVG/Attribute/transform

### DIFF
--- a/lib/extract/extractTransform.js
+++ b/lib/extract/extractTransform.js
@@ -1,6 +1,7 @@
 import Matrix2D from '../Matrix2D';
 import _ from 'lodash';
 let pooledMatrix = new Matrix2D();
+import peg from 'pegjs';
 
 function transformToMatrix(props, transform) {
     pooledMatrix.reset();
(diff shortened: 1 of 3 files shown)


lib/extract/extractTransform.js
```diff
@@ -13,71 +14,172 @@ function transformToMatrix(props, transform) {
...
+                const [a, c, e, b, d, f] = transformParser.parse(transform);
+                pooledMatrix.append(...[a, b, c, d, e, f]);
+            } catch (e) {
+                console.error(e);
             }
-        }
-        if (typeof transform !== 'string') {
+        } else {
             pooledMatrix
                 .appendTransform(
```
msand: This part, props2transform, transformToMatrix & universal2axis should probably be removed. They're not part of the svg spec and a source of confusion and bugs, as the order of transforms being applied isn't documented nor specified anywhere except the source code of the implementation, and is quite arbitrary. What do you think @magicismight is such a big breaking change possible? To align it with the evergreen browsers?

msand:
@dustinsavery What do you think of this? It implements a parser for the grammar straight from the spec, allowing arbitrary combinations of transforms and specifying the order, instead of only one per type as currently. It also gives errors for malformed syntax, such that incorrect values aren't swallowed silently:

https://www.w3.org/TR/SVG/coords.html#TransformAttribute
The following is the Backus-Naur Form (BNF) for values for the ‘transform’ attribute. The following notation is used:

    *: 0 or more
    +: 1 or more
    ?: 0 or 1
    (): grouping
    |: separates alternatives
    double quotes surround literals

transform-list:
    wsp* transforms? wsp*
transforms:
    transform
    | transform comma-wsp+ transforms
transform:
    matrix
    | translate
    | scale
    | rotate
    | skewX
    | skewY
matrix:
    "matrix" wsp* "(" wsp*
       number comma-wsp
       number comma-wsp
       number comma-wsp
       number comma-wsp
       number comma-wsp
       number wsp* ")"
translate:
    "translate" wsp* "(" wsp* number ( comma-wsp number )? wsp* ")"
scale:
    "scale" wsp* "(" wsp* number ( comma-wsp number )? wsp* ")"
rotate:
    "rotate" wsp* "(" wsp* number ( comma-wsp number comma-wsp number )? wsp* ")"
skewX:
    "skewX" wsp* "(" wsp* number wsp* ")"
skewY:
    "skewY" wsp* "(" wsp* number wsp* ")"
number:
    sign? integer-constant
    | sign? floating-point-constant
comma-wsp:
    (wsp+ comma? wsp*) | (comma wsp*)
comma:
    ","
integer-constant:
    digit-sequence
floating-point-constant:
    fractional-constant exponent?
    | digit-sequence exponent
fractional-constant:
    digit-sequence? "." digit-sequence
    | digit-sequence "."
exponent:
    ( "e" | "E" ) sign? digit-sequence
sign:
    "+" | "-"
digit-sequence:
    digit
    | digit digit-sequence
digit:
    "0" | "1" | "2" | "3" | "4" | "5" | "6" | "7" | "8" | "9"
wsp:
    (#x20 | #x9 | #xD | #xA)

dustinsavery:
Makes sense to me. I'm on board

dustinsavery:
@msand if you can resolve the conflict, let's get this merged in before we create `6.0.1-rc.1`

msand:
@dustinsavery I rebased onto master now, should be clear for merging.

PR #509, opened 2017-11-14
Fix installing via cocoapods

akantsevoi:
Fix compile issue: https://github.com/react-native-community/react-native-svg/issues/484
in case install via cocoapods


dustinsavery:
Adding the path to Header Search Paths would solve this without the need to change it to a framework import.

PR #524, opened 2017-11-22
[FIX] colors array contains null when alpha = 1

Tennen:
the current way to get solid color array may lead to an error that native module receives an array which contains null when alpha = 1.


dustinsavery:
@Tennen Not sure I understand, why would it jump from a length of 3 => 5? If you added the alpha, that'd only be 4, yeah?

dantman:
This looks like the same kind of change as #512, I kind of like the way #512 fixes this with array destructuring instead of mutating the original array.

Tennen:
@dustinsavery ah..yes, it will be length 4, i was thinking after adding the 0 it will be length 5
@dantman thanks, I didn't find #512 yesterday, I think its better to use that way. close this PR after #512 is merged

dustinsavery:
#512 has been merged, closing this PR.

PR #527, opened 2017-11-26
Implement getAlignmentBaseline and getBaselineShift in ios.

msand:
Add comments to glyphcontext and align with android.
Implement correct isWordSeparator predicate.
Cleanup GlyphContext j2objc remnants.


henrychen11:
Approved

PR #534, opened 2017-12-05
Fix crash when passing wrong type to percentage prop on iOS

terribleben:
`RNSVGPercentageConverter::isPercentage:` method can hard crash if somebody uses a type which doesn't respond to `length`, e.g. something like this: `[__NSArrayM length]: unrecognized selector sent to instance 0x61800924baf0`. Instead, just return `NO`, this is not a percentage.

### DIFF
--- a/ios/Utils/RNSVGPercentageConverter.m
+++ b/ios/Utils/RNSVGPercentageConverter.m
@@ -47,6 +47,9 @@ static NSRegularExpression* percentageRegExp;
 
 + (BOOL)isPercentage:(NSString *) string
 {
+    if (![string isKindOfClass:[NSString class]]) {
+      return NO;
+    }
     return [percentageRegExp firstMatchInString:string options:0 range:NSMakeRange(0, [string length])] != nil;
 }
 


ios/Utils/RNSVGPercentageConverter.m
```diff
@@ -47,6 +47,9 @@ + (CGFloat)percentageToFloat:(NSString *)percentage relative:(CGFloat)relative o
 
 + (BOOL)isPercentage:(NSString *) string
 {
+    if (![string isKindOfClass:[NSString class]]) {
```
dustinsavery: The Param is already being cast as a string, this would always be false, yeah?

terribleben: No, `isKindOfClass` checks the actual class of the receiving object at runtime rather than examining the type of the pointer. So this would catch the following case:
```objc
id notAString = @(1234);
BOOL result = [RNSVGPercentageConverter isPercentage:notAString];
```

terribleben: fwiw, we are seeing a nontrivial number of crashes in Expo as a result of people passing the wrong type here, so that's how we caught this and that was the motivation for this PR.

PR #547, opened 2017-12-18
Ammending the Android install instructions to exclude dependencies

dustinsavery:


msand:
I've removed the appcompat dependency completely now, so this won't be needed anymore. Closing for now.

PR #575, opened 2018-01-17
fix an "out of range" error

glacjay:
The half of a char in java has a range of [0, 255], so the array's length should be 256.


msand:
@glacjay Looks good. Thanks for this!

PR #577, opened 2018-01-18
Add exports for prop types, additional type fixes.

simonbuchan:
## Changes

Adds exports for all the type definitions (except `NumberProp`)

Also:
- Extracts and exports the string literal union types
- Adds or renames prop types such that every `Foo`
  component export also has a `FooProps`
- Adds non-default export of `Svg`, so `import * as Svg ...`
  is usable.
- Makes `RadialGradientProps.gradientUnits` optional, like
  for `LinearGradientProps.gradientUnits`.
- Adds stronger type for `TransformProps.transform` based
  on code in `extractTranform.js` (please double-check!)
- Fixes typo in `DefinationProps` name.
- A handful of minor formatting normalizations.

I also found and put a type fix for the following issues:
- `Symbols` -> `Symbol` (!)
- `Use` errors unless `width` and `height` are provided, and strings? This seems like a bug.
- `Svg` errors if `width` and `height` are not provided? This one looks like it comes from RN, since it talks about `Surface`?
- `Definitions` have an `id`, not a `name`.
- There is no `Shape` export.

## Example

```ts
import React from "react";
import * as Svg from "react-native-svg";

const defaultStrokeProps: Svg.StrokeProps = {
  stroke: 'black',
  strokeLinejoin: 'round',
};
export interface ExampleProps {
  strokeProps?: Svg.StrokeProps;
}
export default ({ strokeProps = {} }: ExampleProps) => (
  <Svg.Svg width={50} height={50}>
    <Svg.Rect x={0} y={0} width={50} height={50} {...defaultStrokeProps} {...strokeProps} />
  </Svg.Svg>
);
```

In particular, note that without a type, `defaultStrokeProps` will type `strokeLinejoin` as `string`, which will cause a type error when spreading.

## Questions

1. I could make a precise type for `PreserveAspectRatio`, but it expands out pretty bad: 40 different combinations. Since it's fairly likely this could be computed (e.g. `x${alignX}Y${alignY}`), I'm not sure if this would be better? On the other hand, it's a very easy to typo value written literally!
2. Should `NumberProp` be exported? It is trivial but so commonly used simply having an authoritative name could be useful. Perhaps it could change to use https://github.com/Microsoft/TypeScript/issues/6579 later?
3. Feedback on naming of the enum types would be good: in particular,:
  - Should they have a `Prop` suffix like `NumberProp`? Feels too confusable with `Props` to me.
  - `FillRule` is used for `FillProps.fillRule` and `ClipProps.clipRule`, but I'm not sure what a better name would be?
4. In my `react-dom` SVG rendering code, I currently render defs without `id` then use `React.cloneElement(def, { key, id })` to share generating ids for these: would it be OK to make these optional?

### DIFF
--- a/index.d.ts
+++ b/index.d.ts
@@ -205,8 +205,8 @@ export const Stop: React.ComponentClass<StopProps>;
 
 export interface SvgProps extends ReactNative.ViewProperties {
   opacity?: NumberProp,
-  width?: NumberProp,
-  height?: NumberProp,
+  width: NumberProp,
+  height: NumberProp,
   viewBox?: string,
   preserveAspectRatio?: string,
 }
@@ -215,12 +215,12 @@ export interface SvgProps extends ReactNative.ViewProperties {
 export const Svg: React.ComponentClass<SvgProps>;
 export default Svg;
 
-export interface SymbolsProps {
+export interface SymbolProps {
   id: string,
   viewBox?: string,
   preserveAspectRatio?: string,
 }
-export const Symbols: React.ComponentClass<SymbolsProps>;
+export const Symbol: React.ComponentClass<SymbolProps>;
 
 export interface TSpanProps extends CommonPathProps, FontProps {
   dx?: NumberProp,
@@ -244,7 +244,7 @@ export const TextPath: React.ComponentClass<TextPathProps>;
 
 export interface UseProps extends CommonPathProps {
   href: string,
-  width?: NumberProp,
-  height?: NumberProp,
+  width?: string,
+  height?: string,
 }
 export const Use: React.ComponentClass<UseProps>;
--- a/index.d.ts
+++ b/index.d.ts
@@ -6,7 +6,65 @@ type NumberProp = string | number;
 
 export type FillRule = 'evenodd' | 'nonzero';
 export type Units = 'userSpaceOnUse' | 'objectBoundingBox';
+
 export type TextAnchor = 'start' | 'middle' | 'end';
+export type FontStyle = 'normal' | 'italic' | 'oblique';
+export type FontVariant = 'normal' | 'small-caps';
+export type FontWeight =
+  | 'normal'
+  | 'bold'
+  | 'bolder'
+  | 'lighter'
+  | '100'
+  | '200'
+  | '300'
+  | '400'
+  | '500'
+  | '600'
+  | '700'
+  | '800'
+  | '900'
+  ;
+export type FontStretch =
+  | 'normal'
+  | 'wider'
+  | 'narrower'
+  | 'ultra-condensed'
+  | 'extra-condensed'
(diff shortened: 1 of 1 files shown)


msand:
@dustinsavery Could you review this? I don't have ts in use in any current project of mine. Or, @simonbuchan do you have a test repo where I could test the publicly exposed interface surface? It seems this doesn't require a semver major version change?

simonbuchan:
I could certainly cook up a test repo exercising this code: I don't thing my current use is particularly private, but it's weird enough a setup it wouldn't be terribly informative (it's sharing web and react-native code using a hacky haul config and babel 7 preset-typescript and all sorts of nonsense!) Would have to be after work in ~9hrs?

Apparently there is also [dtslint](https://github.com/Microsoft/dtslint#readme), though I've never used it.

The only changes that may cause additional type errors should be in the second commit, which were direct changes to things which redboxed me. That said, there is legal code that will not typecheck now, for example:

```ts
const svgProps: { width?: string, height?: string } = {};
svgProps.width = 200;
svgProps.height = 200;
return <Svg {...svgProps} />; // width is `string | undefined`, `undefined` cannot be converted to `string`.
```

My understanding is that such "type breaking changes" are to an extent accepted, even expected, by TS users so long as they can directly reduce runtime errors or provide editing assistance, and don't require too much refactoring to resolve the errors - obviously a case-by-case trade-off and will always have some controversy. For example: https://medium.com/the-vue-point/upcoming-typescript-changes-in-vue-2-5-e9bd7e2ecf08

simonbuchan:
Finally got around to a test repo: https://github.com/simonbuchan/ReactNativeSVGTestTypes

Incomplete, but already found some more (existing) issues, I will update above.

simonbuchan:
Looks like a lot of the `NumberProp` must actually be `string` issues are because https://github.com/react-native-community/react-native-svg/blob/master/lib/percentToFloat.js expects `RegExp.match()` to `[[ToString]]` it's arg, which it doesn't (anymore?) - it errors. Probably better to just fix that (in a different PR, I assume) than update the typings to reflect that.

Extracted a bunch of text/font details from props.js, those are pretty heavily prop-typed and it's nicer to get the errors early. Need to cover those too, so I'll look at beefing up the test app tomorrow, hopefully.

msand:
@simonbuchan Could you matching problem have been related to this? https://github.com/react-native-community/react-native-svg/pull/586/files

simonbuchan:
I'd have to double-check, but I recall it being a TypeError. Could be
either a remote debugger difference or core-js though.

simonbuchan:
Thanks for merging! I'll keep poking at this and see if I've missed anything, which seems likely. I'll also see if I can find out why I'm seeing those runtime errors....

PR #593, opened 2018-02-02
Add support for react-native-web

DanBeard:
Added support for [react-native-web](https://github.com/necolas/react-native-web) so that SVG components can be shared between iOS, Android, and web. 

I attempted to touch the existing code as little as possible, though there was some custom logic needed for the Svg and Image components. In order to keep the abstraction that 'web' is just another native platform I made "RNSVG*"  'native' components with HTML that mirror the iOS and Android native components. 

Manually tested with a react-native-web app and android.

### DIFF
--- a/elements/Circle.js
+++ b/elements/Circle.js
@@ -1,5 +1,5 @@
 import React from 'react';
-import createReactNativeComponentClass from '../lib/createReactNativeComponentClass';
+import { createReactNativeComponentClass }  from '../lib/nativeComponents';
 import Shape from './Shape';
 import {CircleAttributes} from '../lib/attributes';
 import {pathProps, numberProp} from '../lib/props';
--- a/elements/ClipPath.js
+++ b/elements/ClipPath.js
@@ -1,6 +1,6 @@
 import React, {Component} from 'react';
 import PropTypes from 'prop-types';
-import createReactNativeComponentClass from '../lib/createReactNativeComponentClass';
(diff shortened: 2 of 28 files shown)


msand:
@DanBeard Thank for the PR! Have you seen https://github.com/godaddy/svgs it does essentially the same thing, but handles some of the quirks with react-native-svg perhaps a bit better. Do you still think merging this pr is needed?

DanBeard:
:flushed: no I wish I saw that earlier. That would have really helped. Especially looking at how it's handling the transforms. 

I'm using react-native-web in my projects because I love that it treats 'web' as just another native platform like 'ios', 'android' , or 'windows'.  It lets it sharing between platforms work out-of-the-box for the most part. This is probably a better conversation  to have elsewhere, but it would be nice to move the svgs web functionality *into* react-native-svg using the *.web.js system that react-native-web uses, and then svgs could reference those files? That way it would work out-of-the-box with react-native-web.

msand:
A pull request for doing that would be awesome and very much appreciated! 👍

msand:
I'm wondering if we should re-open this? https://github.com/react-native-community/react-native-svg/issues/608

necolas:
FWIW, neither this PR nor `svgs` will work as they don't support the React Native style API on the elements rendered. The exported components would need to use `react-native-web`'s `createElement` helper, as described in this post http://nicolasgallagher.com/making-svg-icon-libraries-for-react-apps/

PR #596, opened 2018-02-06
Performance optimisations and support for nesting Svg and View inside Svg.

msand:


koxen:
👀

AntonNixonNixon:
@msand First of, thanks for the work, really appreciated! Now, I've been playing around with these changes and noticed that it's possible to add Text into the nested View on iOs and that it works as expected but that I can't get it to show on Android, so my question is then, is this intentional or a bug of sorts? Or are you actually not supposed to add any elements inside the View at all?

sibelius:
@AntonNixonNixon this is a limitation of android on react native core

I wanna to come up with a lint that detects this case, so we can avoid this bugs on android

msand:
@AntonNixonNixon  @sibelius I have an attempt for supporting nested native text views on android now, can you test it and see if it behaves according to your expectations? https://github.com/react-native-community/react-native-svg/pull/658

msand:
I believe  @sibelius is thinking of nesting native text inside images, which isn't supported on android. Native text and views inside svg are possible, we just need to define how it should work. The current implementation probably doesn't even make sense, the CTM should probably propagate from svg to rn views.

vivek12345:
@msand is the release with this PR done because I am still facing issues setting width to 100%

SaeedZhiany:
@msand can you please provide an example of nesting svg? I not sure how to use it and I can not find any proper example on internet.  every nested svg I had create gives me `unknown bitmap configuration` error

PR #609, opened 2018-02-17
Namespaces or scopes a handful more global iOS symbols

terribleben:
This is a followup to https://github.com/react-native-community/react-native-svg/pull/604. In the process of cutting the latest Expo SDK, we found a handful more things that needed to be namespaced for RN-SVG to avoid interacting with other versions of itself.


msand:
@terribleben Thanks for this! Published a new patch version.

PR #627, opened 2018-03-14
Bugfix/278 memory leak issue

nikDemyankov:
Fix for the https://github.com/react-native-community/react-native-svg/issues/278.

The problem was in a few retain cycles, that kept objects in the memory even if the view has been removed from the stack. I hope that I've found all of them... Tested it with the images that we use in the app (some are simple, some are pretty complex), seems good now: memory is cleaned up. Of course, there are few Managers that are saved - but they are small and as I understand - for optimisation purpose. And they are not re-created each time svg is rendered.

Please, check out the PR.

### DIFF
--- a/ios/Elements/RNSVGClipPath.m
+++ b/ios/Elements/RNSVGClipPath.m
@@ -17,7 +17,7 @@
 
 - (void)parseReference
 {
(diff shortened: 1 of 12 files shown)


msand:
Thanks for this! Merged and released new version 6.3.0

PR #630, opened 2018-03-15
5.5.1 Load Image from URL

nadude:
Tested on Android & iOS

PR #632, opened 2018-03-19
6.3.0 [iOS] Fix Load Image from URL

nadude:
Ref #630 

Quick patch to fix after 6.3.0 merge.


msand:
Merged and released patch version.

PR #651, opened 2018-04-14
Add support for `onLayout` events

maxkomarychev:
you can use playground project for experiments: https://github.com/maxkomarychev/svgplayground

It's now possible to use `onLayout` prop on elements to get coordinates of the elements in coordinate space of a view.

I understand the algo is slightly unoptimal because it requires to recalculate matrixes backwards. If you have an idea how to get reverse transform with single operation I'd happily improve it, at the same time I didn't notice big performance issues.

(red rectangles are simple react `View` with coordinates obtained through new prop).
simulator screen shot - iphone 6 - 2018-04-14 at 20 00 12


maxkomarychev:
thanks @sibelius !
where can we expect this to be merged into upstream?

sibelius:
@msand can we get this merged in?

msand:
Sure, I can take a look perhaps on monday evening and merge the open PRs. I've changed jobs to a startup I cofounded, and I'm not currently working on any react native code base, except some hobby project experiments.  So, we should probably try to find more maintainers for now as well, as I won't have much time for much testing nor further development right now. @dustinsavery @brentvatne Do you have anyone/time available?

android/src/main/java/com/horcrux/svg/GroupShadowNode.java
```diff
@@ -79,6 +82,7 @@ public void run(ReactShadowNode lNode) {
 
                     int count = node.saveAndSetupCanvas(canvas);
                     node.draw(canvas, paint, opacity * mOpacity);
+                    rects.add(node.getClientRect());
```
msand: Seems the unions could be done once here, instead of looping twice.

android/src/main/java/com/horcrux/svg/RenderableShadowNode.java
```diff
@@ -204,6 +208,12 @@ public void draw(Canvas canvas, Paint paint, float opacity) {
                 mPath.setFillType(mFillRule);
             }
 
+            RectF r = new RectF();
+            mPath.computeBounds(r, true);
+            Matrix m = this.getSvgShadowNode().getReverseTransform();
```
msand: The canvas/context CTM should be possible to use instead of computing the reverse transform many times.

ios/Elements/RNSVGGroup.m
```diff
@@ -46,6 +48,9 @@ - (void)renderGroupTo:(CGContextRef)context rect:(CGRect)rect
             }
 
             [svgNode renderTo:context rect:rect];
+            
+            CGRect nodeRect = svgNode.clientRect;
+            [rects addObject:[NSValue valueWithCGRect:nodeRect]];
```
msand: Seems the unions could be done once here, instead of looping twice. (Same as android)

ios/RNSVGRenderable.m
```diff
@@ -187,6 +189,9 @@ - (void)renderLayerTo:(CGContextRef)context rect:(CGRect)rect
         self.path = CGPathRetain(CFAutorelease(CGPathCreateCopy([self getPath:context])));
         [self setHitArea:self.path];
     }
+    
+    const CGRect pathBouding = CGPathGetBoundingBox(self.path);
+    self.clientRect = CGRectApplyAffineTransform(pathBouding, [self.svgView getReverseTransform]);
```
msand: The CGContext already has the CTM no need to recompute over and over.

maxkomarychev: That was my initial thought.

`self.clientRect = CGRectApplyAffineTransform(pathBouding, CGAffineTransformInvert(CGContextGetCTM(context)));`

what can be easier, huh?
But no, that's what it gives:

simulator screen shot - iphone 6s - 2018-06-07 at 23 25 22

instead of

simulator screen shot - iphone 6s - 2018-06-07 at 23 26 06

It is very much appreciated if you can suggest better formula for properly converting coordinate from svg space to view

maxkomarychev: One more thing is to be aware is that initial ctm for the context *is not* identity matrix.
I also tried to save that initial CTM and multiply it on the left final ctm.

sort of: `reversetransform = invert(invert(initialTransform) * currentCTM)` and variations of that

maxkomarychev: `viewTransform = initialCTM * viewBoxCTM * nodeTransform_1 * ... * nodeTransform_n`

I thought I need `invert(viewBoxCTM * nodeTransform_1 * ... * nodeTransform_n)`

I tried `reverseTransform = invert(invert(initialCTM) * initialCTM * viewBoxCTM * .... *)`

am I missing something in how matrix multiplications work?

android/src/main/java/com/horcrux/svg/RenderableShadowNode.java
msand: I think it should be fine to use getMatrix, as we're creating the canvas, and it's never passed outside library code.

ios/RNSVGRenderable.m
msand: Hmm, I don't see you inverting the transforms in the current code, can it be that?

maxkomarychev: That's the point, I couldn't achieve desired behavior with matrix operations. The final code you see was rather a wild guess.

maxkomarychev: as I said I was trying to distill view coordinates through inversions, however the *reverese* multiplication of matrices worked, not the *inversion*

msand: Or, if the initial CTM is not the identity matrix, then you might want invert(initial) * CTM, as if CTM = initial * viewbox * elementTMs then inverting the initial one and multiplying from the same direction should get it (close to, still floats involved) identity.

maxkomarychev: I tried.

`reverse = ^(^initial * ctm) = ^(^initial * initial * vb * elements) = ^(identity * vb * elements) = ^(vb * elements)`

the code below:

```
    const CGAffineTransform ctm = CGContextGetCTM(context);
    const CGAffineTransform invertIntial = CGAffineTransformInvert(self.svgView.initialTransform);
    self.clientRect = CGRectApplyAffineTransform(pathBouding, CGAffineTransformInvert(CGAffineTransformConcat(invertIntial, ctm)));
```

doesn't work

msand: Well, matrix multiplication is associative, should it shouldn't matter in what order you do the multiplication (reverse, in order, or, some order which minimises the number of scalar multiplications needed, although with same size matrices that doesn't even vary). So (A*B)*C=A*(B*C) but it's not commutative, so in general A*B != B*A

maxkomarychev: ok actually,
this:
```
    const CGAffineTransform ctm = CGContextGetCTM(context);
    const CGAffineTransform invertIntial = CGAffineTransformInvert(self.svgView.initialTransform);
    self.clientRect = CGRectApplyAffineTransform(pathBouding, CGAffineTransformConcat(ctm, invertIntial));
```
works..

but its rather

`reverse = ctm * ^initial`

how can this be right?

msand: Seems you have an extra invert there, how about
```
    const CGAffineTransform ctm = CGContextGetCTM(context);
    const CGAffineTransform invertIntial = CGAffineTransformInvert(self.svgView.initialTransform);
    self.clientRect = CGRectApplyAffineTransform(pathBouding, CGAffineTransformConcat(invertIntial, ctm));
```

maxkomarychev: oh wow... damn it!
from docs of `CGContextConcatCTM`:

> CTMnew = transform * CTMcontext.

so basically matrices are added on the left!

`ctm = element_n * ... * element_1 * vb * initial`

then `reverse = ctm * ^initial = element_n * ... * element_1 * vb * initial * ^initial = element_n * ... * element_1 * vb * identity = element_n * ... * element_1 * vb`

but there's still no *inversion* of pure transform

I'm confused %)

I'll update PR later, thank you very much for your comments.

msand: Or can the initial transform be due to the viewbox & align?

msand: Oh great, thanks for your work as well! Seems great to have onLayout available.

maxkomarychev:
@msand updated PR, ready for review

msand:
Looks good, but is the inverse initial transform really needed on android? We're creating the canvas ourselves, and it should have the identity as the initial transform, no?

maxkomarychev:
@msand, I didn't really check value of CTM on android, just translated code from iOS.

Will only be able to look into it next wee.

maxkomarychev:
@msand initial ctm on android is identity matrix indeed, so we don't need to keep neither initial ctm nor invert, just take current ctm and convert rect with it.

ready for review

msand:
Looks good, haven't tested yet though. Also been thinking it would be good to expose https://www.w3.org/TR/SVG11/types.html#InterfaceSVGLocatable
```
interface SVGLocatable {

  readonly attribute SVGElement nearestViewportElement;
  readonly attribute SVGElement farthestViewportElement;

  SVGRect getBBox();
  SVGMatrix getCTM();
  SVGMatrix getScreenCTM();
  SVGMatrix getTransformToElement(in SVGElement element) raises(SVGException);
};
```
But, as this is react-native, perhaps onLayout is sufficient, as e.g. getBBox must return the actual bounding box at the time the method was called, even in case the element has not yet been rendered.
Will merge and test before releasing.

msand:
@maxkomarychev I've added getBBox, getCTM, getScreenCTM, getTotalLength, getPointAtLength and isPointInFill to the develop branch now. Would you be available to help testing?
I have a PR open here: #1135

PR #652, opened 2018-04-17
chore(PropTypes): Fix proptypes for new RN

bayukp:
`Image.propTypes.source` is not available in newer RN.
Change to `PropTypes.object`

### DIFF
--- a/elements/Image.js
+++ b/elements/Image.js
@@ -1,7 +1,7 @@
 import React from "react";
 import PropTypes from "prop-types";
 import { Image } from "react-native";
-import { requireNativeComponent, ImagePropTy } from "react-native";
+import { requireNativeComponent } from "react-native";
 import { ImageAttributes } from "../lib/attributes";
 import { numberProp, touchableProps, responderProps } from "../lib/props";
 import Shape from "./Shape";
--- a/elements/Image.js
+++ b/elements/Image.js
@@ -1,7 +1,7 @@
 import React from "react";
 import PropTypes from "prop-types";
 import { Image } from "react-native";
-import { requireNativeComponent } from "react-native";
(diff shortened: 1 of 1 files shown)


elements/Image.js
```diff
@@ -19,7 +19,7 @@ export default class extends Shape {
         y: numberProp,
         width: numberProp.isRequired,
         height: numberProp.isRequired,
-        href: Image.propTypes.source,
+        href: PropTypes.object,
```
maxkomarychev: it's not very accurate, source can also be a `number` when image is `require`d
https://github.com/facebook/react-native/blob/master/Libraries/Image/ImageSourcePropType.js#L82-L88

const ImageSourcePropType = PropTypes.oneOfType([
  ImageURISourcePropType,
  // Opaque type returned by require('./image.jpg')
  PropTypes.number,
  // Multiple sources
  PropTypes.arrayOf(ImageURISourcePropType),
]);

msand:
Hmm, I'm not having this problem, are you sure this is an issue?

bayukp:
@msand build with js prod enable will crash your app.

elements/Image.js
bayukp: will update accordingly.

maxkomarychev:
regarding original issue with prod build: @bayukp what is your react-native?
Also there's actually special [`ImageSourcePropType`](https://github.com/facebook/react-native/blob/master/Libraries/Image/ImageSourcePropType.js) can be imported from `react-native`

bayukp:
@maxkomarychev I'm using 0.54.

Yes i will update using that one.

But since i doesn't have access to my repo anymore, i will create separate PR.

bayukp:
move to https://github.com/react-native-community/react-native-svg/pull/722

PR #658, opened 2018-04-21
Android nested native text

msand:


msand:
Any interest in this? Otherwise I guess this could be closed.

PR #664, opened 2018-04-25
Try render when fetched bitmap

akihiro0228:
# issues
https://github.com/react-native-community/react-native-svg/issues/662


akihiro0228:
As a result of trial, I canceled pull-req because problem was not fixed.

PR #711, opened 2018-06-13
fix issue that not show transparent on android

xinyer:
set fill default value to transparent to fix the issue that can not show transparent background on Android.

### DIFF
--- a/lib/extract/extractFill.js
+++ b/lib/extract/extractFill.js
@@ -18,7 +18,7 @@ export default function(props, styleProperties) {
 
     return {
         // default fill is black
-        fill: extractBrush(props.fill || "#000"),
+        fill: extractBrush(props.fill || "rgba(0, 0, 0, 0)"),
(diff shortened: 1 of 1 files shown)


kevinjom:
Can we get this merged, I am having the same issue

Amosel:
Why not both?

msand:
Well, the spec https://www.w3.org/TR/SVG11/painting.html#FillProperties says that the default fill is black and the opacity is 1 so this seems like it'll probably break expectations. Do you have a some minimal test available which behaves differently from how e.g. browsers would render it?

msand:
Can you try the latest version? I think I've fixed the real issue. WritableArray was reporting zero for non-empty arrays, when merging properties from ancestors more than one descendant deep in the tree. Rewrote it using ArrayList<String> instead.

msand:
As no new information has arrived I'll assume this has been fixed. Closing.

PR #718, opened 2018-06-18
remove opacity from interface

toytonic:
property is no longer needed in the extended interface


toytonic:
duplicate of https://github.com/react-native-community/react-native-svg/pull/710

PR #722, opened 2018-06-26
chore(Image): Update image prop type to ImageSourcePropType

bayukp:
follow-through this https://github.com/react-native-community/react-native-svg/pull/652

Since i don't have access to old repo, i re-fork this repo.

In RN 0.54 if we create prod build or set JS-DEV = false in dev settings the app will throw error because can't find `Image.propTypes.source`.

### DIFF
--- a/elements/Image.js
+++ b/elements/Image.js
@@ -1,7 +1,7 @@
 import React from "react";
 import PropTypes from "prop-types";
-import { Image } from "react-native";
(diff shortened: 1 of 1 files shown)


bunnyc1986:
can we merge this?

bunnyc1986:
ping.

msand:
I would prefer not importing from internals, only from the main exports, but guess we can merge this for now.

PR #727, opened 2018-07-04
Fixed NullPointerException

ExoMemphiz:
Sometimes a NullPointerException occurs when using the computeBounds function, this fixes these by simply adding additional null pointer checks to 2 specific places.

I posted about this as an issue, but thought it'd be faster to create a Pull Request.

Best Regards


msand:
v6.5.0 published @ExoMemphiz

PR #729, opened 2018-07-05
Upgrade SDK Build Tools for react-native 0.56.0

rghorbani:
Making it compatible with react-native 0.56.0

Still can use
ext {
    buildToolsVersion = "23.0.1"
    minSdkVersion = 16
    compileSdkVersion = 23
    targetSdkVersion = 23
    supportLibVersion = "23.0.1"
}
if you are using older tools.

PR #733, opened 2018-07-16
Updated gradle & android project to use variables from root project if present

XenorPLxx:
I've used changes made to in https://github.com/react-native-community/react-native-camera/blob/master/android/build.gradle and accepted in https://github.com/corymsmith/react-native-fabric/pull/177 as a reference and updated android project so it builds on the React Native 0.56 without problems and it also uses global project variables if those are present.


rghorbani:
Hi @XenorPLxx 
Already created the same commit in #729.

msand:
Merged #729 and published v6.5.0

PR #736, opened 2018-07-19
Fix memory leak: remove references to SvgViews in onDetachedFromWindow

baconz:
It looks like onDropViewInstance does not get called when the entire react
instance is shut down from the native side. To gaurantee that we do not
leak memory, we need to remove the reference to the view in onDetachedFromWindow().


msand:
v6.5.0 published @baconz

PR #739, opened 2018-07-24
improved SDK values -use SDK Version values from root project

yeomann:
Instead of assuming the `compileSdkVersion `, `targetSdkVersion`, etc, we can make it dynamic by reading it from the root project.

Android Target API Level 26 will be required in August 2018.
https://android-developers.googleblog.com/2017/12/improving-app-security-and-performance.html

Therefore I wrote target values of 26 instead of 23

And the React Native team is already working on this:
https://github.com/facebook/react-native/issues/18095
https://github.com/facebook/react-native/pull/17741

### DIFF
--- a/android/build.gradle
+++ b/android/build.gradle
@@ -11,15 +11,19 @@ buildscript {
 
 apply plugin: 'com.android.library'
 
+def safeExtGet(prop, fallback) {
+    rootProject.ext.has(prop) ? rootProject.ext.get(prop) : fallback
+}
+
 android {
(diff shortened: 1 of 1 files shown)


ethanneff:
Can confirm this works for React Native 0.56.0. Can we get this merged?

yeomann:
yup I can confirm this is 💯  working, even I upgraded mine config to SDK 27 all good, test on API 23 on my phone as well.

msand:
Merged #729 and published v6.5.0

PR #740, opened 2018-07-25
On Android, the strokeMinterLimit cannot be cast into Float

NastuzziSamy:
This PR fixes this error (parseFloat or ParseInt, both work for this fix):

37782662_672226019795731_5905516406354477056_n


msand:
Published v6.5.0 @NastuzziSamy

PR #790, opened 2018-09-22
fix ios crash

devapro:
fix "this class is not key value coding-compliant for the key active"

### DIFF
--- a/ios/Elements/RNSVGSvgView.m
+++ b/ios/Elements/RNSVGSvgView.m
@@ -220,11 +220,15 @@
                 return node;
             }
 
-            UIView *hitChild = [node hitTest:transformed withEvent:event];
-
-            if (hitChild) {
-                node.active = YES;
-                return (node.responsible || (node != hitChild)) ? hitChild : self;
+            @try {
+                UIView *hitChild = [node hitTest:transformed withEvent:event];
+                if (hitChild) {
+                    node.active = YES;
+                    return (node.responsible || (node != hitChild)) ? hitChild : self;
+                }
+            }
+            @catch (NSException *exception) {
+                return nil;
             }
         }
         return nil;


msand:
Could you post a reproduction which triggers this exception? I'm not sure if this is the best way to solve this.

msand:
I'll close this because of inactivity. Not sure how  "this class is not key value coding-compliant for the key active" could possibly happen.

g33ktony:
Same here, I have an SVG map with touch events on some areas of the map, when I try to tap on the map it crash, with that error message

msand:
@g33ktony Hmm, do you have react-native Text elements inside the Svg element?

fjcaetano:
I'm also having the same exception being thrown, but it's on [RNSVGGroup.m#L178](https://github.com/react-native-community/react-native-svg/blob/master/ios/Elements/RNSVGGroup.m#L178)

It happens because if any of SVG's subviews is not a custom SVG view, like RCTView or RCTTextView for instance, it doesn't have the key `active`.

Wrapping the code in a try-catch stops the crashes, but it also means any hits are discarded.

fjcaetano:
One approach might be to replace that if with something like this:

```objc
    if (!event) {
        @try {
            NSPredicate *const anyActive = [NSPredicate predicateWithFormat:@"active == TRUE"];
            NSArray *const filtered = [self.subviews filteredArrayUsingPredicate:anyActive];
            if ([filtered count] != 0) {
                return [filtered.firstObject hitTest:transformed withEvent:event];
            }
        }
        @catch(NSException *e) {
            for (id subview in self.subviews) {
                @try {
                    [subview valueForKey:@"active"];
                }
                @catch(NSException *e2) {
                    return [subview hitTest:transformed withEvent:event];
                }
            }
        }
    }
```

msand:
Can you post a full replication? Btw, why do you have RNView/Text elements inside the Svg?

fjcaetano:
I don't have neither. I do have an Animated.View within it, but even if I remove it, there's still a RCTView being rendered inside it, I don't know why.

This is my SVG component, let me know if you can reproduce it:

```jsx
    const animations = {
      // Hides each chart while not currently active
      opacity: this.getOpacityAnimation(index),
      transform: [
        {
          rotate: this.getRotationAnimation(value, index),
        },
      ],
    };

    return (
      <Animated.View
        style={[styles.chart, animations]}
        key={`${value}_${color}`}
        removeClippedSubviews
      >
        <SVG width={TankChartSize} height={TankChartSize} viewBox='0 0 240 240'>
          <Defs>
            <Circle id='a' cx='120' cy='120' r='96' />
            <Mask id='b' x='0' y='0' width='240' height='240' fill='#fff'>
              <Use href='#a' />
            </Mask>
          </Defs>
          <G fill='none' fillRule='evenodd'>
            <Use
              stroke={color}
              mask='url(#b)'
              strokeWidth='80'
              strokeDasharray='301'
              transform='rotate(-90 120 120)'
              href='#a'
            />
            <Animated.View
              style={[
                styles.sunshine,
                {
                  backgroundColor: color,
                  opacity: this.sunshineAnimation(index),
                },
              ]}
            />
          </G>
        </SVG>
      </Animated.View>
    );
```

msand:
```jsx
            <Animated.View
              style={[
                styles.sunshine,
                {
                  backgroundColor: color,
                  opacity: this.sunshineAnimation(index),
                },
              ]}
            />
```
This code creates a native View, try using a Rect element from react-native-svg instead, set the fill instead of backgroundColor, give it a width and height of 240 or whatever you need, and set the opacity. Should work correctly then.

msand:
Alternatively, lift the Animated.View outside the Svg element, set position: absolute and place it where you need it.

fjcaetano:
That's not the issue. As I said before, even if I remove that Animated.View the crash still happens. I've just tested it again to be sure and I encountered the same behavior.

msand:
(With a few corrections) Seems to work fine: https://snack.expo.io/@msand/frowning-bananas

```jsx
import * as React from 'react';
import { Text, View, StyleSheet } from 'react-native';
import { Constants, Svg } from 'expo';

const { Defs, Circle, Mask, Use, G } = Svg;

export default class App extends React.Component {
  render() {
    return (
        <Svg width="100%" height="100%" viewBox="0 0 240 240">
          <Defs>
            <Circle id="a" cx="120" cy="120" r="96" />
            <Mask id="b" x="0" y="0" width="240" height="240">
              <Use href="#a" fill="#fff" />
            </Mask>
          </Defs>
          <G fill="none" fillRule="evenodd">
            <Use
              stroke="green"
              mask="url(#b)"
              strokeWidth="80"
              strokeDasharray="301"
              transform="rotate(-90 120 120)"
              href="#a"
            />
          </G>
        </Svg>
      </View>
    );
  }
}

  container: {
    flex: 1,
    justifyContent: 'center',
    paddingTop: Constants.statusBarHeight,
    backgroundColor: '#ecf0f1',
    padding: 8,
  },
});

```

msand:
But you can get the same output more efficiently with this:

```jsx
        <Svg width="100%" height="100%" viewBox="0 0 240 240">
          <Circle
            r="96"
            cx="120"
            cy="120"
            fill="none"
            stroke="green"
            strokeWidth="80"
            strokeDasharray="301"
            strokeDashoffset={0.5 * 96 * Math.PI}
          />
        </Svg>
```

PR #805, opened 2018-10-09
Remove prop-types

janicduplessis:
React Native is removing prop types from core components so this breaks some of the usages here. I decided to just get rid of prop types altogether instead of trying to add them back somehow. The community seems to be moving towards static typing instead of those runtime checks.

See https://github.com/facebook/react-native/issues/21342

Also removed the extra parameters to requireNativeComponents that used to do validations based on propTypes since it is no longer used (https://github.com/facebook/react-native/blob/master/Libraries/ReactNative/requireNativeComponent.js#L24)

Tested in an app running on react-native master and made sure it works again.


msand:
Thanks for this, should this be under a minor or major version bump? I guess some might have been relying on these being available. Perhaps better with a major bump?

msand:
Ideally we would have the types and interfaces generated from the IDL spec of svg and the related standards using https://github.com/motiz88/webidl-to-flow and 
https://github.com/Microsoft/TSJS-lib-generator
https://github.com/Microsoft/TSJS-lib-generator/blob/master/baselines/dom.generated.d.ts
https://raw.githubusercontent.com/Microsoft/TSJS-lib-generator/master/inputfiles/browser.webidl.preprocessed.json
or using:
https://www.w3.org/TR/SVG2/idl.html 
https://www.w3.org/TR/filter-effects/#idl-index
and any other needed idl files.
It could even be used to generate the react components, java & obj-c (and c++ for fiber) interface and view managers.

janicduplessis:
Are the prop types exported? If not I guess a minor would do, could also be safe and do a major.

janicduplessis:
I guess it'd be nice to add flow typings, I remember seeing a project that can generate flow from typescript bindings which we already have. I agree that ideally we'd have one source of truth for generating these props but that might be a pretty big projet to take on.

n-sviridenko:
We intent to remove prop-types for our production builds, but cannot do this because of this issue.

msand:
@n-sviridenko If you're in a hurry, then you can use his fork for now: https://github.com/janicduplessis/react-native-svg/tree/no-proptypes

n-sviridenko:
@msand thanks for a quick reply

janicduplessis:
Updated to also remove the extra parameters passed to requireNativeComponent since they are not used anymore.

msand:
Merged and released v8.0.0, thanks for the PR again!

PR #810, opened 2018-10-12
Fix logic for validating viewBox prop

### LINKED ISSUE #809: Invalid viewBox prop
Looks like a recent change flipped the logic on the validation of the viewBox prop in `lib/extract/extractViewBox.js`.

Formerly, a viewBox prop of "0 0 640 512" was valid. Now it's not (as of v7.1.1,and starting somewhere between 7.0.3 and 7.1.1)

Getting a console warning (and this might be causing some other things to break as well, but I'm still debugging other rendering differences after upgrading to 7.1.1)
```
  console.warn node_modules/react-native-svg/lib/extract/extractViewBox.js:35
    Invalid `viewBox` prop:0 0 640 512
```

Will post a PR shortly to propose a fix.

mlwilkerson:
Looks like the logic flipped recently on the viewBox prop validation when cleaning up some of the code.

This change seems to reflect the intention.

Fixes #809

### DIFF
--- a/lib/extract/extractViewBox.js
+++ b/lib/extract/extractViewBox.js
@@ -31,7 +31,7 @@ export default function(props) {
 
     const params = viewBox.trim().split(spacesRegExp);
 
-    if (params.length === 4 && params.every(param => !isNaN(+param))) {
+    if (params.length === 4 && params.some(param => isNaN(+param))) {
         console.warn("Invalid `viewBox` prop:" + viewBox);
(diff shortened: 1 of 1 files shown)


msand:
Oops. Thanks for this! Apparently I didn't read the rest of the line correctly when fixing the unused parameter warning.

msand:
Published v7.1.2 now, thanks a bunch!

PR #813, opened 2018-10-15
Fix require cycle between extractText.js and TSpan.js

pro-nasa:
Warning
```
Require cycle: node_modules/react-native-svg/lib/extract/extractText.js -> node_modules/react-native-svg/elements/TSpan.js -> node_modules/react-native-svg/lib/extract/extractText.js

Require cycles are allowed, but can result in uninitialized values. Consider refactoring to remove the need for a cycle.
```

It's not a pretty solution, but short :)

### DIFF
--- a/elements/TSpan.js
+++ b/elements/TSpan.js
@@ -9,7 +9,7 @@ import extractTransform from "../lib/extract/extractTransform";
 import Shape from "./Shape";
 
 // TSpan elements are shadow components
-export default class extends Shape {
+export default class TSpan extends Shape {
     static displayName = "TSpan";
 
     static propTypes = textProps;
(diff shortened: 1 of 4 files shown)


msand:
@pro-nasa Thanks for this! Looks like it should work. Could you move the _extractText helpers outside the component, they're not using `this`, so no need to attach them to `this` either no?

pro-nasa:
Moved :)

pro-nasa:
I replaced to more "clean" solution ;)

msand:
Excellent! Small cosmetic thing still, tspanElement could simply be TSpan instead. Would you mind restoring that naming?

pro-nasa:
Done. :)

msand:
Great, thanks!

PR #824, opened 2018-10-24
Do not create bitmap with 0 height/width during early render.

netmaxt3r:
when specify background color in react-native even parent view height/width could become zero
https://github.com/react-native-community/react-native-svg/issues/823


msand:
@netmaxt3r Thanks for this! Published v8.0.8

PR #835, opened 2018-11-02
Invalidate bitmap when view size is changed

igorlira:
This change fixes a bug on Android where the SVG would not be redrawn when the `SvgView` size changes but the React props did not. 

For instance, when a `<Svg style={{ width: '100%', height: '100%' }} />` is contained within a View that can dynamically change sizes.


msand:
Thanks for this!

PR #851, opened 2018-11-21
Update strokeDasharray docs

usrbowe:
### Why
I found that `strokeDasharray` with default value `[]` will crash on Android (iOS worked fine).
Tested on `RN v0.57.4` and `react-native-svg@8.0.8`.

Example:
```
<Path
      d={'M10 10 H 90 V 90 H 10 L 10 10'}
      fill={'transparent'}
      strokeWidth={2}
      stroke={'red'}
      strokeDasharray={[4, 3, 2, 1]}
    />
```
👉 I changed it to this and Android stop crashing.
```
<Path
      d={'M10 10 H 90 V 90 H 10 L 10 10'}
      fill={'transparent'}
      strokeWidth={2}
      stroke={'red'}
      strokeDasharray={'4 3 2 1'}
    />
```

Crash error:
`Error while updating property 'd' in shadow node of type: RNSVGPath`


msand:
Hmm, that seems to work just fine for me. I'll close this for now. Is this still an issue for you with the latest version?

maestor:
image

Had similar situation. After upgrading to version 9 (repeatable with 9.0.0 and 9.0.4), it crashes every time when dashed Line fired (and only Android). Screenshot about specific error attached. Happened first with RN 0.57.8 and now 0.58 versions. Didn't get reproduce that with latest 8 version.

I tried that solution to change strokeDasharray prop format from array to string, and looks like it worked, didn't get app crashing anymore. Not sure how that happens but maybe you can @msand find something with that information.

PR #860, opened 2018-11-29
Allow fontWeight to be a number

kristerkari:
Currently passing `fontWeight` property as a number (e.g. `fontWeigeht={20}`) causes a crash. This is a problem because SVGR always turns numeric `fontWeight` values to numbers.

Always casting fontWeight property to string allows the value to be a number without causing a crash.

More info in SVGR's repo:
https://github.com/smooth-code/svgr/issues/243

@msand

### DIFF
--- a/lib/extract/extractText.js
+++ b/lib/extract/extractText.js
@@ -51,7 +51,6 @@ export function extractFont(prop) {
         fontData,
         fontStyle,
         fontVariant,
-        fontWeight,
         fontStretch,
         textAnchor,
         textDecoration,
@@ -61,10 +60,11 @@ export function extractFont(prop) {
         fontVariantLigatures,
         fontFeatureSettings,
     } = props;
-    let { fontSize, fontFamily, font } = props;
+    let { fontSize, fontFamily, fontWeight, font } = props;
 
     fontFamily = extractSingleFontFamily(fontFamily);
     fontSize = fontSize ? "" + fontSize : null;
+    fontWeight = fontWeight ? "" + fontWeight : null;
 
     const ownedFont = _.pickBy(
         {


msand:
Hmm, I'd rather have this fixed on the native side, such that setNativeProps can give either a plain number or a string with units. @kristerkari Would you be willing to make changes on the native side instead?

kristerkari:
> Would you be willing to make changes on the native side instead?

I can if you can help by pointing out which files need the changes on Objective-C/Java side.

msand:
Would need to add support for (numeric) font-weights with values 1-999
https://github.com/react-native-community/react-native-svg/blob/2f513504d21a8138ac851784b3b110ccb8cceb3a/ios/Text/RNSVGFontData.m#L71-L72

And set it here: https://github.com/react-native-community/react-native-svg/blob/2f513504d21a8138ac851784b3b110ccb8cceb3a/ios/Text/RNSVGTSpan.m#L238-L260
Using
kCTFontVariationAttribute ios(3.2)
and perhaps
kCTFontFeatureSettingsAttribute ios(3.2)

And on Android:
https://github.com/react-native-community/react-native-svg/blob/2f513504d21a8138ac851784b3b110ccb8cceb3a/android/src/main/java/com/horcrux/svg/FontData.java#L94

And using setFontVariationSettings on Android O (version 26) and up,
to set it somewhere around here (before or after): https://github.com/react-native-community/react-native-svg/blob/2f513504d21a8138ac851784b3b110ccb8cceb3a/android/src/main/java/com/horcrux/svg/TSpanView.java#L256-L266

https://docs.microsoft.com/en-us/typography/opentype/spec/otvaroverview
https://developer.mozilla.org/en-US/docs/Web/CSS/CSS_Fonts/Variable_Fonts_Guide
https://medium.com/over-engineering/variable-fonts-in-android-p-c5c918275646
https://webkit.org/blog/7051/variable-fonts-on-the-web/
https://glyphsapp.com/tutorials/creating-a-variable-font

msand:
Alternatively, just adding some logic to round any number between 1-999 to the closest discrete font-weight, (and/or clamp any numbers outside the range), and mapping that to the corresponding enum would be fine for now as well. What do you think @kristerkari ?

msand:
Of course, we could also just merge this first, and make those enhancements in other PRs. Might be that animating the font-weight using the native driver would be quite niche/rare in practise. And, it would probably be quite intensive computationally for any significant number of even medium complexity glyphs

kristerkari:
I was working on other things on the weekend, so I did not have time to look at the changes yet.

> Of course, we could also just merge this first, and make those enhancements in other PRs. Might be that animating the font-weight using the native driver would be quite niche/rare in practise. And, it would probably be quite intensive computationally for any significant number of even medium complexity glyphs

If that's ok, we could merge and create a new issue with details for doing the changes on the native side.

> Alternatively, just adding some logic to round any number between 1-999 to the closest discrete font-weight, (and/or clamp any numbers outside the range), and mapping that to the corresponding enum would be fine for now as well.

So would that mean that you could pass in `fontWeight={420}`? I think that if the changes are done on the native side, then it's worth doing them properly by only supporting the correct font weight values.

kristerkari:
You are absolutely correct. When I wrote my comment I was only thinking about the values that React Native supports, but I completely forgot that the Web has a better support for font-weight.

kristerkari:
I started looking at making the changes based on your suggestions. I does not seem that difficult, but there are a couple of things that are a bit unclear to me still:

1) Did you have some idea also of how to organize the code when adding support for `font-weight` values `1-1000`? For example on Android everything is handled inside the `FontWeight` enum inside `TextProperties.java`. Should that function only handle `normal`, `bolder`, etc. after the changes, and where should the code for validating the value between `1-1000` live? 

2) How do you usually handle testing of the new features? Are there some fixtures where the new SVG features can be tested? Also, what would be a good test file to use to test variable font weights?

msand:
@kristerkari Hey, finally got back to looking at this, sorry for taking so much time. 

1. We can either add another data-structure, or another enum value to signify variable fonts weights together with a member in FontData called fontWeightNumeric of type double. Then in the property parsing, instead of assuming that it's a string, it would check if it's a number, and in that case check that it's in the correct interval, set the enum to variable and store the weight, or fallback to default in case of validation errors.
Similarly if it's a string and none of the enums match, it should check if the string can be parsed as a double in (1, 1000) and store the value. Alternatively, something like a FontWeightValue object/class could be defined, to contain the enum, the double, the parsing logic & getters. And then the fontFeatureSetting logic would check if the weight is variable, and in that case set WGHT to the given value.

I wish we had this written in rust at this point, it could represent this very nicely. Or, union types / algebraic data types in java/obj-c wouldn't hurt either.

2. Lately I've been setting up e2e testing over here: https://github.com/msand/react-native-svg-e2e
It works already, but is not in any way complete or optimized yet. Otherwise there's the example git submodule in this repo, but that hasn't been updated for a long time. And then I have a collection of random examples I've used in development and debugging, but there's no proper testing yet, or any collection of blessed reference screenshots to compare to. The web-testing group has some good material, which we could probably you with your transformer to be able to test significant amounts of logic and spec conformance. But that should probably run locally, it would likely cause timeouts in any free CI system.
For this case, I think a single text element with some text, and either a slider or textinput or buttons, to be able to set the font-weight to various values, or even just rendering two-three examples with the same text but different weight and no interactions/state would but perfectly fine as tests. A screenshot showing what some text looks like with default, minimum and maximum font-weight should cover most of the things which could break.

msand:
@kristerkari Are you still interested in working on this? Would be much appreciated ❤️

msand:
This might be good inspiration: https://github.com/servo/servo/blob/427003210b3e16a82f958e35ff4212f26ffa35ab/components/style/values/specified/font.rs#L71-L193

msand:
Also: https://github.com/servo/servo/blob/8de1b8d3f4d0e814980a1f7836a9e865e0d0f53b/components/style/values/computed/font.rs#L112-L138

kristerkari:
Thanks for the explanation and examples. I'll have another look at implementing the font changes, and I'll let you know if it is still unclear how to implement it.

msand:
Great to hear! Btw, I intend to merge the vector-effect branch today, it includes a commit to support native animation of font size: https://github.com/react-native-community/react-native-svg/commit/9b1ccf0e7f5b5d4e4e7dc22082e8fce63cca9be1 
It might be useful to get similar support for the font-weight
But, at least I'll need to fix the toDataUrl functionality before I merge and publish it, it's getting support for an optional options parameter, where one can specify the size of the canvas used for making the png.

msand:
As a side-note. Kotlin would have when expressions, and swift has a more general case statement, but not worth changing language to get it, they perform worse than java and obj-c in general in both compile and run time. For java there's also http://www.vavr.io/vavr-docs/#_the_basics_of_match_for_java but not worth bringing in a dependency, just to simplify a single if statement with regards to handling computations of a single data structure / enum / sum type. When/if a cross-platform rust rewrite would be done, it would make more sense, as it can be more performant and safe than java, kotlin, swift or obj-c anyway and handles it syntactically clearer than any of them.

msand:
Found a nice example of animating font-weight here: https://alligator.io/css/variable-fonts/

dl-husky73:
Can I ask if this was ever resolved?  I am having the same issue with an SVG with numbers that have font weight.  Crashes the Android app.

msand:
@dl-husky73 Can you try with the latest commit in the develop branch?

msand:
@kristerkari Could you test the latest commit from the develop branch?

kristerkari:
Sure, I'll do that tonight.

dl-husky73:
> @dl-husky73 Can you try with the latest commit in the develop branch?

Will do, I will try it out tomorrow.  I noticed another post about saving the SVG as optimized and got a few working that way.   I will try this so I don't need that step.   Thanks!

msand:
Now with an attempt for ios support as well: https://github.com/react-native-community/react-native-svg/commit/0a94b1ee891bf441445d1cb5d4af7036bbc89a22

kristerkari:
@msand I quickly tested the dev branch together with SVGR. The crash with `font-weight` is gone, and the variable font-weight also seems to be working, at least on iOS. Does it depend on the used font too? Because on Android with the default font there did not seem to be any difference between a font-weight of, let's say `700` and `900`, but on iOS they were rendered correctly.

msand:
@kristerkari can you try with bf0adb4 from https://github.com/react-native-community/react-native-svg/compare/develop
It has improved support for variable font weights, still better in ios than android but at least something.
It requires API level 28 / P to have all the relevant apis available, some others from 26 / O, earlier than that, it only has a heuristic for if its normal or bold font weight / style, with absoluteFontWeight >= 550 as the criteria for being bold (halfway from normal 400 to bold 700)

msand:
Screen Shot 2019-07-20 at 20 14 35

msand:
```jsx
import React, { Component } from 'react';
import { StyleSheet, View, Dimensions, Animated } from 'react-native';
import { Svg, Text, TSpan } from 'react-native-svg';

const { width, height } = Dimensions.get('window');
const AnimatedText = Animated.createAnimatedComponent(Text);

function getInitialState() {
  const anim = new Animated.Value(0);
  const oneTo1E3 = anim.interpolate({
    inputRange: [0, 1],
    outputRange: [1, 1000],
  });
  return { anim, oneTo1E3 };
}

  state = getInitialState();

  componentDidMount() {
    const { anim } = this.state;
    Animated.timing(anim, {
      toValue: 1,
      duration: 10000,
      useNativeDriver: true,
    }).start();
  }

  render() {
    const { oneTo1E3 } = this.state;
    return (
        <Svg width={width} height={height} viewBox="0 0 100 140">
          <Text fontSize={3}>
            <TSpan x={20} y={10}>San Francisco</TSpan>
            <TSpan x={20} y={30}>IBMPlexSansVar</TSpan>
            <TSpan x={20} y={50}>IBMPlexSansVar-Roman</TSpan>
            <TSpan x={20} y={70}>PublicSans-Thin_Regular</TSpan>
            <TSpan x={20} y={90}>Public-Sans-Roman-VF</TSpan>
            <TSpan x={20} y={110}>Default</TSpan>
            <TSpan x={20} y={130}>PublicSans-Thin_Black</TSpan>
          </Text>
          <AnimatedText fontWeight={oneTo1E3}>
            <TSpan x={20} y={20} fontFamily="San Francisco">Testing</TSpan>
            <TSpan x={20} y={40} fontFamily="IBMPlexSansVar">Testing</TSpan>
            <TSpan x={20} y={60} fontFamily="IBMPlexSansVar-Roman">Testing</TSpan>
            <TSpan x={20} y={80} fontFamily="PublicSans-Thin_Regular">Testing</TSpan>
            <TSpan x={20} y={100} fontFamily="Public-Sans-Roman-VF">Testing</TSpan>
            <TSpan x={20} y={120}>Testing</TSpan>
            <TSpan x={20} y={140} fontFamily="PublicSans-Thin_Black">Testing</TSpan>
          </AnimatedText>
        </Svg>
      </View>
    );
  }
}

  container: {
    backgroundColor: '#ecf0f1',
  },
});

```

msand:
cloneable example: https://github.com/msand/VariableFontWeightAnimation

msand:
Screen Shot 2019-07-20 at 22 35 30
Seems support for variable font weights isn't perfected yet. Some fonts work great in android, some in ios, but haven't found any font where animation from min to max would work perfectly in both. Anyways, at least it's possible to adjust it much more than before.

msand:
Seems subpixel positioning and linear text scaling made variable fonts really smooth on android as well. Please try with https://github.com/react-native-community/react-native-svg/commit/ea829103b34dfcbd8be5dcd53e5b1e518fff311b

msand:
@msageryd Can you test ea82910 ? I think it might resolve the issue you had with fonts on android in small viewBoxes as well.

msand:
Now with proper handling of min and max axis value for font weight variation on ios. Think this is quite usable at this point.

msand:
@wcandillion Perhaps you could get some ideas for some fancy animation of font weights somewhere? 😉 https://github.com/msand/VariableFontWeightAnimation

msand:
@kristerkari  Forgot to answer the second question. The variable font support (including font weights from 1-1000), requires that the font creator has added support for it, there can be several other design axes as well, allowing to create some really quite advanced animations of glyphs using interpolation of control points etc. You can find a bunch of variable fonts e.g. here: https://v-fonts.com/ and my test repo has three open source / public license fonts to test with. In addition, the default system font on iOS, San Francisco has some basic support for it.

msand:
On android, you can give additional inherited font variation settings, to drive e.g. secondary animation axes, by giving fontVariationSettings as a prop or a style on G, Text and Tspan elements. Haven't added native animation support for this yet, but should work with js driven ones at least.

msand:
I've merged the develop branch into master, and will be publishing a new version now, including support for numeric font weights and native animation of it, and support for variable font weights in opentype fonts which support them. So, I guess this can be closed now. Comment, create and issue or re-open if any problems arise.

kristerkari:
> I've merged the develop branch into master, and will be publishing a new version now, including support for numeric font weights and native animation of it, and support for variable font weights in opentype fonts which support them. So, I guess this can be closed now. Comment, create and issue or re-open if any problems arise.

Great! Thanks a lot for the effort

msageryd:
> @msageryd Can you test [ea82910](https://github.com/react-native-community/react-native-svg/commit/ea829103b34dfcbd8be5dcd53e5b1e518fff311b) ? I think it might resolve the issue you had with fonts on android in small viewBoxes as well.

@msand After a very long vacation, I'm now back in production. I have now tried RNSVG 9.6.2 with RN 0.60.4. I assume that your fix is in 9.6.2? 

I can confirm that the font scaling issues I had are now gone on iOS. 
Unfortunately the problem still exists on Android.

msand:
@msageryd Yes, 9.6.2 should have it. Can you make a small reproduction?

msageryd:
Yes, I suppose I could make a reproduction, but I'd like to elaborate some first so I'll know if it's realistic to solve this. After doing some more tests, I realise that the iOS version still has some quirks.

Here is roughly what I'm doing. This concept has been in production for a year and works fine, but it would be nice to get rid of the workaround, since it involves an unnecessary transform that I suspect might steal some FPS from me (this is only a guess).

All my shape coordinates are stored in 1x1 virtual coordinate system, i.e. (0.5;0.5) is in the middle of the coordinate system. Since the system is so small, all shape data is in a very small scale, i.e. linewidths, font sizes, etc. A typical font size is 0.001 in this virtual scale.

When I render the shapes, everything is scaled up. A typical "output size" 3000x3000 pixels. In this scale the 0.001 font size will become 3 (3000*0.001).

The scaling works fine for strokeWidth, dashes, etc. But fonts does not fare well in this small scale. My workaround is to scale up the `fontSize` and then scale down the rendered font with a transform. This way I won't bother the `fontSize` property` with my really small font size.

```
const VIRTUAL_TEXT_SCALE = 1000;

<View
  style={{left: 0, top: 0, right: screenWidth, bottom: screenHeight, position: 'absolute'}}
>
  <Svg width={screenWidth} height={screenHeight} >
    <G ref={ref => (this.svgGroupRef = ref)}>

      //Many shapes are rendered here, among those the text rendering looks like this:
      <SvgText fontSize={fontSize * VIRTUAL_TEXT_SCALE} >
        <TSpan
          transform={{scale: 1 / VIRTUAL_TEXT_SCALE}}
          x={x * VIRTUAL_TEXT_SCALE}
          y={y * VIRTUAL_TEXT_SCALE}
        >
          {this.props.labelText}
        </TSpan>
      </SvgText>

    </G>
  </Svg>
</View>


//New transform is set whenever needed (i.e. when the user pinch and pans)
this.svgGroupRef.setNativeProps({
  matrix: [matrix.a, matrix.b, matrix.c, matrix.d, matrix.e, matrix.f],
});
```

Here is a screen shot of some shapes as a reference. All text elements are treated with my workaround, i.e. "2.8m", "3" and "4". (the underlaying blueprint is just a jpg)

image

Issue #893, opened 2018-12-27
Memory leak (CGPathRetain?)


msageryd:
I'm just at the beginning of my investigation. Something in my app is clearly leaking. I tried to dig into the leaks tool in Instruments, but I'm not sure of what I'm seeing.

RNSVG versions tested **8.0.3** and **8.0.10**. Leaks are the same with booth versions.

For every SVG element it seems that there is a leak. Some of the leak-counts matches the number of SVG elements. The below screenshot is taken after jumping in and out of a drawing a couple of times. The drawing has 18 SVG groups. Each time I get 18 new leaks.

![image](https://user-images.githubusercontent.com/4483085/50479004-89bb9700-09d4-11e9-82c1-f29c6d38ea9e.png)

Here is one of the rows expanded
![image](https://user-images.githubusercontent.com/4483085/50478971-54af4480-09d4-11e9-9a66-e6d96941cfcd.png)

Could this be related to #863 ?

msageryd:
While testing some fixes by @msand (PR #897 ), I'm not able to reproduce the above leaks. I performed the same app interactions as above and got the following:

![image](https://user-images.githubusercontent.com/4483085/51106986-cd443c80-17ed-11e9-8e55-b0bcf4852148.png)

I think this confirms that the leaks are fixed.

Issue #646, opened 2018-04-10
How to translate <Use>?


msageryd:
I need to rotate my symbols, but I can't make out where the center of rotation is at.

Here is what I want to do:
```
<Use
  href={'#mySymbol'}
  transform={{ 
    rotation: 45,
    translateX: ?,
    translateY: ?
  }}
  x={x}
  y={y}
  width={w}
  height={h}
/>
```

I can't figure out how to translate in order to put the center of rotation in the center of the Symbol.  The Circles and the Symbols are rendered with the same center point as seen in the first two images. When I rotate the Symbol without translation I get very odd results.

Is this a bug? Or do I need to calculate the translation differently because of the scaling of the Symbol? In what order are the transforms performed? 
- scale Symbol via w*h
- rotate via transform
- translate via transform

**Rotation: 0, No translation**
![image](https://user-images.githubusercontent.com/4483085/38559502-793fbda8-3cd3-11e8-8430-32b273f47a6a.png)
![image](https://user-images.githubusercontent.com/4483085/38559512-86eaf8a0-3cd3-11e8-94e8-4a1a546b02d8.png)

**Rotation: 45, No translation**
![image](https://user-images.githubusercontent.com/4483085/38559555-9f738f7c-3cd3-11e8-848d-5a8ec9a32950.png)
![image](https://user-images.githubusercontent.com/4483085/38559734-17ea5dc8-3cd4-11e8-84a2-0ee6838e1844.png)

msand:
There seems to be a bug here, made a simple test to demonstrate, currently it seems to have the origin in the top left corner of the referenced symbol, when it should be using the CTM.
```jsx
import React, { Component } from 'react';
import { View, StyleSheet, Dimensions } from 'react-native';
import { Constants, Svg } from 'expo';
const { Symbol, Circle, Use, Rect } = Svg;

const { width, height } = Dimensions.get('window');

const hair = 100 * StyleSheet.hairlineWidth / width;

export default class App extends Component {
  render() {
    return (
      <View style={styles.container}>
        <Svg height={height} width={width} viewBox="0 0 100 100">
          <Symbol id="symbol" viewBox="0 0 100 100" width="10" height="10">
            <Rect x="0" y="0" width="100" height="100" fill="blue" />
            <Circle
              cx="50"
              cy="50"
              r="40"
              strokeWidth="8"
              stroke="red"
              fill="red"
            />
          </Symbol>

          <Rect
            x="0"
            y="0"
[...]

PR #897, opened 2019-01-05
Optimize startup

msand:


msand:
Ah ok. Well, I'm relatively confident the js code changes shouldn't break anything. But good get a bit more testing of it anyway.

msand:
@msageryd I tried porting the caching logic to iOS now as well, available in the latest commit. Needs regression testing, and validation that it actually improves performance.

msand:
Also, the memory retention needs to be analyzed.

msand:
I think this might be ready for a release candidate to npm now. Unless someone finds a bug/regression.

msageryd:
That sounds fantastic!
If you’re not in a hurry to release I’d be glad to help with some testing. I can do it on monday or maybe tuesday.



On 12 Jan 2019, at 22:47, Mikael Sand <notifications@github.com<mailto:notifications@github.com>> wrote:


I think this might be ready for a release candidate to npm now. Unless someone finds a bug/regression.

msageryd:
@msand Is the translation center fix related to #646 ?

msageryd:
@msand when you say that memory retention needs to be analyzed, do you mean that this PR could fix  #893 ?

msageryd:
#893 seems to be fixed by this PR.

msageryd:
I still cannot get `<Use>`to work with rotation transforms as per #646 . I'm not certain that this PR was aimed to fix this or if I have made some error. @msand are you able to rotate **Use** elements with correct pivot point?

msageryd:
Subjectively I see no performance boost. It's of course hard to see this by eye, but if the PR would , for example, have doubled the rendering performance I would have seen it. If there is a boost in the range of a couple of 10% I'm not sure I'd pick it up by eye.

msand:
@msageryd Alright, great. As long as there's no noticeable regression of performance it should be fine. The improvements are probably mostly noticeable if a Svg element has several G elements with lots of content, but only one/a few of the groups get changed. In this case, it should be able to reuse the Path objects representing the unchanged groups. But splitting things into separate Svg roots, if they have different update frequencies, should still be faster. As it can reuse the entire bitmap/raster output in those cases and just do the composition using the normal view rendering.

Can you try the latest commit? It should fix handling of Use in iOS.

msageryd:
I tried the latest commit as well (I saw the small iOS fix in this Android commit), but my positioning of `<Use>` element is still broke. It did change, though. My elements are now rendered in the lower right instead of the upper left corner (no transform applied to the **Use** element, but a parent Group (actually a grand-parent) has a scale transform.

msand:
@msageryd Do you have any example where it renders different from normal svg use on the web? At least https://github.com/react-native-community/react-native-svg/issues/646#issuecomment-381366800 seems to render the same in the browser, ios and android for me now.

msand:
screen shot 2019-01-14 at 16 21 47


```jsx
    render() {
        return (
            <Svg height="200" width="200" viewBox="0 0 100 100">
                <Symbol id="symbol" viewBox="0 0 100 100" width="10" height="10">
                    <Rect x="0" y="0" width="100" height="100" fill="blue" />
                    <Circle
                        cx="50"
                        cy="50"
                        r="40"
                        strokeWidth="8"
                        stroke="red"
                        fill="red"
                    />
                </Symbol>

                <Rect
                    x="0"
                    y="0"
                    width="100"
                    height="100"
                    stroke="blue"
                    fill="none"
                />
                <Use
                    href="#symbol"
                    x="0"
                    y="0"
                    width="20"
                    height="20"
                    transform="rotate(45, 10, 10)"
                />
                <Use
                    href="#symbol"
                    x="40"
                    y="40"
                    width="20"
                    height="20"
                    transform="rotate(-45, 10, 10)"
                />
                <Use
                    href="#symbol"
                    x="0"
                    y="80"
                    width="20"
                    height="20"
                    transform="rotate(-45, 10, 90)"
                />
            </Svg>
        );
    }
}
```

msageryd:
I did restart the packager. Should I have deleted the ios/build folder as well?

msand:
@msageryd Perhaps something like this might help:
watchman watch-del-all && rm -rf $TMPDIR/react-* && rm -rf $TMPDIR/metro* && rm -rf $TMPDIR/haste-*
react-native run-ios --reset-cache

msageryd:
Good catch Mikael. I did a "Clean build folder" and a build from within XCode. It now works as before. But the pivot point still seems to be at the 0,0 of the root Svg and not 0,0 of the parent Group. Which should it be?

Here is 45 degree rotation:

image

msageryd:
BTW. is the watchman and folder removals something that needs to be done everytime a native package is updated?

msageryd:
My bad!
I thought I had the circle in a group which was positioned further out in the coordinate system. It turned out that X and Y of the parent group is in the upper left corner of the drawing.

Everything works perfectly when I add originX and originY to the transform.

I'd say **ship it** :)

@msand Thank you soo much for your outstanding work with RNSVG!

msand:
These are actually for clearing the javascript cache, seems your native code had updated correctly, but used the old javascript for the react-native-svg module.
The origin of the Use element should be that of the containing element.
Thus e.g. in this content:

```html
    <svg height="200" width="200" viewBox="0 0 100 100">
      <symbol id="symbol" viewBox="0 0 100 100" width="10" height="10">
        <rect x="0" y="0" width="100" height="100" fill="blue" />
      </symbol>

      <g transform="translate(20, 20)">
        <rect x="0" y="0" width="100" height="100" stroke="blue" fill="none" strokeWidth="1" />
        <use href="#symbol" x="0" y="0" width="20" height="20" transform="rotate(45, 10, 10)" />
        <use href="#symbol" x="40" y="40" width="20" height="20" transform="rotate(-45, 10, 10)" />
        <use href="#symbol" x="0" y="80" width="20" height="20" transform="rotate(-45, 10, 90)" />
      </g>
    </svg>
```

you need to set the origin of the rotation using the optional x and y parameters of the rotate transform.
So, the non translated use element (all of which have width and height of 20 units), needs the origin at 10,10 to rotate about the center of it. And the use elements which have non-zero x or y attributes, need a different rotate definition to account for the translation.

msand:
That syntax is probably more efficient (but non-standard / not part of the svg spec), as it is already in object form, while the string needs to be parsed using the transform parser. Otherwise they should be equivalent with regards to input/output. 
The only remaining question now is if this should be a major, minor or patch version. There is certainly potential for breaking changes, so I guess it should be a major one, although the affected users might be relatively few. Just hope it doesn't scare people away from updating.

msand:
Oh, and btw, the bounds, center and frame calculation should fix the display of bounding boxes when using the inspector / standalone React Native Debugger, and it should improve pan responder and touch handling in several cases.

msageryd:
Apart from the rare possibility of a bug in this PR, the only breaking change would be if someone already has solved the `<Use>` problem with some kind of workaround. 

Such a workaround would have been quite complicated and the person doing it would probably have search the Issues section before tampering with the workaround.

To me this is not a major patch. Expo is relying on RNSVG quite a bit, I think. Maybe @ide have any thought on major vs minor patch.

msand:
True. Actually I already released v9.0.0
Although the potential for breaking change is relatively minor, it still changes the semantics/effects of the Use element x, y, and transform attributes. It's mostly in consideration of any user(s) who might have thought it was behaving correctly and found values which work for them using trial and error rather than using the standard / specified behavior of svg.

msand:
Hmm, haven't seen that with font sizes. Could you make a repro?

msageryd:
I really don't have time right now. I'll get back to you with a repro.
Here is a simple example of my workaround:

```
const VIRTUAL_TEXT_SCALE = 1000;

return (
  <Text
    transform={{scale: 1 / VIRTUAL_TEXT_SCALE}}
    fontSize={fontSize * VIRTUAL_TEXT_SCALE}
  >
    TEST
  </Text>
);
```

PR #920, opened 2019-01-31
Handle href for Linear Gradient

gosuperninja:
For Issue https://github.com/react-native-community/react-native-svg/issues/919

Cache the children incase an href is used.
This wont handle id's which are loaded after a href


msand:
The resolution probably needs to be done on the native side, and per svg root. Such that id's only get resolved from within a single svg document rather than globally, and such that the gradients can be defined in any order, as I suspect some people/tools might author such content.

gustavopch:
So — 2 years later — any news?

Issue #921, opened 2019-02-02
Logo renders for a split second at top left of screen before parent loads.


Orbyt:
### Description
I'm in the process of swapping some bitmap logos to SVGs for various performance and usability reasons. I am successfully using the the SVG component I built using `react-native-svg` on another screen, however on our "splash" screen a strange issue has become apparent.

### Problem
The logo appears to render at an incorrect position of `0, 0` before it's own parent `<Container/>` renders. Once the `<Container/>` renders, the logo is displayed in the correct position. However, this duplicate render and the change of position causes a flickering effect. This can be seen as the parent `<Container/>` has a bright red background, and the logo can be seen for a short amount of time before the `<Container/>` renders. The issue is reproducible using various SVGs built with the components provided by `react-native-svg`.

Here is the `render` function of the splash screen:

```
<Container style={{ backgroundColor: "red",  alignItems: 'center', flex: 1, justifyContent: 'flex-end' }}>
    <BasketballOutlineIcon size={100} }/>
</Container>
```

The issue is also reproducible when using a simple `<View/>` parent without flexbox properties.

### Video
Below is a video demonstrating the issue. The blue background is the window background, the red background is the `<Container/>` that contains the `<MyLogo/>` component. Notice how the logo renders around position `0, 0` before the red appears.

[Video](https://gfycat.com/thriftytatteredamethystinepython)
 
### Versions

- `"react-native-svg": "9.1.1"`
- `"react": "=16.6.3"`
- `"react-native": "=0.57.8"`
- `"native-base": "^2.10.0"`

msand:
@Orbyt Can you try adding a `return null;` statement in between these two lines:
https://github.com/react-native-community/react-native-svg/blob/b3d29c904508aa8141e678a03049efbee544915c/android/src/main/java/com/horcrux/svg/SvgView.java#L213-L214

Orbyt:
@msand I went into `node_modules/react-native-svg/android/src/main/java/com/horcrux/svg/SvgView.java` and made the following changes:

```java
if (early) {
   ViewParent viewParent = getParent();
   View parent = null;
    if ((viewParent instanceof View)) {
    parent = (View)viewParent;
    }
    float parentWidth = parent == null ? 0 : parent.getWidth();
    float parentHeight = parent == null ? 0 : parent.getHeight();
    width = (float) PropHelper.fromRelative(mbbWidth, parentWidth, 0, mScale, 12);
[...]

PR #924, opened 2019-02-04
Prevents early rendering.

Orbyt:
Related to #921.

### DIFF
--- a/android/src/main/java/com/horcrux/svg/SvgView.java
+++ b/android/src/main/java/com/horcrux/svg/SvgView.java
@@ -222,19 +222,7 @@ public class SvgView extends ReactViewGroup implements ReactCompoundView, ReactC
         float width = getWidth();
         float height = getHeight();
         boolean early = Float.isNaN(width) || Float.isNaN(height) || width * height == 0 || (Math.log10(width) + Math.log10(height) > 42);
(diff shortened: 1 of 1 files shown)


Orbyt:
@msand Does anything further need to be done before this can be merged?

msand:
Is the separate width and height parts needed? If so, you can probably replace the width * height == 0 in the early predicate

Orbyt:
@msand Which part are you referring to? 

There was previously a simple if block:
```
if (width == 0 || height == 0) {
    return null;
}
```
this was combined with the if statement above it as they both return `null`.

msand:
Yeah, i just noticed that the code that was changing those values got removed/doesn't exist anymore, and only the product, or comparing both width and height to zero separately, is needed. Doing both is redundant now that they're only set once.

Orbyt:
@msand Are you suggesting the removal of `width == 0 || height == 0` or `width * height == 0 ` found in the declaration of the `early` boolean?

msand:
@Orbyt there's another pr https://github.com/react-native-community/react-native-svg/pull/925 to check for negative values as well, so I would remove the `width * height == 0`, and make it check for `width < 1 || height < 1` or <= 0, instead.

PR #944, opened 2019-02-18
[ts] Fix SvgProps and ResponderProps interfaces.

flsilva:
As per discussion on https://github.com/react-native-community/react-native-svg/issues/943, `<Svg>` component accepts the same props as `<G>`, so `SvgProps` should extend `GProps`.

This PR does just that, but on doing so we get the following error from TypeScript compiler:

```
Interface 'SvgProps' cannot simultaneously extend types 'GProps' and 'ViewProps'.
Named property 'pointerEvents' of types 'GProps' and 'ViewProps' are not identical.ts(2320)
```

`GProps` extends `CommonPathProps`, which extends `ResponderProps`. No problem here. The problem seems to be that `pointerEvents` on `ResponderProps` seems to have a wrong type:

```
export interface ResponderProps extends ReactNative.GestureResponderHandlers {
  pointerEvents?: (event: any) => any,
}
```

`pointerEvents` is a prop of `ReactNative.ViewProps` as well, but with a different type:

`pointerEvents?: "box-none" | "none" | "box-only" | "auto"`

Searching for `pointerEvents` on `react-native-svg` codebase I could only find the following matches:

```
lib/extract/extractResponder.js

const pointerEvents = props.pointerEvents;
  if (pointerEvents) {
    extractedProps.pointerEvents = pointerEvents;
  }
```

Since those props are passed down to `<NativeSvgView>` and not handled anywhere else, I'm assuming that it ends up being handled by React Native itself somewhere, which would mean that the type of `pointerEvents` is indeed wrong, and it should be the same as the one in `ReactNative.ViewProps`. So alongside changing `SvgProps` I also changed that in this PR. Let me know if I'm missing something and that's not the case, in which case we might have another issue, i.e., a naming conflict between `ResponderProps`'s `pointerEvents` and `ReactNative.ViewProps`'s `pointerEvents`.

### DIFF
--- a/index.d.ts
+++ b/index.d.ts
@@ -80,7 +80,7 @@ export interface TouchableProps {
 }
 
 export interface ResponderProps extends ReactNative.GestureResponderHandlers {
-  pointerEvents?: (event: any) => any,
+  pointerEvents?: "box-none" | "none" | "box-only" | "auto",
 }
 
 // rgba values inside range 0 to 1 inclusive
--- a/index.d.ts
+++ b/index.d.ts
@@ -306,7 +306,7 @@ export interface StopProps {
 }
 export const Stop: React.ComponentClass<StopProps>;
 
-export interface SvgProps extends ReactNative.ViewProperties {
+export interface SvgProps extends GProps, ReactNative.ViewProperties {
   width?: NumberProp,
   height?: NumberProp,
   viewBox?: string,
--- a/index.d.ts
+++ b/index.d.ts
@@ -311,6 +311,8 @@ export interface SvgProps extends GProps, ReactNative.ViewProperties {
   height?: NumberProp,
   viewBox?: string,
   preserveAspectRatio?: string,
+  color?: int32ARGBColor | rgbaArray | string,
+  title?: string,
 }
 
 // Svg is both regular and default exported


flsilva:
Just noticed I forgot to add `color` and `title` to `SvgProps`, gonna push another commit.

msand:
I just noticed we're not really spec conformant. The svg spec requires support for a few more pointer-events values: https://www.w3.org/TR/SVG11/interact.html#PointerEventsProperty

Name: | pointer-events
-- | --
Value: | bounding-box \| visiblePainted \| visibleFill \| visibleStroke \| visible \| painted \| fill \| stroke \| all \| none
Initial: | visiblePainted
Applies to: | container elements, graphics elements and ‘use’
Inherited: | yes
Percentages: | N/A
Media: | visual
Computed value: | as specified
Animatable: | yes



https://www.w3.org/TR/SVG2/interact.html#PointerEventsProp
But, for now we can just keep it in sync with react-native.

flsilva:
Hmm they seem different, you think they can be represented by a same interface? What would the mapping between React Native's `"box-none" | "none" | "box-only" | "auto"` to those SVG spec values look like?

msand:
https://facebook.github.io/react-native/docs/view#pointerevents

> `pointerEvents`
> Controls whether the View can be the target of touch events.
> 
> `'auto'`: The View can be the target of touch events.
> `'none'`: The View is never the target of touch events.
> `'box-none'`: The View is never the target of touch events but it's subviews can be. It behaves like if the view had the following classes in CSS:
> ```css
> .box-none {
>      pointer-events: none;
> }
> .box-none * {
>      pointer-events: all;
> }
> ```
> 'box-only': The view can be the target of touch events but it's subviews cannot be. It behaves like if the view had the following classes in CSS:
> ```css
> .box-only {
>      pointer-events: all;
> }
> .box-only * {
>      pointer-events: none;
> }
> ```

Html only supports auto, inherit and none:
https://developer.mozilla.org/en-US/docs/Web/CSS/pointer-events#Values
https://developer.mozilla.org/en-US/docs/Web/SVG/Attribute/pointer-events
https://tympanus.net/codrops/css_reference/pointer-events/

msand:
https://developer.mozilla.org/en-US/docs/Web/CSS/pointer-events#Values
`auto`
The element behaves as it would if the pointer-events property were not specified. In SVG content, this value and the value visiblePainted have the same effect.
`none`
The element is never the target of pointer events; however, pointer events may target its descendant elements if those descendants have pointer-events set to some other value. In these circumstances, pointer events will trigger event listeners on this parent element as appropriate on their way to/from the descendant during the event capture/bubble phases.

https://webkit.org/specs/PointerEventsProperty.html
The pointer-events property was originally defined for SVG content and, as such, accepts property values specific to vector graphics (eg. relating to graphical fill and stroke). The following definition only applies to the property when used with HTML content. Note that the property has a new initial value, "auto", which behaves as "visiblePainted" in SVG content and "visible" in non-SVG content.

`auto`
In SVG content, behave as visiblePainted. Otherwise, behave as visible.
`visible`
The given element receives pointer events.
`none`
The given element does not receive pointer events.

flsilva:
Thanks for the info, I learned something. So it seems they are compatible, but even so, I don't think it's possible to extend `ResponderProps`'s `pointerEvents` to add extra SVG-only values like:

```
export interface ResponderProps extends ReactNative.GestureResponderHandlers {
  pointerEvents?: "box-none" | "none" | "box-only" | "auto" | "stroke" | "visibleStroke"
,
}
```

TypeScript will complain with the same error, i.e. `pointerEvents` attributes are not identical. At a first thought I can see only three solutions:

- Change React Native `ViewProps`'s `pointerEvents` to consider SVG (probably tricky since RN itself doesn't support SVG).
- Change the name of `ResponderProps.pointerEvents` to something else. Probably not very nice.
- Change `SvgProps` so it doesn't extend `ReactNative.ViewProps`, removing the conflict altogether and letting `ResponderProps.pointerEvents` free to have all SVG-related options, which prompts me to the following question: does `<Svg>` supports all ReactNative.ViewProps, or just some of it? If that's just some of it, removing this inheritance could be the solution, adding to `SvgProps` only what it implements.

Anyway, that seems a deeper problem that could be discussed in another issue I guess, unless you think we can find a quick solution to include in this PR. I think the changes in this PR represent an improvement for TypeScript support, fixing the compile errors I was getting when trying to use those props (`color`, `fill`, etc).

flsilva:
Let me know if there’s anything else I can do to move this PR forward. I’ve been using it locally on a project I'm working on and everything seems to be working as expected.

Thanks.

msand:
Sorry for not getting back to this earlier, had other priorities I needed to deal with. Thanks for the PR!

PR #947, opened 2019-02-19
[docs] Add pencil example to showcase color inheritance.

### LINKED ISSUE #943: Is <Svg fill="..."> officially supported?
Hello,

First of all, thank you for your hard work on such an amazing library.

I'm looking for an easy way to color all SVG elements at once. My use case is one of the most common for SVGs: icons. When we work with third-party icons we don't have control over their internal structures, and to be honest, we don't care, do we? All we want to do is to render them, and (easily) color them, all elements with a same color. So most of the times we don't want to have to use fill on each internal SVG element (`<Path>`, `<Circle>`, etc). That'd also prevent us to create a simple and reusable way to change the color of the elements, since each SVG can have its own internal structure. Multicolor icons are a different story though, but they are rare and demands a custom manipulation anyway.

I tried doing `<Svg fill="#FF0000">` on some third-party SVG icons and it does work, i.e., everything is filled with such color. I've tested with icons that have only a single `<Path>`, but also multiple `<Circle>`, and `<Path>` and `<Circle>`. All of them seems to work correctly, i.e., all elements are colored, except the ones that have themselves a fill prop, which is expected to take precedence over a `fill` on the root `<Svg>` anyway.

So my question is: is it officially supported by react-native-svg? Or is there any other way to do that?

AFAIK the current docs only show the use of the `fill` prop on SVG elements (`<Path>`, `<Circle>`, etc), but not on `<SVG>` component itself. So even though it does work, I'd like to make sure that it's officially supported. It seems to me that it should be (or something similar), for the reasons I mentioned above, but not sure about further implications.

If that's officially supported, it'd be great to make it explicit in the docs by having at least one example of it. The other required change I can see would be to add `fill?: string` to `SvgProps` TypeScript type. I'd be glad to send a PR with such changes if that's officially supported. Otherwise, could we get clarifications on why that's not officially supported, even though it works as expected? Or if there's a different way to achieve the same thing?

Thank you for your time.

flsilva:
Closes #943.


flsilva:
Ah sure, makes sense, changing now.

msand:
Thanks for this!

Issue #945, opened 2019-02-18
Add support for react-native-dom


dutzi:
[react-native-dom](https://github.com/vincentriemer/react-native-dom/) is a complete implementation of React Native for the web.

react-native-video already [implemented support for it](https://github.com/react-native-community/react-native-video#installation), it would be great if we could have it here also. 

I'm not sure if it's currently possible, but I'm [figuring it out](https://github.com/vincentriemer/react-native-dom/issues/98#issuecomment-464891595).

Issue #265, opened 2017-02-27
Reuse on Web?


timscott:
I have created a graph using react-native-svg and D3. Is there a simple way to to reuse it for web, in an HTML page?

If not, how hard would it be to add this capability?

RangerMauve:
The best way would be to go to the `elements` folder and create web versions of each of those components by using `React-DOM`.

For example the [G](https://github.com/react-native-community/react-native-svg/blob/master/elements/G.js) component could look like:

```javascript
export default class extends Shape{
    static displayName = 'G';

    static propTypes = pathProps;

    render() {
        let {props} = this;

        return <g
            {...extractProps(props, this)}
        >
            {props.children}
        </g>;
    }
}
```

In this example, I've replaced the `RNSVGGroup` component with the HTML tag `g` which is a tag that's supported out of the box.

ndbroadbent:
[react-native-web](https://github.com/necolas/react-native-web) support would be awesome. Like @RangerMauve said, it would just require `*.web.js` versions of each component. This shouldn't be too hard, since this is just a subset of SVG, and it's just rendering XML tags.

I might start working on some of this, but I don't know if I'll do everything. Right now I'm just using SVG to add an outline to some text, so I'll get that working first. I'll post a link to my branch when I'm finished, if anyone wants to take over and finish everything.

timscott:
@ndbroadbent - Let me know how it goes. I might be able to jump in at some point. 

Regarding PR, are you saying it would need to take a dependency on ```react-native-web```? Maybe it could be done without such a dependency?

RangerMauve:
@ndbroadbent Worst case, you can have this be a separate project (call it `react-native-svg`, maybe?), and then people can alias `"react-native-svg": "react-native-svg-web"` in their projects

ndbroadbent:
@timscott Hi, no there would be no dependency on `react-native-web`, the only thing we need to do is create the `*.web.js` versions and then it will just work. But I was just saying that the `web` platform is not really officially supported by React Native, so the maintainers of this library might only want to support Android and iOS. But it would be cool if they would agree to merge in web support.

sibelius:
you should check what https://github.com/FormidableLabs/victory-native does

msand:
You can also use https://github.com/godaddy/svgs instead, svgs is a compatiblity layer between svg and react-native-svg

timscott:
@msand, Sweet. That looks perfect. I'll try it out. Thanks!

[...]

PR #950, opened 2019-02-20
[Web] Added SVG web implementation from Expo

EvanBacon:
I've added the basic implementation of SVG for `react-native-web` that I've been using internally. I noticed that there were some other PRs and issues regarding this so I wanted to get feedback before doing anything drastic. 

# Related:

#593 
#945
#265

CC: @necolas @ide

### DIFF
--- a/index.web.js
+++ b/index.web.js
@@ -1,0 +1,59 @@
+import { createElement } from 'react';
+
+function generate(componentName, tagName): any {
+  function SvgElement(props) {
+    return createElement(tagName, props, props.children);
+  }
+  SvgElement.displayName = componentName;
+  return SvgElement;
+}
+
+const types = {
+  Circle: 'circle',
+  ClipPath: 'clipPath',
+  Defs: 'defs',
+  Ellipse: 'ellipse',
+  G: 'g',
+  Image: 'image',
+  Line: 'line',
+  LinearGradient: 'linearGradient',
+  Path: 'path',
+  Polygon: 'polygon',
+  Polyline: 'polyline',
+  RadialGradient: 'radialGradient',
+  Rect: 'rect',
+  Stop: 'stop',
+  Symbol: 'symbol',
+  Text: 'text',
+  TextPath: 'textPath',
+  TSpan: 'tspan',
+  Use: 'use',
+};
+
+export const Svg = generate('Svg', 'svg');
+
+for (const [componentName, tagName] of Object.entries(types)) {
+  Svg[componentName] = generate(componentName, tagName);
+}
+
+export const Circle = Svg.Circle;
+export const ClipPath = Svg.ClipPath;
+export const Defs = Svg.Defs;
+export const Ellipse = Svg.Ellipse;
+export const G = Svg.G;
+export const Image = Svg.Image;
+export const Line = Svg.Line;
+export const LinearGradient = Svg.LinearGradient;
+export const Path = Svg.Path;
+export const Polygon = Svg.Polygon;
+export const Polyline = Svg.Polyline;
+export const RadialGradient = Svg.RadialGradient;
+export const Rect = Svg.Rect;
+export const Stop = Svg.Stop;
+export const Symbol = Svg.Symbol;
+export const Text = Svg.Text;
+export const TextPath = Svg.TextPath;
+export const TSpan = Svg.TSpan;
+export const Use = Svg.Use;
+
+export default Svg;
--- a/index.web.js
+++ b/index.web.js
@@ -1,4 +1,4 @@
-import { createElement } from 'react';
+import { createElement } from 'react-native-web';
 
 function generate(componentName, tagName): any {
   function SvgElement(props) {
--- a/package.json
+++ b/package.json
@@ -7,7 +7,7 @@
         "url": "https://github.com/react-native-community/react-native-svg"
     },
     "license": "MIT",
-    "main": "./index.js",
+    "main": "./index",
     "keywords": [
         "react-component",
         "react-native",


msand:
Does this handle giving arrays of styles which have been processed by Stylesheet from react-native-web?
It optimizes them to be represented as integers, and I think createElement imported from react won't handle them correctly: https://github.com/react-native-community/react-native-svg/pull/593#issuecomment-394803437

Will probably need createElement from react-native-web and perhaps Stylesheet.compose:
https://github.com/necolas/icon-builder-example/blob/master/scripts/createReactPackage.js#L31-L36

@necolas Please advice

necolas:
Yeah @msand is right. I mentioned this somewhere in one of the existing issues (as well as the blog post I wrote about using SVGs with RNW at Twitter)

msand:
@necolas Is it sufficient to use createElement from react-native-web, to handle the integer representation? Or does the style prop need processing before that? It seems the Stylesheet.compose you have in your article might only be needed if you have a fallback stylesheet. But, if you don't have/need that, you can just call createElement with the props straight away, no?

necolas:
Only createElement is needed. Stylesheet.compose is identical to RN and is useful to avoid array allocations when one of two style values may be falsy

msand:
Great, thanks for the information! I guess it might be good to wrap the import/require of react-native-web in a try catch block, and import from react in case react-native-web isn't available. Then it can be used in plain react projects as well.

necolas:
Except that would result in bizarrely inconsistent APIs and might result in errors

msand:
Well, if it's used in a library, then it should define a peer dependency. And when used inside a single project, the authors should probably be aware of if they're using plain react or if they've configured react-native-web to work and if they're using Stylesheet or not. In either case, it should work as expected no? Or am I forgetting something critical?

necolas:
I think React DOM projects can just use the built-in React DOM SVG elements, and cross-platform projects can use this cross-platform API

msand:
Well, e.g. all the download of svgs builds essentially on the want to have support for that use case, but with support for the additional attributes / behavior which is specific to react-native-svg rather than the svg spec. I guess people can just keep using svgs for that/those use cases. But seems like it would make sense to support it in this library straight away as well, rather than forcing people who want to develop svg components for both native and web, to use react-native-web in the plain web projects. Or are there any drawbacks to supporting that? I don't think I've seen any argument against it yet, or what lies behind your reasoning? I'm not sure I completely understand it or what would be bizarre about the api.

ide:
Wrapping require calls in try-catch is prone to errors for client-side software because the dep graph is built statically. It also doesn't work with the JS standard (ESM). For that reason alone I'd lean towards unconditional requires/imports.

msand:
Shouldn't require calls obey the [CommonJS spec](http://www.commonjs.org/specs/modules/1.0/) ?

> Contract
> Module Context
> 1. In a module, there is a free variable "require", that is a function.
>     1. The "require" function accepts a module identifier.
>     2. "require" returns the exported API of the foreign module.
>     3. If there is a dependency cycle, the foreign module may not have finished executing at the time it is required by one of its transitive dependencies; in this case, the object returned by "require" must contain at least the exports that the foreign module has prepared before the call to require that led to the current module's execution.
>     4. If the requested module cannot be returned, "require" must throw an error.
> 
> 2. In a module, there is a free variable called "exports", that is an object that the module may add its API to as it executes.
> 3. modules must use the "exports" object as the only means of exporting.

I haven't seen it fail in practise when used in dependencies. Can you give an example?
If this is an actual issue. Then I would suggest keeping this kind of interop layer outside the react-native-svg repo, as there are multiple options, and either choice made would upset the users preferring the other option. If its kept in external modules, such as `svgs` if you want plain react, and `swgs` if you want react-native-web, then you get choose which import you prefer.

Then you get to choose if you want your svg components to work with all react projects, or only with projects which provide react-native compatible Stylesheet (for both native and web project) and the react-native-web optimized createElement (for web projects, not used in native apis at the moment).

Alternatively, a third module could be made (maybe react-svg-compat or something like this), which requires react-native-web and if that fails, requires react. This would allow using the optimization where possible, and still work in other environments, and find out if it causes problems in practise, (given enough exposure).

But, if the support for using react-native-web is unanimous, then we can probably merge it here. But, then the api on the web should probably match the api on native. Thus it should have the logic developed in the svgs package, rather than a stripped down implementation like this. Otherwise the api would actually have quite significant differences. And, if a stripped down web implementation is wanted, then this can be made into another external package which merely does what this PR includes, and uses the normal react-native-svg implementation otherwise.

What do you think?

msand:
@3rd-Eden Any thoughts on this?

ide:
API: if react-native-svg were to gain a web implementation, its interface probably should be close to / the same as the iOS and Android implementations (and same for future platforms as well e.g. macOS). This includes being able to accept StyleSheet.create() objects.

requires: The CommonJS spec snippet doesn't talk about how modules are made available -- Node behaves differently than Webpack despite both implementing the spec because the spec leaves some things up to the implementation, namely fetching the modules. In Webpack's case, part of the module fetching happens statically during the build and is at odds with dynamic features like `try-catch`. ESM imports need to be top-level unconditional statements as well for similar reasons.

Fortunately, we can still do build conditional dep graphs statically as long as the conditions are statically known. Probably the simplest way to do this today is with conditional extensions like `*.web.js` and Webpack/Metro/whatever can choose whether or not to include those files in the dep graph. Rewriting imports with Babel is another option but has more moving parts.

EvanBacon:
Echoing @ide, if you wrap a require with a try/catch then treeshaking won't work properly. This will cause major performance problems.

msand:
Afaik, tree shaking doesn't work at all with require, nor does it work well with modules which have side-effects in their initialization. In this case, both svgs and react-native-svg have very little js code to tree-shake to begin with.

And, I think you might be thinking of project code which gets transpiled. If you have a dependency which you import (or require for that matter), if it uses require internally, it'll use CommonJS compatible behavior and won't be able to use import/export syntax, but will need to use the module.exports syntax instead. So, if you have a dependency in node_modules which uses require (or has been transpiled to use it), then it'll resolve the modules correctly, and won't include any additional code. At least what I've seen in the wild. But if anyone has an example of actual code demonstrating this issue then I'm very interested to see it and know to avoid it.

msand:
Regarding the Stylesheet api, currently .create in the react-native implementation is the identity function (in production, dev mode has validation as well): https://github.com/facebook/react-native/blob/2f32fea79a50bdae303126f6d5288ea447f5331b/Libraries/StyleSheet/StyleSheet.js#L346-L362

So, supporting that in the web version is relatively trivial.
The main thing to support is the additional attributes that react-native-svg has. And, when using a react-native-web project, to resolve the integer representations to the actual classnames/styles. This can already be done by using the swgs package, it is essentially exactly the same as the svgs package, except it imports react-native-web instead of react for the web version.

If anyone needs to override the web implementation, then they could import e.g. svgs to use plain react. Or, some other implementation like the one in this pr, to leave out the extra attributes (if it would be published as a separate package).

So, I would just like know what @3rd-Eden thinks of this, and that he would be fine with merging our efforts into this repository (with react-native-web imported, so essentially deprecating swgs and moving it in here), and keeping svgs going as is for the plain react-web projects.

@EvanBacon @ide @necolas Would this satisfy all your wishes?

3rd-Eden:
Apologies for the late response, life happened. I've spend some time looking through the PR and related discussions. 
	
	First things first, the PR looks solid and I don't see any issues with the way it's currently drafted. It should not provide any incompatibility issues with `svgs` for those who are currently using it together with WebPack and are already importing `.web.js` dependencies, as the `svgs` project only imports the `react-native-svg` project in the dedicated `.native.js` file. But that is definitely something that we need to be aware of as any changes regarding this setup could potentially break.
	
	From a more philosophical standpoint I would rather see us all working together and joining forces and put all our efforts in a single blessed solution for the whole community instead of creating competing packages. It's a waste from everyone's development time as well as confusing for users..

msand:
I'm very much in support of merging our efforts 👍 I'm thinking that two working example projects: one plain react and another using react-native-web, importing from a single library, without fetching/bundling unused code, would be the ideal situation. I wonder if we could use different file extensions to handle it nicely, or provide differing implementations under different subfolders and allow to choose which one is wanted using the webpack/babel/metro config. I haven't seen any library with differing implementations for plain react web and react-native-web, but perhaps .rweb.js and .rnweb.js, and a default .web.js could work?

Or, actually, perhaps the plain react web implementation could be in react-native-svg/web/index.js Such that if you're interested in the use case of not using Stylesheet and want compatibility with both plain react web projects and react-native-svg, then you can import from there, knowing that you won't force your library users to adopt react-native-web and Stylesheet in their web projects. In case they meet too much resistance to it in their organization/project, or just want a wider potential user base for svg based react components. 

And then, we would provide the react-native-web compatible implementation as index.web.js in the root as the default. As you're relatively likely to be using react-native components and Stylesheet for something in your project anyway, in case you're importing react-native-svg.

ide:
It seems much simpler to focus on supporting react-native-{ios,android,web}. There's already an excellent solution for React DOM -- it just works with JSX intrinsic elements -- and the audience here is people who are using SVGs with react-native-{android,ios}. The point & scope of this PR, as I understand it, are to give react-native-svg users a uniform API across platforms in the same way that react-native-web and unimodules already do.

msand:
@ide The discussion with @3rd-Eden in this case is about https://www.npmjs.com/package/svgs which is the library that implements the react-native-svg compatible layer for plain react web projects (and doesn't use/require react-native-web). It's useful for anyone wanting to make platform independent svg components without necessarily depending on react-native-web. 

I've forked svgs as swgs, and only changed it to import react-native-web instead of react, to get the api compatibility you're asking for (or am I missing something?). So in this case I guess it just means including the https://github.com/msand/swgs/blob/master/index.js file as index.web.js in this project instead?

And if we want to merge the svgs use case into this repo as well, then we can either provide it from a subfolder, such that no conflicts should arise for the react-native-web use case and module fetching/bundling, or alternatively a monorepo could be used. But the monorepo structure would essentially entail maintaining separate package-names, and in that case, separate repos might be just as good a tradeoff.

msand:
At least as a library author, if I can choose to release svg based components which work with either 
a) only react-native and react-native-web projects
b) normal react projects as well
Then the choice seems obvious, there are much more plain react projects than react-native and react-native-web combined.

msand:
And I think the benefits go both ways, more react-native-svg based components work in more web projects, and more svg components for plain web can be made native compatible, essentially just by capitalizing the first letter of the tags in jsx, and importing the svgs library.

msand:
I just realized, that the same techniques used to alias react-native to react-native-web, can be used to alias react-native-web to react. So the power to override the default is still in the hands of the end user, even if we only provide a single .web.js implementation, as we're not using anything else than createElement from react-native-web anyway.

msand:
Ideally, I would like to get rid of the api differences to the web/spec. But they've been there from before I started maintaining the code, and are relatively grandfathered in. And I guess it exists mainly because there was no proper transform parser before that (v5 and earlier).

If we were to remove the differences, e.g. in a major v10 release, then this PR (plus adding the Pattern and Mask elements which exists in the latest version) should cover everything. What do you think of this? It's a relatively large breaking change, and I suspect many react-native-svg based modules would break if it was done, not to mention project code needing non-value producing churn.

The change could cause:
- 👍  a minor performance and bundle size improvement from removing the extra logic, but not enough to validate the churn, it's relatively simple js logic and only done on render so sCU can handle optimizations if needed. 
- 👍  improved alignment with the spec/web (possibly bigger impact, but still not enough to motivate the change). 
- 👎 some users might find the current api more user-friendly for their needs (speculative, but when I've accidentally broken it at times, the issues opened can seem relatively emotional at least).

It could possibly be done in a two stage process, with deprecation warnings given first, and once the community has had time to make their libraries spec conformant, the logic could be cleaned up.

Alternatively, we just merge swgs into this library, and keep the api as is. Or any other suggestions/preferences?

ide:
**swgs:** From a technical level, it sounds like the swgs library lets developers use RN or RNW -- this is the pragmatic goal in my mind. The reasons to move RNW support into RNSVG are that people already import RNSVG in their apps and I imagine most people looking for a library called RNSVG are already using RN and are the target audience for RNW. I think merging swgs's functionality into RNSVG addresses this -- the goal is for someone who already has an app that imports `react-native-svg` and uses RN APIs like StyleSheet to be able to target web without changing their code.

At the same time, components that use the svgs library won't pull in RNSVG's web implementation at all and I'd expect this PR to have no effect on svgs -- web projects will use svgs's web implementation instead.

**API:** I don't have strong opinions on the actual SVG API. I think it is useful to have a unified RN{Android, iOS, Web} API, whatever the API ends up being.

**Future:** Other libraries like all of the Foundation Unimodules are moving in the direction of targeting RN{Android, iOS, Web}. The spectrum of `[RN]<===>[RN or RNW]<===>[React DOM or RN or RNW]<===>[any browser framework or RN or RNW]` applies to most libraries and requires tooling and guides to work well over the course of several years. One day, if RN has the same API as React DOM (i.e. if RN reduces itself just to the renderer), then perhaps RNW ends up being the same as React DOM and we get to a place where targeting React means you automatically target all the platforms.

msand:
But, the cleaned up api has been something I've planned for when Fabric has been released. At which point I'd like the make the js code for the react-native implementation look much more like the code for the web implementation in this pr. That would also prepare the logic to be used as a Unimodule in e.g. Flutter, or in plain Android / iOS projects. The next step after that would be a rewrite to a cross-platform rust implementation, similar to the strategy used in [azul](https://github.com/maps4print/azul/wiki/SVG-drawing)
So, perhaps the best course of action here is to just merge swgs for now. Or does anyone have any objections or other thoughts relating to any of this?

msand:
I've refactored the implementation from swgs a bit and added it under the web branch here: https://github.com/react-native-community/react-native-svg/compare/web

I think this should cover the main use cases, and allow the end users to override if needed.

msand:
I noticed there were bugs in swgs for handling react-native-web StyleSheets, and bugs in both svgs and swgs for handling arrays of styles. Fixed both, and ensured it works in both react-native-web and plain react web projects.

msand:
proof of concepts examples:
react-native-web: https://codesandbox.io/s/q7nqlr1xq6
plain react-web: https://codesandbox.io/s/mjj41y558j

msand:
I think I'm satisfied with the implementation in the web branch for now, if anyone finds differences between the web and the native version, we'll fix it in the native code if it's a spec conformance bug, and in the web code if it's missing compatibility for some added feature from react-native-svg. I'd be ready to merge and release that at this point. I'll probably cut a release tomorrow, unless anyone comes up with any blockers.

msand:
v9.3.3 is available with a .web.js implementation now, perhaps we can close this? Or is there anything left to this?

ide:
Yes, I think we can close this in lieu of the new .web.js file.

PR #953, opened 2019-02-26
Add Svg `fill` typing

scottmas:
It already works just fine but I keep getting typescript compiler errors without adding the prop.


scottmas:
Closing in favor of https://github.com/react-native-community/react-native-svg/pull/944, which is more generalized.

PR #955, opened 2019-02-27
Update typings for vector effects

scottmas:


msand:
@scottmas Great, thanks for this! In practise there are a few more inputs which get mapped to enum values on the native side:

```jsx
const vectorEffects = {
  none: 0,
  default: 0,
  nonScalingStroke: 1,
  'non-scaling-stroke': 1,
  inherit: 2,
  uri: 3
};
```

But only (the default values, and) 'nonScalingStroke' / 'non-scaling-stroke' have an implementation, and I guess 'non-scaling-stroke' is the value used even in normal React web svg content, so this is probably good as is. Or what do you think, should we add more values to the types? 

https://www.w3.org/TR/2004/WD-SVG12-20041027/vectoreffects.html#vector-effect-prop
https://www.w3.org/TR/SVGTiny12/painting.html#NonScalingStroke
https://svgwg.org/svg2-draft/single-page.html#coords-VectorEffects

scottmas:
Updated to your implementation. If having a camelCase option is the convention used elsewhere in the project for string enums, makes sense to add these additional ones as well.

PR #958, opened 2019-03-01
Update react-native-svg-transformer documentation

kristerkari:
- Use `metro.config.js` instead of `rn-cli.config.js`.
- React Native 0.59.x ships with `metro.config.js` file, so it will be used as the default config for Metro.
- `metro.config.js` file works on React Native 0.57 and 0.58 too.


msand:
Thanks for this!

PR #1053, opened 2019-07-18
remove extra *255 in hsl conversion

rphuber:
this is currently happening on lines 437-441 when converting to the
int32Color

# Summary
when using hsl values and the saturation set to 0% the conversion from hsl -> rgb is multiplying by 255 to get an rgb value of (255, 255, 255).  This is rgb is used when creating the int32Color which then also multiplies the number by 255 again.  Removing the additional multiplication in the hslToRgb function will give consistent rgb values used to generate the int32Color.
 

## Compatibility

| OS      | Implemented |
| ------- | :---------: |
| iOS     |    ✅❌     |
| Android |    ✅❌     |

## Checklist

### DIFF
--- a/lib/extract/extractColor.js
+++ b/lib/extract/extractColor.js
@@ -371,7 +371,7 @@ function rgbFromHslString(string) {
   const l = clamp(parseFloat(match[3]), 0, 100);
   const a = clamp(isNaN(alpha) ? 1 : alpha, 0, 1);
 
-  return hslToRgb([h, s, l, a]);
+  return [...hslToRgb([h, s, l, a]), a];
 }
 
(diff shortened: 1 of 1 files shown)


rphuber:
This will also fix where alpha is being dropped when using the `toRGB` functions.

msand:
I guess this should actually multiply by 2.55 instead of 255 or no factor.

rphuber:
I'm not sure.  I think that the current way is correct.  You take the lightness percentage and divide by 100 giving you a percentage.  So say 74%/100 = .74, which is the percentage of 255 that rgb should be (255 * .74 = 188.7).

msand:
Yeah sorry for that, not sure what I was thinking, a bit too sleepy at that point. Seems correct as is, the l in hsl can be 0 to 100, divided by 100 makes it normalized.

rphuber:
Anything else you need from me to get this merged in?

msand:
No seems good like this. I'll merge the develop branch and this to make a new release soon.

rphuber:
Thank for your help with this!

PR #1136, opened 2019-10-04
Fix warning Calling bridge.imageLoader is deprecated

r0b0t3d:
# Summary

Hi, I am using `RN 0.61` and got this warning when using `Image` element on iOS.
The root cause is `self.bridge.imageLoader` has been deprecated 

https://github.com/facebook/react-native/blob/0.61-stable/React/CoreModules/RCTImageLoader.h

Simulator Screen Shot - iPhone 8 - 2019-10-04 at 18 40 13

## Compatibility

| OS      | Implemented |
| ------- | :---------: |
| iOS     |    ✅     |
| Android |    ❌     |

## Checklist

- [ ] I have tested this on a device

### DIFF
--- a/ios/Elements/RNSVGImage.m
+++ b/ios/Elements/RNSVGImage.m
@@ -40,7 +40,7 @@
         _reloadImageCancellationBlock = nil;
     }
 
-    _reloadImageCancellationBlock = [self.bridge.imageLoader loadImageWithURLRequest:[RCTConvert NSURLRequest:src] callback:^(NSError *error, UIImage *image) {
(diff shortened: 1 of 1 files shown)


msand:
@r0b0t3d Thanks for this! Have you tried this patch with any older versions of react-native? I think we might need to change the minimum supported version to merge this, and probably needs a major version bump. 

There's actually a breaking change in the develop branch already, so this probably makes sense to release at the same time.

msand:
Some of the new features in the develop branch are experimental/incomplete/broken still, but I think as long as I label them as such, I might be able to cut a release soon anyway. Would be great to get some examples/test case showing how they're broken to be able to fix them at this point, or create some synthetic but representative test scenarios. But haven't had any time to put into this lately.

msand:
This is fixed in a slightly different way in the master branch / latest version. Closing. Thanks for you pr anyway!

PR #1154, opened 2019-10-20
Remove ts-ignore comment

sairus2k:
# Summary

It removes the `@ts-ignore` comment and adds types for generated `transform.js` file.


msand:
Ah great, thanks for this!

Issue #1078, opened 2019-08-21
On Web(SPA), Unknown event handler property `onPress` click on Rect


sunildate05:
# Issue

👉  I'm using react-native-svg  9.6.4 package. onPress event works on Android & iOS. but on web,
Its giving below error - 
[Unknown event handler property `onPress`. It will be ignored.](url)

Code - 

```
const Decorator = ({ x, y, data }) => {
      return data.map((value, index) => (
        <G key={"G"+index}>
          <Rect key={"Rect"+index} x={x(index) - 10} y={y(value.value) - 10} width="23" height="23" fill="none" stroke="#ffffff" strokeOpacity={0}
           onPress={() => this.setState({ tooltipX: value.date, tooltipY: value.value, tooltipIndex: index, })}
             />
          <Circle
            key={index}
            cx={x(index)}
            cy={y(value.value)}
            r={4}
            stroke={'rgb(124,181,236)'}
            fill={'rgb(124,181,236)'}
          />
        </G>
      ))
    }
```

msand:
Can you make a full reproduction, as in a git repo I can clone, npm install, start, open in a browser and see the issue, or a codesandbox?

PR #1187, opened 2019-11-11
Fix onPress handler mapping on web elements

rodolfovilaca:
# Summary
So onPress handlers are not working on web elements like `<Path />`. I've noticed that because I am creating a graph in `react-native-svg` that works on all platforms and I needed the onPress functionality on a Path element. I've searched for some issues and saw that others are suffering the same problem:
 - [608](https://github.com/react-native-community/react-native-svg/issues/608#issuecomment-438376252)
 - [599](https://github.com/react-native-community/react-native-svg/issues/599#issuecomment-507403339) 
 - #1078

I think that this problems are related to [this line](https://github.com/necolas/react-native-web/blob/250ee3c234196fdddf6e5189c572a292566c3cb2/packages/react-native-web/src/exports/createElement/index.js#L58) that validates if the prop name is a valid prop for touchables on the web so what I did (to be as simple as possible) is to map the `onPress` handler to `onClick` handler on the web. Only impacts the web implementation.

```
obs: I don't actually know if the current behaviour is the desired one and if this is
actually a problem that should be fixed, or we won't have touchables events on path
elements.
```

No prerequisites.

Draw a simple `<Path />` element and add some onPress handler on the **web**.

## Compatibility

| OS      | Implemented |
| ------- | :---------: |
| Web     |    ✅     |
| Android |    N/A     |
| IOS |    N/A     |

## Checklist

- [x] I added the documentation in `README.md` (N/A)
- [x] I mentioned this change in `CHANGELOG.md`
- [x] I updated the typed files (TS and Flow)(N/A)
- [x] I added a sample use of the API in the example project (`example/App.js`) (N/A)

### DIFF
--- a/src/ReactNativeSVG.web.ts
+++ b/src/ReactNativeSVG.web.ts
@@ -62,6 +62,10 @@ function prepare(props) {
     clean.ref = forwardedRef;
   }
 
+  if (props.onPress && !props.onClick){
+    clean.onClick = props.onPress;
+  }
+
   const styles: {
     fontStyle?: string;
     fontFamily?: string;


msand:
Great, thanks for this. You can remove the note from the changelog, it's created automatically, I should update the pr template not to include it in instructions. Additionally, could you check that there's no onClick property from before? In case someone has different event handlers for web and native, then it would be better not to override the existing web handler.

rodolfovilaca:
For sure, I made the changes you requested. Should we try to map other web events? `onPressIn` and `onPressOut` like in the docs? I mean we could possibly map them to other web events(`onTouchEnd`, `onTouchStart`), but maybe we should discuss them idk...

msand:
Looks good for now, can make another pr if more handlers are needed. Thanks!

mikeaustin:
View has onLayout, but it looks like SVG components do not? Where should I look for adding that support?

PR #1197, opened 2019-11-19
add missing imports in web root file

billouboq:
Hi, when working on web version of our app we just noticed some exports where missing in the web version.

This PR just add the missing ones.

### DIFF
--- a/src/ReactNativeSVG.web.ts
+++ b/src/ReactNativeSVG.web.ts
@@ -2,6 +2,14 @@ import { createElement } from 'react-native-web';
 import { resolve } from './lib/resolve';
 import { Component } from 'react';
 import { NumberProp } from './lib/extract/types';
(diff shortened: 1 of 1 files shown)


msand:
I would recommend using an img tag and set the svg content as a base64 encoded string, I'd expect the built-in parsers and css engines in browsers to be more efficient than this implementation.

msand:
Alternatively, if you have a url, just use that with an img tag. Would you mind defining new components which do that, on the web, rather than process everything in js?

billouboq:
Yeah of course, will do that ! thanks for your reply

msand:
Actually don't need base64: https://css-tricks.com/probably-dont-base64-svg/

billouboq:
Ok closing this, thanks

PR #1203, opened 2019-11-28
fix: initialize PathView with empty path to support reanimated on Android

jayu:
# Summary

I was trying to animate `<Path/>` `d` attribute with `react-native-reanimated`. On iOS it worked well, but on Android, it was rising null pointer exception and the app was crashing. I found that `mPath` is null when it's trying to be used. For some reason, the `reanimated` nodes didn't manage to initialize on time, and initial `setD` was invoked after the initial render of the path. Setting default path value to an empty path solves the issue.

PS: It would be awesome to release this patch before Expo releases SDK 36, so new fixed version of `react-native-svg` will be included 🙏 

`react-native-reanimated` and `react-native-svg` installed in the project

Following code used to fail on Android, works well on iOS
It's a minimal repro, so no visible path will be rendered.
```js
import React from 'react';
import {Svg, Path} from 'react-native-svg';
import Reanimated from 'react-native-reanimated';

const AnimatedPath = Reanimated.createAnimatedComponent(Path);

export default class MyExample extends React.PureComponent {
  render() {
    return (
      <Svg width={10} height={10}>
        <AnimatedPath
          d={new Reanimated.Value('')} // this used to fail on Android. Any other value fails too.
        />
      </Svg>
    );
  }
}
```

## Compatibility

| OS      | Implemented |
| ------- | :---------: |
| iOS     |    n/a     |
| Android |    ✅     |

## Checklist

### DIFF
--- a/android/src/main/java/com/horcrux/svg/PathView.java
+++ b/android/src/main/java/com/horcrux/svg/PathView.java
@@ -24,6 +24,7 @@ class PathView extends RenderableView {
     public PathView(ReactContext reactContext) {
         super(reactContext);
         PathParser.mScale = mScale;
+        mPath = new Path();
     }
 
     @ReactProp(name = "d")


jayu:
Hi @msand Can you take a look on that? At @callstack we are planning to release small lib with Animated Arc component and It would be awesome to have a native performance on both android and ios :D

msand:
@jayu  Sure, I'll try to make some time available for it today. Very busy at work atm. Would you & others at callstack be interested in taking a more active role in maintaining this library? I haven't actively worked on any react-native app for 1.5 years now (mostly react-native-web atm)... Feels like it might be about time to have more maintainers in this relatively central piece of the ecosystem.

jayu:
@msand I will ask other folks at Callstack. I'm making first steps in native development, but maybe it will be a great opportunity to go deeper into it :D

Esemesek:
@msand That would be the perfect opportunity to officially deprecate `react-native-art` (since I am maintaining it) and replace it with `react-native-svg`.

msand:
I think react-native-art might still be useful for some high performance things, it needs to support and do much less than what the svg spec says. 

E.g. when animating many instances of simple paths art can be quite fast if the graphics are possible to draw reasonably using it. In svg, symbols, use elements, style inheritance, clipping, masking, (filtering wip) etc all have additional requirements vs the logic in art, which can limit itself to almost only creating the path objects and rendering them straight. 

The extra functionality makes more advanced use cases easier to create, but has computational costs. 
Especially on android, there are use cases where art is smooth but react-native-svg is not yet.

I'd love to see http://sluglibrary.com/ available for open use, or see someone try the approaches set out in https://github.com/react-native-community/react-native-svg/issues/297#issuecomment-518640832 possibly as a separate library from this, as the current and any new rendering backends can have quite different tradeoffs

And perhaps some of the experiments from here: https://github.com/react-native-community/react-native-svg/issues/878

Also, with the fabric/jsi rewrite there's several optimisation opportunities available. I'd also like to see the logic in the javascript side made minimal, essentially just passing the inputs straight to the native side, and have a single cross-platform native implementation for the rendering engine. These components could be generated straight from the svg idl spec, and the native marshalling / binding code generated from that with the fabric approach.

PR #1206, opened 2019-11-28
Fix middle-anchoring for lines of multiple TSpans

mjmasn:
# Summary

Fixes https://github.com/react-native-community/react-native-svg/issues/1205

Not really intended as the final fix for this, I only dabble in Java so I'm sure there is probably a better way to do this. Happy to make changes as required :)

Clone the repo here: https://github.com/mjmasn/TSpanIssue
Install react-native-svg from git branch / yarn link etc.

Run the app before and after installing this patch.

## Compatibility

| OS      | Implemented |
| ------- | :---------: |
| iOS     |    n/a (issue not tested on iOS)     |
| Android |    ✅     |

## Checklist

- [x] I have tested this on a device 
- [ ] and a simulator

### DIFF
--- a/android/src/main/java/com/horcrux/svg/TSpanView.java
+++ b/android/src/main/java/com/horcrux/svg/TSpanView.java
@@ -213,6 +213,35 @@ class TSpanView extends TextView {
             return advance;
         }
 
+        ViewParent parent = getParent();
+
+        if (parent instanceof TextView) {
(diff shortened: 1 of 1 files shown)


mjmasn:
Hmm, so this doesn't actually work if the spans have different `fontSize`s. Can't seem to work out how to get the font for each span from `getSubtreeTextChunksTotalAdvance` of another span.

PR #1212, opened 2019-12-05
Fixes readme

moay:
# Summary

This line in the readme triggers a warning when running tests with jest. It should be updated to the new, correct import as promoted by expo.

It's just a readme update, no further checks and tests should be necessary. :-)


msand:
Thanks for this!

PR #1217, opened 2019-12-09
fix memory font memory leaks iOS

adamczyk777:
# Summary

When using React Native SVG to draw a simple chart with some texts I noticed that our bug reporting system reports Out Of Memory exceptions on iOS. After some investigation, I noticed there was a horrendous amount of memory leaks each time I entered a screen with a chart. The memory could easily spike when fast switching application tabs. I check what was happening in Xcode memory profiler and memory leaks lead to RNSVG. After some investigation, I noticed that some of those leaks were fixed around 9.12 release, but there were still some memory leaks. So I went through traces what was causing it and a little bit through code and ended up with a solution that results in no memory leaks from RNSVG. The solution does one release of reference. And for fonts `CFAutorelease` because I was not able to track all the places where I can release memory by myself. If some of the maintainers know how to do it better, I would be glad if you either pointed me how to do it or simply opened a PR to my fork.

## Test Plan

Make yourself a chart (I used react-native-svg-charts) that has text labels in it. And take a look at the memory leaks.

Some react native app and preferably React Native SVG Charts library.

* Make a chart
* Build release app in XCode
* Open Debug tab
* Go to memory Tab
* click transfer to instruments button
* now take a look at the second trace which shows checkmarks if there are no leaks or red crosses when there are leaks. 

## Compatibility

| OS      | Implemented |
| ------- | :---------: |
| iOS     |    ✅     |
| Android |    ❌     |

## Checklist

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,3 +1,10 @@
+## [9.13.3](https://github.com/react-native-community/react-native-svg/compare/v9.13.2...v9.13.3) (2019-10-25)
+
+
+### Bug Fixes
+
+* **android:** Image onPress [#1170](https://github.com/react-native-community/react-native-svg/issues/1170) ([5c967b7](https://github.com/react-native-community/react-native-svg/commit/5c967b7c6992246ebd602e9a86d2544f0662ce10))
+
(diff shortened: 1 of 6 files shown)


adamczyk777:
As for the checks, I am not sure If some of them are required. If I have to check them just let me know.

msand:
Looks good, thanks for this!

PR #1242, opened 2020-01-06
Added support for ForeignObject

### LINKED ISSUE #447: <ForeignObject> support
Forgive me if this is covered, but I've searched the documentation and the issues and see no reference to this.

Is `<foreignObject>` supported or on the roadmap?

SaeedZhiany:
# Summary
This PR adds `ForeignObject` support to this library and resolve #447 

## Test Plan

It can be used like this:

```js
<View style={{flex: 1, justifyContent: 'center'}}>
  <Svg height="50%">
    <Defs>
      <LinearGradient
        id="Gradient"
        gradientUnits="userSpaceOnUse"
        x1="0"
        y1="0"
        x2="800"
        y2="0">
        <Stop offset="0" stopColor="white" stopOpacity="0.2" />
        <Stop offset="1" stopColor="white" stopOpacity="1" />
      </LinearGradient>
      <Mask
        id="Mask"
        maskUnits="userSpaceOnUse"
        x="0"
        y="0"
        width="800"
        height="300">
        <Rect x="0" y="0" width="800" height="300" fill="url(#Gradient)" />
      </Mask>
    </Defs>
    <G mask="url(#Mask)">
      <Circle cx={50} cy={70} r={65} />
      <ForeignObject x={50} y={50}>
        <View style={{width: 200, height: 400, flex: 1}}>
          <Image
            style={{width: 200, height: 200}}
            source={{
              uri: 'https://picsum.photos/200/200',
            }}
          />
        </View>
        <ForeignObject x={50} y={50}>
          <View style={{transform: []}}>
            <Text style={{color: 'blue'}}>Testing</Text>
            <Text style={{color: 'green'}}>Testing2</Text>
          </View>
        </ForeignObject>
      </ForeignObject>
    </G>
  </Svg>
</View>
```

## Compatibility

| OS      | Implemented |
| ------- | :---------: |
| iOS     |    ✅     |
| Android |    ✅     |

## Checklist

- [x] I updated the typed files (TS and Flow)

### DIFF
--- a/android/src/main/java/com/horcrux/svg/ForeignObjectView.java
+++ b/android/src/main/java/com/horcrux/svg/ForeignObjectView.java
@@ -10,6 +10,11 @@
 package com.horcrux.svg;
 
 import android.annotation.SuppressLint;
+import android.graphics.Canvas;
+import android.graphics.Paint;
+import android.view.View;
+
+import androidx.annotation.NonNull;
 
 import com.facebook.react.bridge.Dynamic;
 import com.facebook.react.bridge.ReactContext;
@@ -25,6 +30,23 @@ class ForeignObjectView extends GroupView {
 
     public ForeignObjectView(ReactContext reactContext) {
         super(reactContext);
+    }
+
+    @Override
+    void draw(Canvas canvas, Paint paint, float opacity) {
+        float x = (float)relativeOnWidth(mX);
+        float y = (float)relativeOnHeight(mY);
+        float w = (float)relativeOnWidth(mW);
+        float h = (float)relativeOnHeight(mH);
+        canvas.translate(x, y);
+        canvas.clipRect(0, 0, w, h);
+        super.draw(canvas, paint, opacity);
+    }
+
+    @Override
+    public void onDescendantInvalidated(@NonNull View child, @NonNull View target) {
+        super.onDescendantInvalidated(child, target);
+        invalidate();
     }
 
     @ReactProp(name = "x")
--- a/android/src/main/java/com/horcrux/svg/ForeignObjectView.java
+++ b/android/src/main/java/com/horcrux/svg/ForeignObjectView.java
@@ -1,0 +1,53 @@
+/*
+ * Copyright (c) 2015-present, Horcrux.
+ * All rights reserved.
+ *
+ * This source code is licensed under the MIT-style license found in the
+ * LICENSE file in the root directory of this source tree.
+ */
+
+
+package com.horcrux.svg;
+
+import android.annotation.SuppressLint;
+
+import com.facebook.react.bridge.Dynamic;
+import com.facebook.react.bridge.ReactContext;
+import com.facebook.react.uimanager.annotations.ReactProp;
+
+@SuppressLint("ViewConstructor")
+class ForeignObjectView extends GroupView {
+
+    SVGLength mX;
+    SVGLength mY;
+    SVGLength mW;
+    SVGLength mH;
+
+    public ForeignObjectView(ReactContext reactContext) {
+        super(reactContext);
+    }
+
+    @ReactProp(name = "x")
+    public void setX(Dynamic x) {
+        mX = SVGLength.from(x);
+        invalidate();
+    }
+
+    @ReactProp(name = "y")
+    public void setY(Dynamic y) {
+        mY = SVGLength.from(y);
+        invalidate();
+    }
+
+    @ReactProp(name = "width")
+    public void setWidth(Dynamic width) {
+        mW = SVGLength.from(width);
+        invalidate();
+    }
+
+    @ReactProp(name = "height")
+    public void setHeight(Dynamic height) {
+        mH = SVGLength.from(height);
+        invalidate();
+    }
+}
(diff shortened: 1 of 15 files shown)


SaeedZhiany:
@msand 

currently, I'm getting this error
image

SaeedZhiany:
@msand

currently, I'm getting this error
Screenshot_1578390972

msand:
Did you run pod install ? And rebuild the app?

SaeedZhiany:
I haven't tested it on mac yet, this is Android app. and yes I reinstall the app

msand:
Do the files show up in android studio?

SaeedZhiany:
yes

image

SaeedZhiany:
Does `ForeignObject` need to override `saveDefinition` method?

SaeedZhiany:
Ah, I think I found the problem, I forgot to add `ForeignObjectManager` instance in `createViewManagers` method

SaeedZhiany:
Unfortunately, result is not what we want

image

msand:
Ah, yeah, it probably needs some logic for clipping to the given bounds. And also inform yoga about the layout, or at least the width and height of the foreignElement, otherwise it won't be able to know how much it can flex / what percentages should refer to.

SaeedZhiany:
But it extends G, so it should mimic `G`'s behavior. is it possible to convert `x` and `y` properties to `translate(x, y)`? because this library already can support foreignObject but using `G` syntax, we should try to build `<G translate(x, y)>` using `<ForeignObject x={} y={}>`, am I right?

msand:
Well, actually, the G element doesn't have and y prop in the spec, any elements which have it should interpret it according to spec, although in this case it should behave exactly the same. The G element isn't specified to clip anything, the foreignobject is, and the spec is more the source of truth than what happens to inherit from what in the implementation.

SaeedZhiany:
I'm not sure I could understand your mean or not. already we can nested any native view withing `G`, right?

```
<G translate(x, y)>
  <View>
    <Text>text1</Text>
    <Text>text2</Text>
    <Image/>
    ...
  </View>
</G>
```

and we want something like that:

```
<ForeignObject x={x} y={y}>
  <View>
    <Text>text1</Text>
    <Text>text2</Text>
    <Image/>
    ...
  </View>
</ForeignObject>
```

so I'm thinking of a way to convert `ForeignObject`'s `x` and `y` to `translate` property of `G`, isn't it possible? or I'm missing something else?

msand:
For that it should be enough to replace 'withoutXY' with 'extract' in this one line: https://github.com/react-native-community/react-native-svg/pull/1242/files#diff-69a8312d4b83055c5e0312ee1cecbb09R28

msand:
Also, foreignObjectProps should extend GProps https://github.com/react-native-community/react-native-svg/pull/1242/files#diff-b739f95ace6bbf2f0fd5e8cf1dab97d8R262

msand:
Actually, better keep the withoutXY there, otherwise this will cause strange interaction when animating both x and y props and transform styles. The logic for handling the x and y props should be handled in the native side, overriding the rendering logic of G slightly. I can take a look at it later. Until then you can probably use the transform prop or style directly.

SaeedZhiany:
So it seems the logic is more complicated than I think and unfortunately I haven't enough knowledge about that right now. Thanks for your notes. 👍

msand:
Seems this renders the same across android ios and web now:

```jsx
import React, { Component } from "react";
import { Text, View, Image } from "react-native";
import {
  Svg,
  Defs,
  LinearGradient,
  Stop,
  Mask,
  Rect,
  G,
  Circle,
  ForeignObject
} from "react-native-svg";

  render() {
    return (
      <View style={{ flex: 1, justifyContent: "center" }}>
        <Svg height="50%">
          <Defs>
            <LinearGradient
              id="Gradient"
              gradientUnits="userSpaceOnUse"
              x1="0"
              y1="0"
              x2="800"
              y2="0"
            >
              <Stop offset="0" stopColor="white" stopOpacity="0.2" />
              <Stop offset="1" stopColor="white" stopOpacity="1" />
            </LinearGradient>
            <Mask
              id="Mask"
              maskUnits="userSpaceOnUse"
              x="0"
              y="0"
              width="800"
              height="300"
            >
              <Rect
                x="0"
                y="0"
                width="800"
                height="300"
                fill="url(#Gradient)"
              />
            </Mask>
          </Defs>
          <G mask="url(#Mask)">
            <Circle cx={50} cy={70} r={65} />
            <ForeignObject x={50} y={0} width={100} height={100}>
              <View style={{ width: 200, height: 400, transform: [] }}>
                <Image
                    style={{ width: 200, height: 200 }}
                    source={{
                      uri: "https://picsum.photos/200/200"
                    }}
                />
              </View>
            </ForeignObject>
            <ForeignObject x={50} y={0} width={100} height={100}>
              <View style={{ width: 200, height: 400, transform: [] }}>
                <Text style={{ color: "blue" }}>Testing</Text>
                <Text style={{ color: "green" }}>Testing2</Text>
              </View>
            </ForeignObject>
          </G>
        </Svg>
      </View>
    );
  }
}

```

Screenshot 2020-01-11 at 21 34 10

PR #1245, opened 2020-01-08
Delete regression that make every component accessible by default

zoontek:
# Summary

Fixes https://github.com/react-native-community/react-native-svg/issues/1244

1. Init a project with `create-react-app` and `react-native-web`
2. Add this fork of `react-native-svg` as a dependency and add a random SVG using it without setting `accessible` prop

1. Try to focus SVG elements: it will not be possible unless you enforce `accessible={true}`

## Compatibility

| OS      | Implemented |
| ------- | :---------: |
| iOS     |    ❌     |
| Android |    ❌     |

## Checklist

### DIFF
--- a/src/ReactNativeSVG.web.ts
+++ b/src/ReactNativeSVG.web.ts
@@ -66,7 +66,6 @@ const prepare = <T extends BaseProps>(
   props = self.props,
 ) => {
   const {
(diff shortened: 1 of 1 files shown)


msand:
lgtm

PR #1249, opened 2020-01-14
fix(android): animated stroked path with useNativeDriver: true

HelloCore:
# Summary

1. When animating `Path` with `d` attribute and `nativeDriver:true`, React sometimes sets `null` to the property, and it got crashed. (Still happens on `RN 0.61.4` and `react-native-svg 10.1.0`)
https://github.com/react-native-community/react-native-svg/issues/951#issuecomment-487968402
https://github.com/react-native-community/react-native-svg/issues/951#issuecomment-488825825

2. When animating `stroke` on Android, the color seems to be wrong, so I modified it by copying the implementation detail from `fill`

## Test Plan

Try animating stroke, and path, using `useNativeDriver:true`
https://gist.github.com/HelloCore/bafdf02644012bb47d85e828dd734e41

## Compatibility

| OS      | Implemented |
| ------- | :---------: |
| Android |    ✅     |

## Checklist

### DIFF
--- a/android/src/main/java/com/horcrux/svg/PathParser.java
+++ b/android/src/main/java/com/horcrux/svg/PathParser.java
@@ -33,8 +33,11 @@ class PathParser {
 
     static Path parse(String d) {
         elements = new ArrayList<>();
-        char prev_cmd = ' ';
         mPath = new Path();
+        if(d == null){
+            return mPath;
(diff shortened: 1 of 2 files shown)


msand:
Lgtm, thanks for this!

PR #1251, opened 2020-01-15
Fix/ios crash animated path a

HelloCore:
# Summary
When animating `Path` using `A` with different `outer` and `nativeDriver:true`, sometimes the value of `outer` could be between 0 - 1, and it will crash.
alt text

## Test Plan

https://gist.github.com/HelloCore/725c5be62d9b29b4c9b23d9c0f5e0a61

## Compatibility

| OS      | Implemented |
| ------- | :---------: |
| iOS     |    ✅     |

## Checklist


msand:
Seem you're trying to interpolate a number linearly, where you need a step function. I cannot support merging this, this is not syntax which is part of the svg spec. Rather a logic bug in your animation.

PR #1255, opened 2020-01-18
perf: optimize extraction of fill, stroke, responder, matrix & display

msand:
potentially breaking change
if any issues show up, then they'll be considered regressions
should be resolved in ways which cause less traffic from js to native

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,3 +1,15 @@
+## [11.0.1](https://github.com/react-native-community/react-native-svg/compare/v11.0.0...v11.0.1) (2020-01-18)
+
+
+### Bug Fixes
+
+* snapshot ([a37afcb](https://github.com/react-native-community/react-native-svg/commit/a37afcb485b8804dbc212195bc5695738a3634f7))
+
+
+### Performance Improvements
+
+* optimize passing of colors to native ([f138c9b](https://github.com/react-native-community/react-native-svg/commit/f138c9b1aba568ac3ac0c272af1d62811519f526))
+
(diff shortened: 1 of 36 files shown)


msand:
Ideally we would get rid of empty font and proplist objects as well.

msand:
@msageryd @wcandillon Would you mind testing this for performance changes and regressions?

msand:
@SaeedZhiany @maxkomarychev @iqqmuT @kristerkari 🙏Could you help out as well? I think large svgs will notice the largest performance difference, probably most on low-end android. Especially if there are many G elements.

msageryd:
@msand I tried 0fa4177 in my app and found no regressions. As of now I cannot tell anything about performance. I'll give it a go with a larger dataset in a couple of days. This is roughly what I tested:
- a couple of 100 elements on screen
- panning/zooming via setNativeProps (matrix)
- Moving and rotating elements
- Extensive use of caching with ``<Symbol>``

msand:
Ah great, good to know. I'd expect that it requires really large svg documents to make it noticeable by eye, unless the device is really constrained in available cpu and memory. Might be that the difference is most noticeable in memory consumption. Should be at most a few percentage difference either way, unless one makes some kind of worst case example to exaggerate the difference, most of the work is still in rendering. But this also make e.g. browsing the element hierarchy much nicer in react native debugger, as there's much less noise now. 

I'd expect regressions to be mostly related to inheritance of fill and stroke, and possibly some geometry / transform related things, if any part is still relying on the transform props being extracted and sent using the extractProps helper.

PR #1259, opened 2020-01-22
[chore] Fix podspec/gitignore to work with Bazel builds (iOS)

kexin-zhang:
# Summary

Hello! In order to manually link react-native-svg in an existing iOS project that uses Bazel, I have added two changes:
* Change `build` to `build/` in `.gitignore`. This doesn't add any files (it should still exclude Android gradle files), but it allows us to include our generated Bazel BUILD file when we build our app.
* Add `React/RCTImage` as a dependency to the podspec. This allows the package to build without errors (was originally getting the issue `'React/RCTImageLoader.h' file not found` as described in the README).

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -26,7 +26,7 @@ project.xcworkspace
 .idea
 .gradle
 local.properties
-build
+build/
 
 # node.js
 #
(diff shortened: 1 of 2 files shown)


msand:
Lgtm, thanks for this!

PR #1265, opened 2020-01-27
fix: removed missing unnecessary React headers import error caused by…

alechill:
… non-framework style import

# Summary

Latest release (11.0.1) causes a compilation error by attempting to use a non framework import of `RCTEventDispatcher.h` in `RNSVGForeignObject.h`

image

This is fixable using framework style import `#import <React/RCTEventDispatcher.h>` however this is unnecessary as it is no longer used in any of the signatures, so have removed instead

## Test Plan

Built and ran locally in dependant project

Build fails using 11.0.1

## Compatibility

| OS      | Implemented |
| ------- | :---------: |
| iOS     |    ✅     |
| Android |    n/a     |

## Checklist

### DIFF
--- a/ios/Elements/RNSVGForeignObject.h
+++ b/ios/Elements/RNSVGForeignObject.h
@@ -1,7 +1,6 @@
 
 #import "RNSVGGroup.h"
 #import "RNSVGLength.h"
-#import "RCTEventDispatcher.h"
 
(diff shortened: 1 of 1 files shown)


SaeedZhiany:
@msand 

please review this. 

Thanks

msand:
Lgtm, thanks for this!

PR #1281, opened 2020-02-11
chore: Refactored Gradle setup

SaeedZhiany:
# Summary

- moved safeExtGet to out of `buildscript` block, because it's no longer needed to be there (previously we needed it there for getting AGP version from root project)

- bumped AGP version to 3.5.3

- replace `${safeExtGet('reactnativeVersion', '+')}` with `+`, because react-native version automatically resolved by npm and it does not need to handled using `safeExtGet`

## Compatibility

| OS      | Implemented |
| ------- | :---------: |
| iOS     |    ✅     |
| Android |    ✅     |

## Checklist


msand:
Lgtm, thanks for this!

PR #1294, opened 2020-02-25
fix: [web] Don't apply SvgTouchableMixin on every component

zoontek:
Hi 👋

Do you remember [this bug](https://github.com/react-native-community/react-native-svg/issues/1244)? Well, it happens again, for a different reason. 😄

By default, every component of this lib implements the `SvgTouchableMixin`, even if no Touchable properties are used, making every component accessible by default again.

# Summary

- `hasTouchableProperty` is now a function, `SvgTouchableMixin` is applied only if it returns `true`.

1. Init a project with `create-react-app` and `react-native-web`
2. Add `react-native-svg` dependency and render a random SVG using it

1. SVG elements should not be accessible and focusable if Touchable props are not present.
2. SVG elements should be accessible and focusable if Touchable props are present.

## Checklist

- [X] I have tested this on a browser

### DIFF
--- a/.eslintrc.js
+++ b/.eslintrc.js
@@ -41,7 +41,10 @@ module.exports = {
         'ts-expect-error': 'allow-with-description',
       },
     ],
-    '@typescript-eslint/no-unused-vars': ['error', { argsIgnorePattern: '^_' }],
+    '@typescript-eslint/no-unused-vars': [
+      'error',
+      { argsIgnorePattern: '^_', ignoreRestSiblings: true },
+    ],
     '@typescript-eslint/no-var-requires': 'warn',
     eqeqeq: 'error',
     'no-unreachable': 'error',
--- a/src/web/WebShape.ts
+++ b/src/web/WebShape.ts
@@ -1,6 +1,6 @@
 import React, { type JSX } from 'react';
 import {
-  GestureResponderEvent,
+  TouchableWithoutFeedback,
   // @ts-ignore it is not seen in exports
   unstable_createElement as createElement,
 } from 'react-native';
@@ -8,9 +8,27 @@ import {
 import { BaseProps } from './types';
 import { prepare } from './utils/prepare';
 import { convertInt32ColorToRGBA } from './utils/convertInt32Color';
-import { getAttributeName, remeasure } from './utils';
+import { getAttributeName } from './utils';
 import { hasTouchableProperty } from './utils/hasProperty';
-import SvgTouchableMixin from '../lib/SvgTouchableMixin';
+
+/**
+ * `TouchableWithoutFeedback` clones its child and injects props on it
+ * (`focusable`, `accessibilityDisabled` and the responder handlers). Those props
+ * need `unstable_createElement` to be turned into DOM attributes, so the child
+ * has to be a component that calls it, not an already created element.
+ */
+const TouchableSvgElement = React.forwardRef<
+  SVGElement,
+  {
+    tag: React.ElementType;
+    elementProps: Record<string, unknown>;
+    children?: React.ReactNode;
+  }
+>(({ tag, elementProps, ...injectedProps }, ref) =>
+  createElement(tag, { ...elementProps, ...injectedProps, ref })
+);
+
+TouchableSvgElement.displayName = 'TouchableSvgElement';
 
 export class WebShape<
   P extends BaseProps = BaseProps,
@@ -83,30 +101,6 @@ export class WebShape<
     }
   }
 
-  _remeasureMetricsOnActivation: () => void;
-  touchableHandleStartShouldSetResponder?: (
-    e: GestureResponderEvent
-  ) => boolean;
-
-  touchableHandleResponderMove?: (e: GestureResponderEvent) => void;
-  touchableHandleResponderGrant?: (e: GestureResponderEvent) => void;
-  touchableHandleResponderRelease?: (e: GestureResponderEvent) => void;
-  touchableHandleResponderTerminate?: (e: GestureResponderEvent) => void;
-  touchableHandleResponderTerminationRequest?: (
-    e: GestureResponderEvent
-  ) => boolean;
-
-  constructor(props: P) {
-    super(props);
-
-    // Do not attach touchable mixin handlers if SVG element doesn't have a touchable prop
-    if (hasTouchableProperty(props)) {
-      SvgTouchableMixin(this);
-    }
-
-    this._remeasureMetricsOnActivation = remeasure.bind(this);
-  }
-
   render(): JSX.Element {
     if (!this.tag) {
       throw new Error(
@@ -114,9 +108,63 @@ export class WebShape<
       );
     }
     this.lastMergedProps = {};
-    return createElement(
-      this.tag,
-      prepare(this, this.prepareProps(this.props))
+
+    const cleanProps = prepare(this, this.prepareProps(this.props));
+
+    // Only touchable elements are wrapped, so only they become keyboard
+    // focusable. A plain shape stays out of the tab order.
+    if (!hasTouchableProperty(this.props)) {
+      return createElement(this.tag, cleanProps);
+    }
+
+    const { children, ref, ...elementProps } = cleanProps as Record<
+      string,
+      unknown
+    >;
+
+    const {
+      delayLongPress,
+      delayPressIn,
+      delayPressOut,
+      disabled,
+      focusable,
+      onBlur,
+      onFocus,
+      onLongPress,
+      onPress,
+      onPressIn,
+      onPressOut,
+      rejectResponderTermination,
+    } = this.props;
+
+    return React.createElement(
+      TouchableWithoutFeedback,
+      {
+        delayLongPress,
+        delayPressIn,
+        delayPressOut,
+        disabled,
+        // A touchable element is a keyboard tab stop unless it opts out with
+        // `focusable={false}`.
+        focusable,
+        onBlur,
+        onFocus,
+        onLongPress,
+        onPress,
+        onPressIn,
+        onPressOut,
+        rejectResponderTermination,
+      },
+      React.createElement(
+        TouchableSvgElement,
+        {
+          tag: this.tag,
+          elementProps,
+          // TouchableWithoutFeedback merges this ref with its own host ref
+          ref: ref as React.Ref<SVGElement>,
+        },
+        children as React.ReactNode
+      )
     );
   }
 }
--- a/src/web/types.ts
+++ b/src/web/types.ts
@@ -22,6 +22,7 @@ export interface BaseProps {
   delayPressIn?: number;
   delayPressOut?: number;
   disabled?: boolean;
+  focusable?: boolean;
   hitSlop?: EdgeInsetsProp;
   href?: RNImageProps['source'] | string | number;
   nativeID?: string;
--- a/src/web/utils/index.ts
+++ b/src/web/utils/index.ts
@@ -12,39 +12,6 @@ export const getBoundingClientRect = (node: SVGElement) => {
   throw new Error('Can not get boundingClientRect of ' + node || 'undefined');
 };
 
-const measureLayout = (
-  node: SVGElement,
-  callback: (
-    x: number,
-    y: number,
-    width: number,
-    height: number,
-    left: number,
-    top: number
-  ) => void
-) => {
-  const relativeNode = node?.parentNode;
-  if (relativeNode) {
-    setTimeout(() => {
-      // @ts-expect-error TODO: handle it better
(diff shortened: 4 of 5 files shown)


msand:
Oh interesting, I suspect this won't handle cases where at first there's no touchable property, but in a later re-render there is. Wonder what makes it accessible to begin with, have you dug deeper into this?

zoontek:
@msand Yes! By applying `TouchableMixin` [this exact function](https://github.com/necolas/react-native-web/blob/master/packages/react-native-web/src/exports/StyleSheet/modality.js#L58) will return `true`, making every component of this lib keyboard focusable.

msand:
Hmm, I can't seem to figure out why that return value would depend on using TouchableMixin, can't find anything inside it that should effect it. Guess I'd have to run the actual code to figure it out.

zoontek:
@msand Instead of using a `TouchableMixin`, we might conditionally wrap the rendered element with a `TouchableWithoutFeedback`, no?

msand:
I think that probably wont work well together with gesture responders. Not sure though.

msand:
This seems to work:

```jsx
const isWeb = Platform.OS === 'web';
const keys = Object.keys(SvgTouchableMixin);
const touchKeys = isWeb ? keys.filter(key => key !== 'componentDidMount') : keys;
```

msand:
Seems related to this._isTouchableKeyboardActive in Touchable Mixin

msand:
Actually, deciding factor seems to be:
this._touchableNode.addEventListener('blur', this._touchableBlurListener);

msand:
Not sure how the presence / absence of a blur handler decides if a svg element can become the active element or not.

msand:
I see

https://svgwg.org/svg2-draft/interact.html#Focus

> User agents may treat other elements as focusable, particularly if keyboard interaction is the only or primary means of user input. In particular, user agents may support using keyboard focus to reveal ‘title’ element text as tooltips, and may allow focus to reach elements which have been assigned listeners for mouse, pointer, or focus events. Authors should not rely on this behavior; all interactive elements should directly support keyboard interaction.

msand:
@necolas Is this behaviour intentional? Is the blur event handler partially there to make svg some elements focusable? What are the cases in which they should they be focusable and not? 

I'd assume if there's no focus/press/gesture related handlers, they shouldn't be focusable, while if that changes, it should adapt correctly. 

Seems like disabling the blur event listener / skipping componentDidMount is not a good solution. 

Could perhaps attach/detach as a side-effect based on the presence of event handlers. Wdyt?

msand:
@necolas Sorry to ping you twice like this. Feel free to instruct me not to ping you and I won't cause any further interruptions. Was suspecting it went past your radar / forgot about this. Would you mind having a quick look and giving a bit of guidance with regards to the touchable api in react-native-web? I'm considering I should write a react-native-svg specific handler, that de/attaches the blur event listener on demand. Do you see any problems with this approach?

msand:
@zoontek What do you think of this implementation?

zoontek:
@msand This should work, I can try it tomorrow to be sure if you want. But this (kind of) seems fragile. Maybe `react-interactions` will help at some point? For now the implementation target React DOM, but I think it's their solution to get rid of Touchable Mixins at some point.

msand:
Yeah, this certainly isn't very clean, but the closest api to the react-native implementation and most react-native-web compatible I can think of at the moment. I'd love to see a cleaner implementation of this and certainly willing to merge it.

msand:
I think the mixin should be run on the prototype once rather than on each instance object as well. Only the event listeners need to be attached, and the state initialized, per instance.

necolas:
I don't have enough context to understand what's going on here. You're trying to mixin Touchable on an SVG element? Why is that part of an svg library?

msand:
It's used in the react-native implementation to get support for touch/gesture events, so quite a few components are implemented based on that api of react-native-svg, and thus a compatibility layer for the web components has been requested.

msand:
@necolas Perhaps I should elaborate, the Touchable mixin has been part of the (Shape) api since v4.1.0 or 9 Jun 2016 
https://github.com/react-native-community/react-native-svg/commit/73077157124c84cfa87c00055b96cfb697288d9d#diff-8a2f6ca181e875c2bffa1bebff2800b2 

Many of the use cases of this library aren't static content, but interactive visualisations and complex ui elements with various touchable filled and/or stroked areas. Many of them are simple press handlers, others have turing complete dependence on the history of gesture events / other inputs.

In the case of this pr, as we've made the api work the same way on the web as it already does in native (i.e. support touch and gesture handler directly on the components), causes the dom elements to be focusable, as the componentDidMount handler from Touchable in react-native-web attaches a blur event handler to the dom elements, causing them to be considered accessible elements, and thus when you tab thru the page, the elements receive focus, even if they don't have any press / gesture event handlers actively attached to the react-native-svg elements.

This pr currently detaches the blur event handler after componentDidMount, in case there are no touchable props given, and reattaches if at a later point there is. Ideally we would have had two sets of elements, one with touchable support, and one without, but this api was grandfathered in before I made my first contributions.

@zoontek Did you have time to test this?

zoontek:
@msand Yes, I use the last version on two projects and it works correctly now.

Otherwise if I understand the necessity of interactive SVG, why can't we just wrap elements in TouchableWithoutFeedback? (Ex: to make a Path pressable). It's added complexity and possible side effects for not so much.

msand:
@zoontek You mean the last released version? last commit from develop branch? or from this branch/pr?

That would be ideal, but I'm not sure I could sustain the amount of opened issues that kind of breaking change to the api would cause. I would likely stop supporting this library, given the history of types of issues opened over the past few years. So, perhaps in case you would commit to handling all issues of that type I could consider it.

zoontek:
My bad, I was thinking about the other issue I had (height & width at `100%` by default). I forgot to try this one.

I installed it on my project (using this branch, up-to-date) and all the elements are still focusables.

colinhostetter:
Hey, is there any plan to merge this? I'm encountering the same problem (all SVG elements being tabbable) that this is discussing.

src/ReactNativeSVG.web.ts
```diff
@@ -247,7 +251,40 @@ export class WebShape<
     super(props, context);
     SvgTouchableMixin(this);
     this._remeasureMetricsOnActivation = remeasure.bind(this);
+    const componentDidMount = this.componentDidUpdate as () => void;
+    this.componentDidMount = () => {
+      componentDidMount();
```
EyMaddis, line 256: throws because you are not passing props.

Also: Why rename componentDidUpdate as componentDidMount?

Freddy03h:
Hi! Is there any progress on this?
For now, `react-native-svg` is not usable on an RNW project with accessibility considerations.

I'm also using react-native-svg on a RN project and wasn't aware that elements can be touchable.

I'm surely miss a point. What is preventing to use the library like this:
```jsx
<Svg>
  <Touchable>
    <Rect />
  </Touchable>
</Svg>
```
instead of having `onPress/*` directly on svg elements ?

zoontek:
IMHO, the cleanest way to do this should be dropping RNW 0.13 support and relying on [`usePressEvents`](https://github.com/necolas/react-native-web/blob/0.14.9/packages/react-native-web/src/modules/usePressEvents/index.js). Exactly like the `TouchableWithoutFeedback` [implementation](https://github.com/necolas/react-native-web/blob/0.14.9/packages/react-native-web/src/exports/TouchableWithoutFeedback/index.js).

I can easily make the change. Even if I also think that the best change should be to drop support for press events il the whole codebase (like @Freddy03h said), and rely to `Touchable*` and `Pressable` instead.

zoontek:
Huge 🆙

I rewrote the web implementation completely without the `TouchableMixin`. Instead, I apply `TouchableWithoutFeedback` conditionally. Accessibility is fixed, performances are superior (even if the way the style is resolved might be improved), codebase is a bit more up to date.

I also bumped some type definitions / some dependencies versions in order to fix the build (Bob build / type generation failed with the version set), added an `/example` project to help future contributors.

I also switched to `yarn`. Sorry about this one, but `npm` delete the `example` project when your run the `reinstall` script with it - for some reason). As I found a bit silly to have a `yarn.lock` in `example` and a `package-lock.json` at root. I can revert it.

lejsue:
I have this problem too, is it possible to merge these commits into main branch?
Thx

ElForastero:
I think it's time to merge this. The problem is really serious

exneval:
already test the implementation and so far so good ty @zoontek

WoLewicki:
Since https://github.com/react-native-svg/react-native-svg/pull/1585 has been merged, and I think this PR takes quite another approach to the problem, could you update this PR to only concern the `ReactNativeSVG.web.ts` file if it is possible? As for the other changes, the example app is already in the repo. As a side note, mixing multiple things in one PR makes it really hard to review and test easily 😅  Or maybe https://github.com/react-native-svg/react-native-svg/pull/1585 is enough for most use cases and we can close this? I believe the problem occurs when re-rendering the component with touchable props added. Am I right? cc @zoontek

zoontek:
I will rebase against master, remove the exemple project.

zoontek:
I added a working web example with RNW + kept my implementation (using `TouchableWithoutFeedback`) because the current implementation did not worked when I tried it.

There's still a small issue with ref forwarding, but it could be fixed in another PR.
I also deleted the `PanReponder` example (it just does not work, either on web or mobile)

Screenshot 2022-03-03 at 13 51 16

zoontek:
My recommandation: Cleanup this repository to the max, update dependencies, and schedule some breaking changes.
Delete all press handling props / methods (for the web **and** native).

It just add a lot of complexity for nothing. Developers can wrap `Svg` elements with `Pressable`, `View` + `PressResponder`, `Animated.View`… it's fine!

zoontek:
@WoLewicki I had several errors in the console. I can create a new PR to fix the example app (and maybe delete `TestsExample`? I think this directory is useless)

WoLewicki:
`TestsExample` will be used for providing tests examples for new PRs, which can be then tested to check if there are no regressions e.g. before a release. Example app is to be treated as a showcase of how to do things with the library, you can see the same structure in `react-native-screens`. PR with the changes you mentioned would be very welcome for sure :tada:

zoontek:
@WoLewicki The PR is here: https://github.com/react-native-svg/react-native-svg/pull/1717

Like I said here, the `onPress*` are currently broken on the web, the latest fix removed the focus on all SVG elements, but also broke pressability. And the `PanResponder` example doesn't work.

But IMHO it's a good start to implement breaking changes. I can help creating a roadmap for v13.

WoLewicki:
I added a review of #1717 and it seems to work just fine, thanks for that 😄  Are you sure #1585 broke the pressability? I commented the `if` checking for pressable options and on web the elements still do not react to being pressed, but I may be doing something wrong.

zoontek:
@WoLewicki Indeed, it might was super broken with focus on all elements **and** pressability broken too, I didn't checked 😄

WoLewicki:
Could you merge the current `main` to resolve conflicts?

zoontek:
@WoLewicki Done. However, I recommend using that as a temporary solution and remove press handling in next version.

It sures fixes the onPress related events, but:
- svg are not focusables
- some press responders errors are printed in the console when onPress* is set. It's working tho (but it should be removed, it's easier to wrap the whole Svg with a Pressable)
- PanResponder does not work when set directly on the Svg (but who does that? just wrap them with a View, they accept GestureHandlers)

WoLewicki:
Sorry for the late response. So do you think this PR provides better solution than the current implementation? About wrapping the whole Svg with a Pressable, do you think it won't add new problems, as the clickable area?

zoontek:
It's better than the current implementation for sure. But not perfect.

Wrapping SVG with Pressable / Animated.View / View with PanResponder will allow deletion of a lot of code / the suppression of a lot of issues 🙂

msand:
I think quite a few people depend on the press related functionality to only respond when interacting with filled / stroked parts of them canvas. I'm quite sure that wrapping with Pressable / PanResponder won't respect that part of the api. Or?

zoontek:
@msand IMHO, if a part of the SVG should be animated / move using PanResponder but not the rest, that's not an SVG, but two.

And even with that, components are implemented using basic UIView / View, it shouldn't be an issue.

msand:
Well, in practise, if the pressability would change to be based on bounding boxes rather than fill / stroke, a bunch of use cases would no longer be possible or would be buggy. The ability to have events based on this comes from the svg spec, so I'd say it is in scope of what should be supported by react-native-svg for single svg components.

cloudratha:
This is a blocker for our A11y audits. I've moved our implementation to module alias `react-native-svg` to `react-native-svg-web` in the meantime.

PR #1304, opened 2020-03-04
bugfix: Memory leak in SVG with clip path

meros:
* Do not retain copy of CGPath

# Summary

This PR fixes a memory leak related to clip paths

When rendering SVG's containing clip paths we noticed a memory leak when cycling the app between foreground/background. Looking at the history related allocation it was recently changed from simply a CGRetain of _clipNode to a copying and transform of the same. The CGPathRetain call is still there which seems wrong as a copy will introduce the reference needed.

## Test Plan

Run code with SVG that triggers the affected code path and make sure there are no memory leaks. Make sure the Retain call did not have other obscure reasons for being left in.

## Compatibility

| OS      | Implemented |
| ------- | :---------: |
| iOS     |    ✅     |
| Android |    ❌     |

## Checklist


msand:
Ah, great catch, thanks for this!

PR #1315, opened 2020-03-18
IOS ForeignObject fix

Zein-Fakhreddine:
## Summary
Fix ForeignObject not being linked correctly to the IOS project data file. This allows you to use the ForeignObject component on IOS. 

## Test Plan
Import and render a ForeignObject component if it doesn't crash then it works.

## Compatibility

| OS      | Implemented |
| ------- | :---------: |
| iOS     |    ✅     |
| Android |    ✅     |

## Checklist

### DIFF
--- a/ios/RNSVG.xcodeproj/project.pbxproj
+++ b/ios/RNSVG.xcodeproj/project.pbxproj
@@ -1,4 +1,4 @@
-// !$*UTF8*$!
+// !$*UTF9*$!
 {
 	archiveVersion = 1;
 	classes = {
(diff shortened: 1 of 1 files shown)


ios/RNSVG.xcodeproj/project.pbxproj
```diff
@@ -1,4 +1,4 @@
-// !$*UTF8*$!
+// !$*UTF9*$!
```
msand, line 1: Why this change?

Zein-Fakhreddine: @msand my mistake it seems to have automatically changed that. I got rid of it.

Issue #1418, opened 2020-07-13
Error while updating property 'd' of a view managed by: RNSVGPath


nitzanbueno:
I have an `SvgUri` component in my code.
The component causes a crash when the URI points to the following SVG file:
``` svg
<?xml version="1.0" encoding="UTF-8" standalone="no"?>
<!DOCTYPE svg PUBLIC "-//W3C//DTD SVG 1.1//EN" "http://www.w3.org/Graphics/SVG/1.1/DTD/svg11.dtd">
<svg version="1.1" xmlns="http://www.w3.org/2000/svg" width="500" height="500" viewBox="-0.9 -0.9 1.8 1.8">
	<rect fill="#FFFFFF" x="-0.9" y="-0.9" width="1.8" height="1.8"/>
	<g style="stroke-width:0.1;stroke-linejoin:round;opacity:1">
		<polygon fill="#000000" stroke="#000000" points="-4.9165444344952E-17,-0.71734170954349 0.70405037145575,-0.41272706360467 6.3108540577985E-17,-0.021725090572532 -0.70405037145575,-0.41272706360467"/>
		<polygon fill="#000000" stroke="#000000" points="6.3108540577985E-17,-0.021725090572532 0.70405037145575,-0.41272706360467 0.62948028357061,0.36901272915735 5.5589468959362E-17,0.81107056444244"/>
		<polygon fill="#000000" stroke="#000000" points="-0.70405037145575,-0.41272706360467 6.3108540577985E-17,-0.021725090572532 5.5589468959362E-17,0.81107056444244 -0.62948028357061,0.36901272915735"/>
	</g>
	<g style="opacity:1;stroke-opacity:0.5;stroke-width:0;stroke-linejoin:round">
		<polygon fill="#FEFE00" stroke="#000000" points="-4.9439549272153E-17,-0.74757064564692 0.19598754651203,-0.66277461469571 -1.6979580126642E-17,-0.57123720961754 -0.19598754651203,-0.66277461469571"/>
		<polygon fill="#FEFE00" stroke="#000000" points="0.23200530924361,-0.64654708450724 0.44357407294753,-0.55500967942906 0.24823152717746,-0.45589370157761 0.03601776273158,-0.55500967942906"/>
[...]

PR #1419, opened 2020-07-14
Fix issue #1418

nitzanbueno:
Fix bug in extractPolyPoints where an uppercase exponent (E) in a polygon (e.g. 4.7E-17) wouldn't be considered an exponent, only a lowercase one (4.7e-17)


WoLewicki (approved):
I tried the code you provided here: https://github.com/react-native-svg/react-native-svg/issues/1418#issuecomment-657768297 and with this PR, all seems to work fine :tada:. Thanks for your contribution.

Issue #1279, opened 2020-02-09
 'SvgXml' for web


hyxos:
I'm building an app with expo, and I get this error when using expo and trying to compile for web: 

Attempted import error: 'SvgXml' is not exported from 'react-native-svg'.

I've noticed this issue has kind of come up before https://github.com/react-native-community/react-native-svg/issues/1236

But why isn't there SvgXml support for web?  

Shouldn't we be able to apply this to web? It would be nice to be able to dynamically generate svgs from strings and for it to work across all platforms. SVGs are already first class citizens in the browser, I feel like there should be some type of mechanism to just pass the markup to the DOM.

I'm not sure I understand the technical reasons for not including it or if it just hasn't been implemented yet. If it is a matter of implementation, I would be willing to put in a PR but I need some direction and what the scope of that would be.

msand:
It's mainly due to the different processing modes available in the browser, and the different security tradeoffs it has. https://www.w3.org/TR/SVG2/conform.html#examples

People are likely not to understand the full implications of it, and use it when inappropriate. Thus I've left it up to consumers of the library to implement that if they need it, and then according to the security / risk model of the product it's applied to. Besides, svg / xml content from the web can be shown using e.g. <img /> tags, and making a component do that is very easy for the ones where that is suitable. If you just want to dump a bunch of xml inside the dom, just use https://reactjs.org/docs/dom-elements.html#dangerouslysetinnerhtml

nyousaf47:
any solution?

msand:
Many solutions with different tradeoffs, thus no prepared solution.

msand:
You probably want to read these two threads:

https://github.com/react-native-community/react-native-svg/issues/1235

https://github.com/expo/expo/issues/6660

It depends very much on your web setup, what bundlers you use, other plugins, security considerations etc. There are too many. possible ways to want it to function to satisfy all possibilities of the configuration space. A maximally restrictive (security wise) might make sense, i.e. not to parse the xml of content loaded from uri but rather let the browser handle it using img or css backgrounds https://svgwg.org/specs/integration/#secure-static-mode

<img src='data:image/svg+xml;utf8,<svg ... > ... </svg>'>

.bg {
  background: url('data:image/svg+xml;utf8,<svg ...> ... </svg>');
}

Or something like this for ie compat: "Optimized URL-encoded" https://css-tricks.com/probably-dont-base64-svg/

[...]

Issue #1236, opened 2020-01-02
Attempted import error: 'SvgXml' is not exported from 'react-native-svg'.


kopax:
# Bug

I am not able to use `SvgXml`.

Since #1235 , I have a different import between web and native, I expected `SvgXml` to help me use svg on the web, using `babel-plugin-inline-svg`

## Environment info

- react-native-svg 9.13.3
- react-native-svg-transformer 0.14.3
- expo SDK: 36 Documentation: https://docs.expo.io/versions/latest/sdk/svg/

## Steps To Reproduce

```jsx
import { SvgXml } from 'react-native-svg';
import { Platform } from 'react-native-web';
import Logof from '../Logo';

const Logo = Platform.OS === 'web' ? <SvgXml xml={Logof} /> : <Logof />;
```

It should import the component

Instead, I have:

    Attempted import error: 'SvgXml' is not exported from 'react-native-svg'.

kopax:
Apparently SvgXml does not exist for web and I was using wrong `import { Platform } from 'react-native-web'` instead of `import { Platform } from 'react-native'`

msand:
I would recommend defining a separate Logo.js and either a Logo.web.js or Logo.native.js (or even Logo.ios.js and Logo.android.js) instead of using Platform. This way it only bundles the platform specific code (faster TTI), and essentially does the conditional logic at build time instead of repeatedly at run time (better performance).

kopax:
Thanks for the recommendation, I am new in react-native ecosystem so I wouldn't have known. I will follow it

PR #1431, opened 2020-08-04
SvgXml implementation for web

MatheusrdSantos:
# Summary

#1279
#1236

This PR contains the implementation of SvgXml for web. This web implementation has exactly the same usage as the native implementation. 
The web implementation relies on [unstable_createElement](https://github.com/necolas/react-native-web/blob/master/packages/react-native-web/src/exports/createElement/index.js)  and [View](https://github.com/necolas/react-native-web/blob/master/packages/react-native-web/src/exports/View/index.js) from [react-native-web](https://github.com/necolas/react-native-web). So, basically I used `unstable_createElement` to create an `<svg>` element with the props that comes from the XML string and set the `innerHTML` from the `<svg>` to be the XML content.
In order to accept react-native props (the same as SvgXml native implementation), I used the `<View>` to wrap the `<svg>` and simulate the same behavior.
None. It just brings a new supported component for web.
## Test Plan

You can check a working demo here: https://codesandbox.io/s/keen-platform-z40rg

## Compatibility

| OS      | Implemented |
| ------- | :---------: |
| iOS     |    ✅     |
| Android |    ✅     |
| Web |    ✅     |

## Checklist

- [x] I have tested this on browser

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -2924,8 +2924,7 @@
     "array-find-index": {
       "version": "1.0.2",
       "resolved": "https://registry.npmjs.org/array-find-index/-/array-find-index-1.0.2.tgz",
-      "integrity": "sha1-3wEKoSh+Fku9pvlyOwqWoexBh6E=",
-      "dev": true
+      "integrity": "sha1-3wEKoSh+Fku9pvlyOwqWoexBh6E="
     },
     "array-ify": {
       "version": "1.0.0",
@@ -2989,8 +2988,7 @@
     "asap": {
       "version": "2.0.6",
       "resolved": "https://registry.npmjs.org/asap/-/asap-2.0.6.tgz",
-      "integrity": "sha1-5QNHYR1+aQlDIIu9r+vLwvuGbUY=",
-      "dev": true
+      "integrity": "sha1-5QNHYR1+aQlDIIu9r+vLwvuGbUY="
     },
     "asn1": {
       "version": "0.2.4",
@@ -5182,8 +5180,7 @@
     "core-js": {
       "version": "2.6.11",
       "resolved": "https://registry.npmjs.org/core-js/-/core-js-2.6.11.tgz",
-      "integrity": "sha512-5wjnpaT/3dV+XB4borEsnAYQchn00XSgTAWKDkEqv+K8KevjbzmofK6hfJ9TZIlpj2N0xQpazy7PiRQiWHqzWg==",
-      "dev": true
+      "integrity": "sha512-5wjnpaT/3dV+XB4borEsnAYQchn00XSgTAWKDkEqv+K8KevjbzmofK6hfJ9TZIlpj2N0xQpazy7PiRQiWHqzWg=="
     },
     "core-js-compat": {
       "version": "3.6.4",
@@ -5226,7 +5223,6 @@
       "version": "15.6.3",
       "resolved": "https://registry.npmjs.org/create-react-class/-/create-react-class-15.6.3.tgz",
       "integrity": "sha512-M+/3Q6E6DLO6Yx3OwrWjwHBnvfXXYA7W+dFjt/ZDBemHO1DDZhsalX/NUtnTYclN6GfnBDRh4qRHjcDHmlJBJg==",
-      "dev": true,
       "requires": {
         "fbjs": "^0.8.9",
         "loose-envify": "^1.3.1",
@@ -5236,14 +5232,12 @@
         "core-js": {
           "version": "1.2.7",
           "resolved": "https://registry.npmjs.org/core-js/-/core-js-1.2.7.tgz",
-          "integrity": "sha1-ZSKUwUZR2yj6k70tX/KYOk8IxjY=",
-          "dev": true
+          "integrity": "sha1-ZSKUwUZR2yj6k70tX/KYOk8IxjY="
         },
         "fbjs": {
           "version": "0.8.17",
           "resolved": "https://registry.npmjs.org/fbjs/-/fbjs-0.8.17.tgz",
           "integrity": "sha1-xNWY6taUkRJlPWWIsBpc3Nn5D90=",
-          "dev": true,
           "requires": {
             "core-js": "^1.0.0",
             "isomorphic-fetch": "^2.1.1",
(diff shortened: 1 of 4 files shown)


kofsiwon:
I hope release this quickly. I need SvgXml for react native web!!

Sharcoux:
For those still needing this, you can try our package: [@cantoo/rn-svg](https://www.npmjs.com/package/@cantoo/rn-svg)

MattFoley:
@WoLewicki @MatheusrdSantos @Sharcoux Should we merge this in? It seems risky long term to have a fork maintain this functionality. Thanks!

Sharcoux:
@MattFoley I'm glad that the project revived.

We can rebase our fork and make it ready for merge, that would be awesome. However, there is a concern you should know about:

We had to wrap the svg inside a View. This has impacts when trying to style the SVG, especially with everything related to layout. We made a system that distribute the styles between the wrapper and the svg itself, but it will very likely introduce problems in some layouts. `unstable_createElement` would not let us pass down every react-native props to the underlying element. Another valid approach could be to only export the element created with `unstable_createElement` and give up on unsupported props.

You'll find all the details here: https://github.com/necolas/react-native-web/pull/1686

MattFoley:
Hmm, I'll be honest I don't fully understand the debate on RNW, and would need more time to dive into the specifics. I also am using v14.1 of rn-svg, and don't think I can downgrade to 13. I started to rebase your fork up to v14.1 myself, but haven't finished yet, working through some BoB type errors.

Sharcoux:
Ouch, I missed your comment, sorry.

No, don't rebase this here. If you want to follow our approach, we need to push our current version of it. Many fixes have been made since this.

Sharcoux:
Moreover, @MattFoley there should not be any rebase needed because we only need to add new files. Not edit existing ones.

MattFoley:
@Sharcoux I would love if you had a new version you could push up on 14.x!

bohdanprog:
Hello, 
@MatheusrdSantos, could you please explain to me what the difference is between that solution and our current implementation?

Thank you.

MatheusrdSantos:
Hi, @bohdanprog

this PR is just adding the web implementation for the SvgXml component, which is currently missing from the main package.
It uses react-native-web to keep the react-native syntax, just as the current web implementation for Svg, G, Path... does.

The main difference is the need to import some hooks that are not exposed by the react-native-web api. Those hooks are important to emulate the native behavior, and therefore, make it possible to watch layout changes, measure the element and listen for touches by using the same native api.

This PR (https://github.com/software-mansion/react-native-svg/pull/1686) has the same implementation I did, but it's not rebased. Anyway it's a good complement to my explanation.

bohdanprog:
Hm, @MatheusrdSantos, could you explain me please:
What kind of supported component does it bring to the web?
Because for me it looks almost the same or I missed something?

Thank you.

bohdanprog:
Hi @MatheusrdSantos,
I’ve improved some parts of your code and hope you don’t mind.
Could you take a look at my changes and let me know what you think?
Thank you!

MatheusrdSantos:
@bohdanprog The changes are good, thanks for the code improvements.

There's just one change I didn't understand. Is there's a reason to move the `react-native-web` package from `peerDependencies/optionalDependencies` to `dependencies`?

I think we could keep this dependency as optional because not all user may want to use the web implementation.

bohdanprog:
Hi @MatheusrdSantos,

We have made some updates to react-native-svg, and as far as I know, we can now display SvgXml, SvgCss, etc.
Would you like to verify this?
Thank you.

Sharcoux:
I tried. SvgXml is undefined if I import `import { SvgXml } from 'react-native-web'`. When I check the code, I see that there is an implementation in the file `src/xml.tsx`. However, I don't understand how that implementation is supposed to be made available for the web, as it is never exported from `ReactNativeSVG.web.ts`

I think that you want to copy the lines 29 to 47 from `ReactNativeSVG.ts` into `ReactNativeSVG.web.ts`:

```
import {
  parse,
  SvgAst,
  SvgFromUri,
  SvgFromXml,
  SvgUri,
  SvgXml,
  camelCase,
  fetchText,
  JsxAST,
  Middleware,
  Styles,
  UriProps,
  UriState,
  XmlAST,
  XmlProps,
  XmlState,
  AstProps,
} from './xml';
```

But also the lines 141 to 146 and 178 to 188:

image


```js
export type {
  JsxAST,
  Middleware,
  Styles,
  UriProps,
  UriState,
  XmlAST,
  XmlProps,
  XmlState,
  AstProps,
};
```

bohdanprog:
Hi @Sharcoux,
Can you please try to run the test app from the main branch?
Those functionalities are already available if you use the library from the main branch.
I believe we are going to create the new release of r`eact-native-svg` today.

bohdanprog:
@Sharcoux FYI, now it should work when you are using `react-native-svg 15.5.0`

jakex7:
Closing as it has been implemented in `15.5.0` https://github.com/software-mansion/react-native-svg/releases/tag/v15.5.0

MatheusrdSantos:
Hi, @bohdanprog
I tested the version 15.5.0. There are two issues.

The first error is happening because the dependency `@react-native/assets-registry` is missing in the package.json file from `react-native-svg`. It can be easily fixed by manually installing the dependency.
```
Module not found: Can't resolve '@react-native/assets-registry/registry' in '/home/matheus/cantoo/cantoo-monorepo/node_modules/react-native-svg/lib/module/lib'
node_modules/react-native-svg/lib/module/lib/resolveAssetUri.js
```


The second issue is that `@react-native/assets-registry` uses flow to check its types and doesn't expose a built version for general js usage.
There are two recent issue trying to fix it:
**fix: Remove Flow from @react-native/assets-registry - registry.js:** https://github.com/facebook/react-native/pull/44002
**Build assets-registry for general consumption:** https://github.com/facebook/react-native/pull/39440
```
Module parse failed: Unexpected token (13:7)
node_modules/@react-native/assets-registry/registry.js
You may need an appropriate loader to handle this file type, currently no loaders are configured to process this file. See https://webpack.js.org/concepts#loaders
| 'use strict';
| 
> export type PackagerAsset = {
|   +__packager_asset: boolean,
|   +fileSystemLocation: string,
```

bohdanprog:
@MatheusrdSantos Can you tell me please, how can I reproduce that issue?

bohdanprog:
Hi @MatheusrdSantos, IDK why but your simple repro isn't simple :D
I created a repository with a simple [example](https://github.com/bohdanprog/RNSVG-XML-web), can you modify that?

MatheusrdSantos:
@bohdanprog I created this pull request: https://github.com/bohdanprog/RNSVG-XML-web/pull/1.
I noticed that `@react-native/assets-registry/registry` was missing because `react-native-svg@15.5.0` requires the `react-native` version to be `^0.72.3`. I was using an older version.

But the second issue remains.

bohdanprog:
@MatheusrdSantos Hm as I see that issue exists when we use the webpack configuration, but IDK why.
Maybe is it a matter of configuration?

MatheusrdSantos:
@bohdanprog yes, it's the same issue.

The webpack issue is a matter of configuration that will impact all `react-native-svg` users that don't have a proper config to parse flow files.

bohdanprog:
@MatheusrdSantos I added the correct webpack config to the example app, can you check that?

PR #1435, opened 2020-08-12
Fix scale 0 is ignored

soh335:
# Summary

```scale={0}```  is not worked currently. Because scale `0` is fallback to default value (1).

## Test Plan

```
<Rect x={0} y={0} width={100} height={100} scale={0} fill="black" />
```

スクリーンショット 2020-08-12 12 09 05

---

スクリーンショット 2020-08-12 12 09 24

## Compatibility

| OS      | Implemented |
| ------- | :---------: |
| iOS     |    ✅     |
| Android |    ✅     |

## Checklist

### DIFF
--- a/src/lib/extract/extractTransform.ts
+++ b/src/lib/extract/extractTransform.ts
@@ -31,7 +31,7 @@ function universal2axis(
   universal: NumberProp | NumberProp[] | undefined,
   axisX: NumberProp | void,
   axisY: NumberProp | void,
-  defaultValue?: number,
+  defaultValue: number,
 ): [number, number] {
   let x;
   let y;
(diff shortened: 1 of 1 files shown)

### CODE REVIEW

**Comment by bohdanprog** [comment]
Hello @soh335,
Here is how it behaves in [web](https://www.svgviewer.dev/s/sZWOKEml).
I guess we expect the same behavior.

**Comment by bohdanprog** [comment]
@soh335
We should use that scale prop like here:
```
 <Svg>
      <Rect
        x={0}
        y={0}
        width={100}
        height={100}
        fill="black"
        transform={'scale(0)'}
      />
    </Svg>
```

jakex7:
> [!WARNING]
> `scale` prop, like other explicit transformation props (`translate`, `rotate` etc.) are deprecated API and should not be used. We'll remove that in the future.
Use `transform` prop instead with CSS like API. Example: `transform="rotate(45), translate(100, -50)"`

Issue #1409, opened 2020-06-29
ClipPath is not working properly on Android without clipRule prop


ObidosDev:
# Bug

<img width="447" alt="Screenshot 2020-06-29 at 18 34 54" src="https://user-images.githubusercontent.com/16224235/86025759-4281ec80-ba37-11ea-9bc5-bc5b28daf983.png">

Looks like ClipPath has only `id` prop, but at the same time code below do not working properly on android without this prop

## Environment info

```bash
 System:
    OS: macOS 10.15.5
    CPU: (8) x64 Intel(R) Core(TM) i7-4980HQ CPU @ 2.80GHz
    Memory: 19.82 MB / 16.00 GB
    Shell: 3.2.57 - /bin/bash
  Binaries:
    Node: 12.2.0 - ~/.nvm/versions/node/v12.2.0/bin/node
    Yarn: 1.22.4 - /usr/local/bin/yarn
    npm: 6.14.5 - ~/.nvm/versions/node/v12.2.0/bin/npm
    Watchman: 4.9.0 - /usr/local/bin/watchman
  Managers:
    CocoaPods: 1.9.1 - /usr/local/bin/pod
  SDKs:
    iOS SDK:
      Platforms: iOS 13.5, DriverKit 19.0, macOS 10.15, tvOS 13.4, watchOS 6.2
    Android SDK: Not Found
  IDEs:
    Android Studio: 3.6 AI-192.7142.36.36.6392135
    Xcode: 11.5/11E608c - /usr/bin/xcodebuild
  Languages:
    Java: 9.0.4 - /usr/bin/javac
    Python: 2.7.16 - /usr/bin/python
  npmPackages:
    @react-native-community/cli: Not Found
    react: 16.11.0 => 16.11.0 
    react-native: 0.62.2 => 0.62.2 
  npmGlobalPackages:
    *react-native*: Not Found
```

Library version: 12.1.0

## Steps To Reproduce

Issues without reproduction steps or code are likely to stall.

1. Use code below
2. Try it with `// @ts-ignore clipRule={'evenodd'}` and without

Describe what you expected to happen:

1. It should work without it OR need to fix the TS (types)

## Short, Self Contained, Correct (Compilable), Example

```
<Svg width={'100%'} height={'100%'} viewBox={'0 0 256 340'}>
              <Defs>
                <Rect
                  id={'backgroundRect'}
                  x={-1}
                  y={-1}
                  width={258}
                  height={342}
[...]

PR #1465, opened 2020-09-30
refactor: add the clip rule property to the clip path types

Gabrielfcs:
# Summary

I identified that one of the most important properties for the ClipPath component and that it is working perfectly, did not have its type declared in the type file. What causes warnings in the IDE's, as if declaring a non-existent type for that component, so I found the #1409 problem and corrected it by adding this missing property.

## Test Plan

I don't know what to put here, but I just added the missing type, as in the code below:
```
export type FillRule = 'evenodd' | 'nonzero';

...

export interface ClipPathProps {
  id?: string;
  clipRule?: FillRule; <- line added
}
```

Any IDE and running code

Create code that depends on the clipRule attribute in a ClipPath component and see if the property works with or without the type

## Compatibility

| OS      | Implemented |
| ------- | :---------: |
| iOS     |   ✅      |
| Android |      ✅.      |

## Checklist

- [X] I updated the typed files (typescript)

### DIFF
--- a/src/index.d.ts
+++ b/src/index.d.ts
@@ -241,6 +241,7 @@ export type Circle = React.ComponentClass<CircleProps>;
 
 export interface ClipPathProps {
   id?: string;
+  clipRule?: FillRule;
 }
 export const ClipPath: React.ComponentClass<ClipPathProps>;
 export type ClipPath = React.ComponentClass<ClipPathProps>;


magicismight:
https://developer.mozilla.org/en-US/docs/Web/SVG/Attribute/clip-rule

> The clip-rule attribute only applies to graphics elements that are contained within a <clipPath> element. The clip-rule attribute basically works as the fill-rule attribute, except that it applies to <clipPath> definitions.

`clipRule` is not a standard prop for non-graphics elements like `ClipPath`
We will fix the `clipRule` prop bug in Android.

PR #1476, opened 2020-10-16
fix(parse): fix issue on trying to parse invalid XML

### LINKED ISSUE #1105: SvgUri On Error
# Feature Request

 We need to have an on Error when we use svgUri as the app crashes if there is an error in the SVG format. A best practice guide on the format of the svg from uri is also required as it is based on trial and error and standard svg's do not load and app crashes

## Why it is needed

  In case the svg's are coming from a third party source and if the svg is not confirming to the specs required by react-native-svg, the app crashes. This can be handled and a default svg can be shown.

## Possible implementation

  It really helps if you could describe from a technical POV how this new feature would work, which code it rely on, etc

### Code sample

  Please show how the new code could work, if doable

uqmessias:
# Summary
It finally fixes #1105, the parser throws an error when trying to parse an invalid `SVG` file, but instead of throwing an error, it should call `onError` prop (if it was set) and let it handle the error.

## Test Plan

It was tested on this repo: https://github.com/vinitgundeti/reactnativesvg

rn-svg-fixes-onerror-handler

Thanks @vinitgundeti, for taking the time to uploading it.

## Compatibility

| OS      | Implemented |
| ------- | :---------: |
| iOS     |    ✅    |
| Android |    ✅     |

### DIFF
--- a/src/xml.tsx
+++ b/src/xml.tsx
@@ -109,9 +109,25 @@ export function SvgAst({ ast, override }: AstProps) {
 export const err = console.error.bind(console);
 
 export function SvgXml(props: XmlProps) {
-  const { onError = err, xml, override } = props;
-  const ast = useMemo<JsxAST | null>(() => (xml !== null ? parse(xml) : null), [
+  const { xml, override } = props;
+  const onError = typeof props.onError == 'function'
+    ? props.onError
+    : (
+      typeof (override as AdditionalProps).onError == 'function'
+        ? (override as AdditionalProps).onError
+        : err
+    )
+  const ast = useMemo<JsxAST | null>(() => {
+    try {
+      return xml !== null ? parse(xml) : null;
+    } catch (e) {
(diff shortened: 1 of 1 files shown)


vinitgundeti:
@uqmessias Thank you very much for fixing this, this was must required feature, i have tested it and it is working fine. when will this be in the main branch ?

halilb:
Thanks for the fix @uqmessias. I wasn't able to add your branch as dependency because of typescript exceptions. I fixed them and created https://github.com/uqmessias/react-native-svg/pull/1 into your branch.

uqmessias:
Thanks @halilb, I merged your PR into mine and made some changes. Could you test it again, please?

halilb:
> Thanks @halilb, I merged your PR into mine and made some changes. Could you test it again, please?

That's working fine for me. Thank you @uqmessias!

PR #1503, opened 2020-12-10
Fix for SvgUri and SvgCssUri crashing on non-existing svg files

### LINKED ISSUE #1500: SvgUri and SvgCssUri crash on URI not existing
When SvgUri or SvgCssUri load an URI that doesn't exist, it tries to parse the HTML error-page as SVG and crashes. The fetchText method in `node_modules/react-native-svg/src/xml.tsx` doesn't check for this condition.

SjaakSchilperoort:
# Summary

Fix for interpreting an HTML error response as svg data, and then crash when this cannot be parsed as well formed XML

closes #1500 

## Test Plan

n/a

Just a a clean project with the latest versions of react-native and react-native-svg

Include this in App.js

```
<SvgUri width={100} height={100} uri={'https://dev.w3.org/SVG/tools/svgweb/samples/svg-files/ruby.svg'} />
```
and you'll get:

Then make the file name invalid
```
<SvgUri width={100} height={100} uri={'https://dev.w3.org/SVG/tools/svgweb/samples/svg-files/xruby.svg'} />
```
and you'll get:

With the fix you'll get this instead:

## Compatibility

| OS      | Implemented |
| ------- | :---------: |
| iOS     |    ✅     |
| Android |    ✅     |

## Checklist

### DIFF
--- a/src/xml.tsx
+++ b/src/xml.tsx
@@ -124,7 +124,7 @@ export function SvgXml(props: XmlProps) {
 
 export async function fetchText(uri: string) {
   const response = await fetch(uri);
-  return await response.text();
+  return response.ok ? await response.text() : null;
 }
 
 export function SvgUri(props: UriProps) {
--- a/src/xml.tsx
+++ b/src/xml.tsx
@@ -127,8 +127,7 @@ export async function fetchText(uri: string) {
   if (response.ok) {
     return await response.text();
   }
-  console.warn(`Fetch ${uri} failed with status ${response.status}`);
-  return null;
+  throw new Error(`Fetching ${uri} failed with status ${response.status}`);
 }
 
 export function SvgUri(props: UriProps) {
--- a/src/xml.tsx
+++ b/src/xml.tsx
@@ -124,7 +124,11 @@ export function SvgXml(props: XmlProps) {
 
 export async function fetchText(uri: string) {
   const response = await fetch(uri);
-  return response.ok ? await response.text() : null;
+  if (response.ok) {
+    return await response.text();
+  }
+  console.warn(`Fetch ${uri} failed with status ${response.status}`);
+  return null;
 }
 
 export function SvgUri(props: UriProps) {


jorgemasta:
Good catch! I think is a good improvement

Ivanacc:
It would be great to add this fix! :)

zrina1314:
why hasn't this library been updated for a long time

src/xml.tsx
```diff
@@ -124,7 +124,7 @@ export function SvgXml(props: XmlProps) {
 
 export async function fetchText(uri: string) {
   const response = await fetch(uri);
-  return await response.text();
+  return response.ok ? await response.text() : null;
```
WoLewicki, line 127: I think it would be good to add some message e.g. in a form of `console.warn` if the `response.ok` is `false`, so we could get information what went wrong. Then I think it is ready to be merged :tada:

SjaakSchilperoort: Thanks, makes sense.

WoLewicki:
I double-checked the code I think it would be even better to `throw Error` instead of `console.warn` there, since this method is used with `catch` in `SvgUri`, so the `Error` is caught and can be managed by the user in `onError` prop of `SvgUri`: https://github.com/react-native-svg/react-native-svg/blob/650275fdf5f0e3ff3fc51e8b2d8b98fd34110d1c/src/xml.tsx#L137. So such an example would work: 

```tsx
import React, {Component} from 'react';
import {  SvgUri } from 'react-native-svg';

class Example extends Component {

    state={
      uri: 'https://dev.w3.org/SVG/tools/svgweb/samples/svg-files/xruby.svg',
    }

  render() {
    return (
      <>
        <SvgUri width={100} onError={err => this.setState({uri: 'https://dev.w3.org/SVG/tools/svgweb/samples/svg-files/ruby.svg'})} height={100} uri={this.state.uri} />
      </>
    );
  }
}
```

What do you think of it?

WoLewicki:
I will also change the target branch since now `main` is the default branch to merge commits to.

SjaakSchilperoort:
Yes, looks good.

Note that component `SvgCssUri`, which I use in my code, doesn't support property `onError`. I guess that adding support for this is a different issue.

WoLewicki:
Yeah, I think it was missed in this commit: https://github.com/react-native-svg/react-native-svg/commit/3c32a6fdd580dea53ecc271288155117ce040b55. Could you make a PR with such change for `SvgCssUri`? I also checked where else `fetchText` is used and it looks like there is another place where it could be handled with `onError`: https://github.com/react-native-svg/react-native-svg/blob/54e40251a491bb1a2d0a75e4748a23ea24fb3f6b/src/LocalSvg.tsx#L58 Could you add same `.catch(err)` with an option of `onError` there too? It would then be handled everywhere then :tada: As for this PR, I think it is ready now so I'll merge it :tada: Thanks for your contribution.

SjaakSchilperoort:
> Yeah, I think it was missed in this commit: https://github.com/react-native-svg/react-native-svg/commit/3c32a6fdd580dea53ecc271288155117ce040b55. Could you make a PR with such change for SvgCssUri?

See PR #1718

CuriosBasant:
The `onError` prop on `SvgCssUri` doesn't work, it also shows ts error when using it.

PR #1542, opened 2021-02-22
fix: move bitmap recycling to onDetachedFromWindow

WoLewicki:
# Summary

This PR fixes the issues in `react-native-screens` and probably other libraries using native `Fragments` for transitions on Android (see https://github.com/software-mansion/react-native-screens/issues/773). It moves the recycling and setting `mBitMap` to `null` to `onDetachedFromWindow` when the view is removed from `react` view hierarchy. `invalidate` is not always the proper place for doing it since the native view can still be visible for the user after the `invalidate` is called (see this comment with video: https://github.com/software-mansion/react-native-screens/issues/773#issuecomment-769418173).

## Test Plan

You can use the code example attached in the above comment.

The project with `svg` view and `react-native-screens` and `native-stack` navigator.

Go back from the screen with an `svg` view.

## Compatibility

| OS      | Implemented |
| ------- | :---------: |
| Android |    ✅.      |

## Checklist

- [x] I have tested this on a simulator

### DIFF
--- a/android/src/main/java/com/horcrux/svg/RenderableViewManager.java
+++ b/android/src/main/java/com/horcrux/svg/RenderableViewManager.java
@@ -1284,6 +1284,8 @@ class RenderableViewManager extends ViewGroupManager<VirtualView> {
             @Override
             public void onChildViewRemoved(View view, View view1) {
                 if (view instanceof VirtualView) {
+                    SvgView svgView = ((VirtualView) view).getSvgView();
+                    svgView.setRemovedFromReactViewHierarchy();
                     invalidateSvgView((VirtualView) view);
                 }
             }
--- a/android/src/main/java/com/horcrux/svg/SvgView.java
+++ b/android/src/main/java/com/horcrux/svg/SvgView.java
@@ -88,8 +88,12 @@ public class SvgView extends ReactViewGroup implements ReactCompoundView, ReactC
             }
             mRendered = false;
             ((VirtualView) parent).getSvgView().invalidate();
-            return;
         }
+    }
+
+    @Override
+    protected void onDetachedFromWindow() {
+        super.onDetachedFromWindow();
         if (mBitmap != null) {
             mBitmap.recycle();
         }
--- a/android/src/main/java/com/horcrux/svg/SvgView.java
+++ b/android/src/main/java/com/horcrux/svg/SvgView.java
@@ -66,6 +66,7 @@ public class SvgView extends ReactViewGroup implements ReactCompoundView, ReactC
     }
 
     private @Nullable Bitmap mBitmap;
+    private boolean mRemovedFromReactViewHierarchy;
 
     public SvgView(ReactContext reactContext) {
         super(reactContext);
@@ -78,6 +79,10 @@ public class SvgView extends ReactViewGroup implements ReactCompoundView, ReactC
         SvgViewManager.setSvgView(id, this);
     }
 
+    public void setRemovedFromReactViewHierarchy() {
+        mRemovedFromReactViewHierarchy = true;
+    }
+
     @Override
     public void invalidate() {
         super.invalidate();
@@ -88,6 +93,16 @@ public class SvgView extends ReactViewGroup implements ReactCompoundView, ReactC
             }
             mRendered = false;
             ((VirtualView) parent).getSvgView().invalidate();
+            return;
+        }
+        if (!mRemovedFromReactViewHierarchy) {
+            // when view is removed from the view hierarchy, we want to recycle the mBitmap when
+            // the view is detached from window, in order to preserve it for during animation, see
+            // https://github.com/react-native-svg/react-native-svg/pull/1542
+            if (mBitmap != null) {
+                mBitmap.recycle();
+            }
+            mBitmap = null;
         }
     }
 


gustavopch:
Hey, maintainers, please don't forget this PR. It prevents SVG icons from disappearing before the screen is fully hidden. I applied it with `patch-package` in my app and it seems to work fine. Pretty simple change, probably 2min to review. :blush:

WoLewicki:
This PR may not be enough in some cases, e.g. when using tab navigator, so I would treat it as a workaround for some cases right now. I'll try to update it with a more generic solution.

alexco2:
I tested it with stack navigator and react-native-screens. SVG elements that I animate with reanimated are not appearing/ animating anymore after applying this PR. I have not testet it yet if this fixes the standard problem with static SVG's.

WoLewicki:
@alexco2 can you provide a repo with reproduction of the problem when this PR applied?

hadnet:
Hey guys, I'm using the react-native-svg 12.1.1 and my svgs disappears during animation transitions using react-native-screens/native-stack on Android (this issue here https://github.com/software-mansion/react-native-screens/issues/773#issue-784154619). Has this problem already been resolved or not?

alexco2:
@WoLewicki sorry for being late. I don't have much time at the moment and don´t know when I will be able to give you an reproduction of the problem. I will try to do it asap.

WoLewicki:
@hadnet does the issue appear when this PR is applied in your code?

hadnet:
> @hadnet does the issue appear when this PR is applied in your code?

It doesnt. When apply this PR the problem is solved!

WoLewicki:
Ok so it is up to the maintainers of this library to merge it.

alexco2:
@WoLewicki
As promised, here the snippet with a representation of the SVG animation I was referring to.
If your PR is applied, the animation stops working.

https://snack.expo.dev/@expoco2/animatedsvg
```
import React, { useEffect } from 'react';
import { Text, View, StyleSheet } from 'react-native';
import Constants from 'expo-constants';
import Animated, {
  useSharedValue,
  useAnimatedStyle,
  withRepeat,
  withTiming,
} from 'react-native-reanimated';
import { Svg, Circle } from 'react-native-svg';

const AnimatedCircle = Animated.createAnimatedComponent(Circle);

  const Animation = useSharedValue(1);

  useEffect(() => {
    Animation.value = withRepeat(withTiming(0, { duration: 2000 }), -1, true);
  }, []);

  const animtedProps = useAnimatedStyle(() => {
    const r = Animation.value * 50;
    return { cx: '50', cy: '50', r, fill: 'green' };
  });

  return (
      <Svg width={'100%'} height={'100%'}>
        <AnimatedCircle animatedProps={animtedProps} />
        <Circle cx={200} cy={200} r={50} fill="yellow" />
      </Svg>
    </View>
  );
}

  container: {
    flex: 1,
    justifyContent: 'center',
    alignItems: 'center',
    paddingTop: Constants.statusBarHeight,
    backgroundColor: '#ecf0f1',
    padding: 8,
  },
});
```

WoLewicki:
@alexco2 it is probably due to the usage of `invalidate` in `VirtualView`: https://github.com/react-native-svg/react-native-svg/blob/38f52a5c1404f75f0e0ee390468b6416d1eeab23/android/src/main/java/com/horcrux/svg/VirtualView.java#L113, which is not used only when the component is removed in React tree, but when its props are changed too. When animating, probably `invalidate` is called many times and it should trigger recycling bitmap, which with my PR it doesn't do. We need to somehow get the information that the whole component has been removed from React hierarchy and then not trigger recycling of bitmap since it will be done in `onDetachedFromWindow`. I might be wrong though and it should be tested first.

alexco2:
@WoLewicki 
Unfortunatly my skills in native/ java are from from enough to be of real help here. I´m happy though to support you in testing potential fixes.

sommcz:
Any updates on this issue?

thomasviaud:
Yes, can you please provide updates please ?

WoLewicki:
I think the only solution now is to use these changes with [patch-package](https://www.npmjs.com/package/patch-package).

android/src/main/java/com/horcrux/svg/RenderableViewManager.java
```diff
@@ -1284,6 +1284,8 @@ public void onChildViewAdded(View view, View view1) {
             @Override
             public void onChildViewRemoved(View view, View view1) {
                 if (view instanceof VirtualView) {
+                    SvgView svgView = ((VirtualView) view).getSvgView();
+                    svgView.setRemovedFromReactViewHierarchy();
```
AlexanderEggers, line 1288: I tested your code and there is a chance that `svgView` is null. I would suggest doing a null check here - just to avoid potential crashes.

ovidiuwin:
Any updates with this issue?

mordechaim:
I applied this PR with patch-package. Now I'm facing issues with Victory charts not updating on props change.

WoLewicki:
Closing since #1844 has been merged. Feel free to comment there if there are any problems. @mordechaim could you post a reproduction of the problem in a form of an issue so we can work on it?

mordechaim:
@WoLewicki I created a reproduction: https://github.com/mordechaim/SvgVictoryTest

There are 2 branches
-  `svg@12+patch` uses the patch I posted up here and works flawlessly
- `svg@13` uses the latest version without the patch, the UI locks up after 3 button presses

PR #1561, opened 2021-04-01
Support PlatformColor

oblador:
# Summary

Since version 0.63 React Native has supported `PlatformColor` structs that among other features has dark mode support built in. Using those values with RNSVG today throws a warning `"[Object object]" is not a valid color or brush` and makes the path have a clear color. 

This PR adds support for `PlatformColor` by utilizing the color parsing code shipped with React Native itself. In order to support the dynamic properties I had to retain the reference as a `UIColor` and convert it to `CGColorRef` as it's read. This might have a performance penalty. 

The Android implementation doesn't yet have support for dynamic values (EDIT: react-native itself doesn't support it, so we're good there I think). Since it relies on the `ColorPropConverter` class, I think it means minimum supported version of RN will be raised to 0.63.

Fixes https://github.com/react-native-svg/react-native-svg/issues/1391

## Test Plan

I couldn't find an example project in the repo, so I used my own example project, see video:

## Compatibility

| OS      | Implemented |
| ------- | :---------: |
| iOS     |    ✅     |
| Android |    ✅     |

## Checklist


src/index.d.ts
```diff
@@ -1,6 +1,6 @@
 import * as React from 'react';
 import * as ReactNative from 'react-native';
-import { GestureResponderEvent } from 'react-native';
+import { GestureResponderEvent, OpaqueColorValue } from 'react-native';
```
Saadnajmi, line 3: Shouldn't this just be "ColorValue"?

src/index.d.ts
```diff
@@ -101,7 +101,7 @@ export type rgbaArray = ReadonlyArray<number>;
 // int32ARGBColor = 0xaarrggbb
 export type int32ARGBColor = number;
 
-export type Color = int32ARGBColor | rgbaArray | string;
+export type Color = int32ARGBColor | rgbaArray | OpaqueColorValue | string;
```
Saadnajmi, line 105: Could we just replace the "Color" type altogether with "ColorValue" from react native? It should represent the same set of types of colors

src/lib/extract/extractBrush.ts
```diff
@@ -40,6 +40,16 @@ export default function extractBrush(color?: Color) {
...
 
+  // iOS PlatformColor
+  if ('semantic' in color) {
+    return [0, color];
+  }
+
+  // Android PlatformColor
+  if ('resource_paths' in color) {
+    return [0, color];
+  }
```
Saadnajmi, line 51: At least on iOS, "semantic" isn't enough, because a color might be a DynamicColor instead. Could we instead do something like a type check? `if (typeof color === ColorValue) ....`

oblador: `DynamicColorIOS` isn't supported in this PR, and I don't really see that API surviving long term as it's iOS only.

src/index.d.ts
oblador: It's not the same, from the type it's `string | OpaqueColorValue` so we could replace the `string` with this one, but not the whole type. I prefer to use `OpaqueColorValue` here as I'm not sure what types will be added to `ColorValue` in the future.

src/lib/extract/extractBrush.ts
Saadnajmi: Surviving in react-native, or surviving in this repo? I don't think it's going anywhere as long as PlatformColor exists. We also have a react-native-macos equivalent `DynacmicColorMacOS` that won't go anywhere as long as DynamicColor exists.

src/index.d.ts
Saadnajmi: I was under the impression that `ColorValue` is just the standard Color type in React Native 0.63+, and that `OpaqueColorValue` will include all the extra types of colors that can be passed to native.  In react-native-macos (which we recently added support for in react-native-svg), I extended ColorValue/OpaqueColorValue to include a new API `ColorWithSystemEffectMacOS` (https://github.com/microsoft/react-native-macos/pull/751). 

But also, I thought the idea of `OpaqueColorValue` is that we do not need to know the specifics of the type, as long as it gets passed over the bridge to `RCTConvert UIColor:]` (or whatever the equivalent on Android is), React Native will make sure it's a valid native color. Is there a reason we need to know the specifics of the types of colors now? I was under the impression we could abstract most of that away to React Native and would need less processing on this repo?

src/lib/extract/extractBrush.ts
oblador: I meant in `react-native`. `DynamicColorIOS` predates `PlaformColor`, and while it's not able to fully replace it just yet, I don't see that API as useful without cross platform support. I guess is possible to add however.

src/index.d.ts
oblador: You can find the definition [here](https://github.com/DefinitelyTyped/DefinitelyTyped/blob/master/types/react-native/index.d.ts#L6077):
```ts
export type ColorValue = string | OpaqueColorValue;
```

So it's not the same, and it can't be replaced as the way this library deals with colors is quite custom. I'm not the author of this library, but i guess it's partly due to color inheritance. If it was not dealing with colors in a custom way then this PR wouldn't be necessary.

gaodeng:
@oblador  @Saadnajmi  @brentvatne @msand 
Based on oblador's code, I added DynamicColorIOS support. Can someone please help to review and merge it?
https://github.com/oblador/react-native-svg/pull/1

Gregoirevda:
Any update to support PlatformColor?

Gregoirevda:
@oblador I've installed your PR branch through npm, but it doesn't compile with Babel: 
`The right-hand side of an 'in' expression must be of type 'any', an object type or a type parameter`
for
`if ('semantic' in color) {` and `  if ('resource_paths' in color) {`

apple/Brushes/RNSVGSolidColorBrush.m
```diff
@@ -13,48 +13,49 @@
 
 @implementation RNSVGSolidColorBrush
 {
-    CGColorRef _color;
+    UIColor *_color;
```
WoLewicki, line 16: I believe we should use `RNSVGColor` so it compiles on `macOS`.

apple/Brushes/RNSVGSolidColorBrush.m
```diff
@@ -13,48 +13,49 @@
...
-    CGColorRelease(_color);
+    _color = nil;
 }
 
 - (CGColorRef)getColorWithOpacity:(CGFloat)opacity
 {
-    return CGColorCreateCopyWithAlpha(_color, opacity * CGColorGetAlpha(_color));
+    CGColorRef baseColor = _color.CGColor;
+    CGColorRef color = CGColorCreateCopyWithAlpha(baseColor, opacity * CGColorGetAlpha(baseColor));
+    CGColorRelease(baseColor);
```
WoLewicki, line 44: We should not release it since it will release `CGColor` prop of `_color`, so trying to use `_color.CGColor` again will lead to crash.

WoLewicki (approved):
I left some suggestions regarding the code, which we can fix in a follow-up PR. Other from that, it looks like it's ready to be merged :tada:

src/lib/extract/extractBrush.ts
```diff
@@ -40,6 +40,16 @@ export default function extractBrush(color?: Color) {
     return int32ARGBColor;
   }
 
+  // iOS PlatformColor
+  if ('semantic' in color) {
```
WoLewicki, line 44: I think here we should also add a check for if `color` is an object, as suggested here by @gaodeng :  https://github.com/oblador/react-native-svg/pull/1/files#diff-955398674a5278517eeeb9b5daca466f8e891b408d79757c3afbfa58a552a324R78

WoLewicki:
One more thing about `Android`, I tried to set `color` prop with `PlatformColor` on `Svg` and it throws error unless we add changes in `android` folder from this PR: https://github.com/oblador/react-native-svg/pull/1/files. We can also do it in a follow-up PR so it is OK for now.

apple/Brushes/RNSVGSolidColorBrush.m
oblador: Are you sure about that? Why would releasing _baseColor affect _color?

oblador: Oh, now I see, good catch 👍

PR #1574, opened 2021-04-27
💄 Update package to use generated types

wcandillon:
# Summary

This now enables react-native-svg to use the generated types from bob. The manually edited types in src/ being incomplete.

The particular issue I was hitting is that getTotalLength is not available in src/index.d.ts

## Test Plan

```
const ref = useRef<Path>(null);
return (
 <Path
        onLayout={() => console.log(ref.current!.getTotalLength())}
        ref={ref}
  />
);
```

## Compatibility

| OS      | Implemented |
| ------- | :---------: |
| iOS     |    ✅     |
| Android |    ✅     |

## Checklist

### DIFF
--- a/package.json
+++ b/package.json
@@ -11,7 +11,7 @@
   "main": "lib/commonjs/index.js",
   "module": "lib/module/index.js",
   "react-native": "src/index.ts",
-  "types": "src/index.d.ts",
+  "types": "lib/typescript/index.d.ts",
   "files": [
     "__tests__",
     "android",


wcandillon:
I recommend to delete https://github.com/wcandillon/react-native-svg/commits/develop/src/index.d.ts if possible

wcandillon:
closing in favor of #1575. It would be nice to use the generated types directly but it looks like it would be a bit more work.

PR #1584, opened 2021-05-09
Exclude macos files on tvOS

jhen0409:
# Summary

Including macos files will cause compile errors for tvOS app.

## Test Plan

Use react-native-svg in [react-native-tvos](https://github.com/react-native-tvos/react-native-tvos) app.


dlowder-salesforce (approved):
This is the right fix for tvOS, makes new projects compile correctly.

Saadnajmi:
@amgleitman

PR #1708, opened 2022-02-25
feat: remove index.d.ts and flow

WoLewicki:
# Summary

PR removing `index.d.ts` files in favor of taking types straight from `lib/typescript`. Since it highly changes the `flow` file generated due to not having one `index.d.ts` file from which `flow` would take types, generating flow types has been removed.

## Test Plan

Check if the types used inside files work the same as the ones from `index.d.ts` and fix incompatibility issues.


Stephen2:
We should finish this feat, it's a good one

WoLewicki:
Yeah, it would be a nice change for sure, but I am a bit stuck with other changes that have higher priority. If you have a time window, it would be great to push it forward.

Stephen2:
I bet I could, I'll figure out how to progress this

WoLewicki:
Closing since #1806 has been merged.

PR #1717, opened 2022-03-04
Add web example

zoontek:
Screenshot 2022-03-04 at 13 35 38

# Summary

Following https://github.com/react-native-svg/react-native-svg/pull/1294

I tried to keep the changes as minimal as possible, but I had to:

- Remove `gradle` from `.gitignore` (not supposed to be here)
- Add `jsxBracketSameLine` until `prettier` is updated (the current version does not support it, it clashes with `prettier-eslint` and it's impossible to work)
- Delete custom `Modal` (now included in `react-native-web`)
- Lint the codebase
- Fix some typescript errors
- Move `getHasTouchableProperty` to the web implem (it's only used here) + remove `any`
- Delete `SnackImageCompat` (doesn't work at all, use inheritance other composition, results in a `Super expression must either be null or a function` error)

Screenshot 2022-03-04 at 13 15 45

And as mentioned in the original PR:

- The `PanResponder` example does not work:

Screenshot 2022-03-04 at 13 44 46

- The focus issue is sure fixed, but pressability just don't work anymore on web:

Screenshot 2022-03-04 at 13 34 50

https://user-images.githubusercontent.com/1902323/156766367-69d13dc1-3c5b-4b1b-b0cf-8a0f25905118.mov

## Checklist


src/ReactNativeSVG.web.ts
```diff
@@ -61,6 +60,9 @@ interface BaseProps {
   style: Iterable<{}>;
 }
 
+const getHasTouchableProperty = (props: BaseProps) =>
```
WoLewicki, line 63: Could you change `getHasTouchableProperty` to just `hasTouchableProperty`? `getHas` sounds weird to me, but no strong opinion on that.

Example/src/examples/Image.tsx
```diff
@@ -83,7 +82,8 @@ class DataURI extends Component {
           y="5%"
           width="90%"
           height="90%"
-          href={{uri: dataUriExample}}
+          // @ts-expect-error
```
WoLewicki, line 85: Why is it needed? Is the type of `href` in `Image` wrong?

WoLewicki (commented):
I added some comments, please answer them and I think we can merge it then. Thanks for your contribution :tada:

src/ReactNativeSVG.web.ts
zoontek: @WoLewicki Done ✅

Example/src/examples/Image.tsx
zoontek: @WoLewicki Yes, it's currently incorrect. I will not display the image on web when data is base64 overwise.

WoLewicki (approved):
LGTM :tada:

PR #1266, opened 2020-01-28
feat(svgUri): add onError prop to SvgUri/Xml/Ast

uqmessias:
# Summary

This solves #1105 and only affects the ts part.

## Compatibility

| OS      | Implemented |
| ------- | :---------: |
| iOS     |    ✅     |
| Android |    ✅     |

PR #1718, opened 2022-03-05
Add support for setting an onError prop on SvgCssUri

SjaakSchilperoort:
# Summary

In PR #1266, support was added for setting an `onError` property on `SvgUri`. This PR adds the same to `SvgCssUri`.

## Test Plan

This should now work: 

```ts
import React, {Component} from 'react';
import {  SvgCssUri } from 'react-native-svg';

class Example extends Component {

    state={
      uri: 'https://dev.w3.org/SVG/tools/svgweb/samples/svg-files/xruby.svg',
    }

    render() {
        return (
            <SvgCssUri width={100} onError={() => this.setState({uri: 'https://dev.w3.org/SVG/tools/svgweb/samples/svg-files/ruby.svg'})} height={100} uri={this.state.uri} />
        );
    }
}
```

## Compatibility

| OS      | Implemented |
| ------- | :---------: |
| iOS     |    ✅❌     |
| Android |    ✅❌     |

## Checklist

## Note

This PR is on request from @WoLewicki in PR #1503.

### DIFF
--- a/src/index.d.ts
+++ b/src/index.d.ts
@@ -393,7 +393,12 @@ export interface StopProps {
 export const Stop: React.ComponentClass<StopProps>;
 export type Stop = React.ComponentClass<StopProps>;
 
-export interface SvgProps extends GProps, ReactNative.ViewProperties {
+export type AdditionalProps = {
+  onError?: (error: Error) => void;
+  override?: Object;
+};
+
+export interface SvgProps extends AdditionalProps, GProps, ReactNative.ViewProperties {
   width?: NumberProp;
   height?: NumberProp;
   viewBox?: string;
--- a/src/index.d.ts
+++ b/src/index.d.ts
@@ -393,12 +393,7 @@ export interface StopProps {
 export const Stop: React.ComponentClass<StopProps>;
 export type Stop = React.ComponentClass<StopProps>;
 
-export type AdditionalProps = {
-  onError?: (error: Error) => void;
-  override?: Object;
-};
-
-export interface SvgProps extends AdditionalProps, GProps, ReactNative.ViewProperties {
+export interface SvgProps extends GProps, ReactNative.ViewProperties {
   width?: NumberProp;
   height?: NumberProp;
   viewBox?: string;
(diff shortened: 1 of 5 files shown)


src/index.d.ts
```diff
@@ -389,7 +389,12 @@ export interface StopProps {
 export const Stop: React.ComponentClass<StopProps>;
 export type Stop = React.ComponentClass<StopProps>;
 
-export interface SvgProps extends GProps, ReactNative.ViewProperties {
+export type AdditionalProps = {
```
WoLewicki, line 392: I think we should not add it to `SvgProps` since `Svg` does not have `onError` prop. Instead, I think it should be added to `UriProps` and `XmlProps`, similar to how it's done here: https://github.com/react-native-svg/react-native-svg/blob/2fa9645cf6588231b8435301f736a721d8c0a311/src/xml.tsx#L89. So the code updated in `index.d.ts` should be `UriProps`: https://github.com/react-native-svg/react-native-svg/blob/2fa9645cf6588231b8435301f736a721d8c0a311/src/index.d.ts#L545 and the one below: `XmlProps`.

WoLewicki:
Also, could you add a test example in [`TestsExample`](https://github.com/react-native-svg/react-native-svg/tree/2fa9645cf6588231b8435301f736a721d8c0a311/TestsExample) app with those simple code snippets?

WoLewicki:
Is it ready to be merged in your opinion @SjaakSchilperoort ?

SjaakSchilperoort:
I looked into adding example code in `TestsExample`, but I think it makes more sense to document the use of `onError` in `README.md`. If you agree I can add an example there, but otherwise I'm fine with this change.

WoLewicki:
If it is possible, I would opt for including both documentation and a test in  `TestsExample`.

SjaakSchilperoort:
> If it is possible, I would opt for including both documentation and a test in `TestsExample`.

Done

WoLewicki (approved):

Issue #1740, opened 2022-04-08
App crash when rendering SVG


antondalgren:
# Bug

Rendering the following SVG causes app to crash, it is reproducible by rendering it in a clean project with the latest versions.

It is throwing errors on `style="transform:translate3d"`, `style="transform:translate"` and `fill="859e9d"` that is missing a `#` to indicate that the value is hex. I have not managed to understand what causes the app crash, but I think it might be related to text animation.

<details>
<summary>SVG</summary>

```xml
<svg width="290" height="500" viewBox="0 0 290 500" xmlns="http://www.w3.org/2000/svg" xmlns:xlink='http://www.w3.org/1999/xlink'>
  <defs>
    <filter id="f1">
      <feImage result="p0" xlink:href="data:image/svg+xml;base64,PHN2ZyB3aWR0aD0nMjkwJyBoZWlnaHQ9JzUwMCcgdmlld0JveD0nMCAwIDI5MCA1MDAnIHhtbG5zPSdodHRwOi8vd3d3LnczLm9yZy8yMDAwL3N2Zyc+PHJlY3Qgd2lkdGg9JzI5MHB4JyBoZWlnaHQ9JzUwMHB4JyBmaWxsPScjODU5ZTlkJy8+PC9zdmc+" />
      <feImage result="p1" xlink:href="data:image/svg+xml;base64,PHN2ZyB3aWR0aD0nMjkwJyBoZWlnaHQ9JzUwMCcgdmlld0JveD0nMCAwIDI5MCA1MDAnIHhtbG5zPSdodHRwOi8vd3d3LnczLm9yZy8yMDAwL3N2Zyc+PGNpcmNsZSBjeD0nMjQzJyBjeT0nMjk4JyByPScxMjBweCcgZmlsbD0nI2M3Nzg0MScvPjwvc3ZnPg==" />
      <feImage result="p2" xlink:href="data:image/svg+xml;base64,PHN2ZyB3aWR0aD0nMjkwJyBoZWlnaHQ9JzUwMCcgdmlld0JveD0nMCAwIDI5MCA1MDAnIHhtbG5zPSdodHRwOi8vd3d3LnczLm9yZy8yMDAwL3N2Zyc+PGNpcmNsZSBjeD0nMjM2JyBjeT0nMjY0JyByPScxMjBweCcgZmlsbD0nIzVhOTFiOCcvPjwvc3ZnPg==" />
[...]

PR #1744, opened 2022-04-11
fix: Handle hex colors that does not start with "#" or "0x".

antondalgren:
# Summary
Minor partial fix of #1740 . This PR fixes the issue with colors containing a hex value, but without the `#`or `0x` start.

## Test Plan

I have not managed to get a local version of the library to work in my test app, so could not take on the major parts of the issue.
A value of `fill="888888"`

Render the SVG with the color.

## Compatibility

| OS      | Implemented |
| ------- | :---------: |
| iOS     |    ✅     |
| Android |    ✅     |

## Checklist

### DIFF
--- a/src/lib/extract/extractBrush.ts
+++ b/src/lib/extract/extractBrush.ts
@@ -30,6 +30,10 @@ export default function extractBrush(color?: Color) {
     return contextStrokeBrush;
   }
 
+  if (typeof color === 'string' && color.match(/^[0-9a-f]{6}$/)) {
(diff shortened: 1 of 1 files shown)


WoLewicki:
Could you check if it is needed on the newest `main` branch? I changed the implementation to use `react-native`'s `processColor` directly, so it should handle all cases which are handled by RN itself.

WoLewicki:
Closing due to no response provided. Feel free to comment to reopen.

PR #1786, opened 2022-06-08
Do not include build artifacts in the npm package

### LINKED ISSUE #1785: NPM library ships with unnecessary android build files
# Bug

The npm version of `react-native-svg` gets pulled in with a bunch of intermediate android build resources in `node_modules/react-native-svg/android/build`.

These files appear to get rebuilt when starting a react-native android app, and removing that directory before building doesn't seem to cause any problems. The entire `android/build` directory could probably be included in `.npmignore` to reduce the package size from 12MB to 2.5MB without affecting anything.

## Environment info

Library version: 12.3.0

TikiTDO:
Fixes #1785

### DIFF
--- a/package.json
+++ b/package.json
@@ -20,7 +20,7 @@
     "lib",
     "src",
     "RNSVG.podspec",
-    "!build"
+    "!android/build"
   ],
   "@react-native-community/bob": {
     "source": "src",
--- a/package.json
+++ b/package.json
@@ -19,7 +19,8 @@
     "elements",
     "lib",
     "src",
-    "RNSVG.podspec"
+    "RNSVG.podspec",
+    "!build"
   ],
   "@react-native-community/bob": {
     "source": "src",
--- a/.npmignore
+++ b/.npmignore
@@ -1,2 +1,5 @@
 screenshots/
 idl/
+
+# Android/IOS Build Artifacts
+build/
--- a/.npmignore
+++ b/.npmignore
@@ -1,5 +1,2 @@
 screenshots/
 idl/
-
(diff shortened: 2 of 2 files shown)


WoLewicki (commented):
Wouldn't it be better to exclude it in [`files`](https://github.com/react-native-svg/react-native-svg/blob/1dc42ea8a8505559d30318377f8c104c7193a3cf/package.json#L15) field of `package.json`? Something like this is done e.g. here: https://github.com/software-mansion/react-native-screens/blob/cc5de7ba28e90ecce9bbb3464ef0a625eaae032f/package.json#L43

TikiTDO:
@WoLewicki That seems reasonable too, both places seem to accomplish roughly the same thing, and both files are present in the codebase, so it's a matter of stylistic preference. I can change the PR to do it in `package.json` if you'd like.

WoLewicki:
I think it would be a better option since I don't see `.npmignore` used too much in other repos.

TikiTDO:
Updated

package.json
```diff
@@ -19,7 +19,8 @@
     "elements",
     "lib",
     "src",
-    "RNSVG.podspec"
+    "RNSVG.podspec",
+    "!build"
```
WoLewicki, line 23: Isn't it too generic now? Maybe something like this would be more reasonable:

```suggestion
    "!android/build"
```

WoLewicki:
I think not since I don't see any excludes in other repos. I think we can merge it now :tada: Thanks for your contribution!

TikiTDO:
Thanks for the feedback. Looking forward to the next release.

PR #1821, opened 2022-08-02
feat: add Fabric on iOS without ComponentViews

WoLewicki:
# Summary

Version of #1754 without usage of `ComponentView`s. It seems like a more proper way, but introduces the necessity of clearing whole state of each component on recycling for it not to be used when view is recycled.

Still known problems:
- We stringify props of type `NumberProp` since codegen only accepts props of a single type. It is the fastest way of dealing with it, but maybe there could be a better way to handle it.
- Image resolving should be probably handled the same as in RN core
- SvgView needs to set `opaque = NO` so it is does not have black background (it comes from the fact that `RCTViewComponentView` overrides `backgroundColor` setter which in turn somehow messes with the view being `opaque`). All other svg components do it already so maybe it is not such an issue.
- `transform` prop won't work when set natively since it is not parsed in Fabric


src/fabric/SvgViewNativeComponent.ts
```diff
@@ -0,0 +1,25 @@
...
+  tintColor: ColorValue,
+  color: ColorValue,
+}
+
+// eslint-disable-next-line @typescript-eslint/ban-types
+const codegenNativeComponent = codegenNativeComponentUntyped as <T extends {}>(
+  name: string
+) => HostComponent<T>;
+
+export default codegenNativeComponent<NativeProps>('RNSVGSvgView');
```
kkafar, line 25: It would be nice thing to have all NativeComponent interfaces declared in the same language (all in Flow / TS). This one stands out.

apple/Elements/RNSVGImage.mm
```diff
@@ -25,13 +24,107 @@
...
+#import "RCTImagePrimitivesConversions.h"
+#import "RCTImageSource.h"
+
+// Some RN private method hacking below similar to how it is done in RNScreens:
+// https://github.com/software-mansion/react-native-screens/blob/90e548739f35b5ded2524a9d6410033fc233f586/ios/RNSScreenStackHeaderConfig.mm#L30
+@interface RCTBridge (Private)
++ (RCTBridge *)currentBridge;
+@end
+
+#endif
```
kkafar, line 44: Just a style thing. Up to you. 
Personally I like when `#endif` is annotated with with information "what if statement it ends". It improves readability in long run & helps to navigate the code especially when there are some nested ifs. 

```
#ifdef RN_FABRIC_ENABLED
...
#endif // RN_FABRIC_ENABLED
```

kkafar (commented):
I left some comments regarding React Native 0.69+, but then I noticed that you went for 0.68.1 in `FabricExample`, so thats fine. 

Pointed some style-regarding things below.

PR #1835, opened 2022-08-15
Web: Support for transform prop

RobertSasak:
# Summary

Transform prop is ignored on web. This PR stringify transform props so it is compatible with web SVG. It is an extension of https://github.com/react-native-svg/react-native-svg/issues/1824

## Compatibility

| OS      | Implemented |
| ------- | :---------: |
| iOS     |    no change     |
| Android |    no change    |
| Web |    ✅     |

## Checklist

- [X ] I have tested this on a device and a simulator

### DIFF
--- a/src/ReactNativeSVG.web.ts
+++ b/src/ReactNativeSVG.web.ts
@@ -151,7 +151,7 @@ const prepare = <T extends BaseProps>(
         transformArray.push(`translate(0 ${value})`);
       } else if (key === 'originX') {
         transformArray.push(`translate(${-value} 0)`);
-      } else if (key === 'originX') {
+      } else if (key === 'originY') {
         transformArray.push(`translate(0 ${-value})`);
       } else if (key === 'origin') {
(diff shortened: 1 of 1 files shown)


src/ReactNativeSVG.web.ts
```diff
@@ -124,36 +137,82 @@ const prepare = <T extends BaseProps>(
...
+    for (const key in transform) {
+      const value = transform[key];
+      // non standard SVG transforms
+      if (key === 'translateX' || key === 'x') {
+        transformArray.push(`translate(${value} 0)`);
+      } else if (key === 'translateY' || key === 'y') {
+        transformArray.push(`translate(0 ${value})`);
+      } else if (key === 'originX') {
+        transformArray.push(`translate(${-value} 0)`);
+      } else if (key === 'originX') {
```
WoLewicki, line 154:
```suggestion
      } else if (key === 'originY') {
```

PR #1842, opened 2022-08-18
Include proguard config

heroic:
# Summary

This will remove the requirement for people to add the proguard configs themselves. A lot of people first release to production with proguard and then realize why the app crashes. This will solve such issues.

This uses android's consumer proguard setting to enable specific proguard config.

## Test Plan

Example app works. All tests pass.

Example app works. All tests pass.

Proguard enabled release should work without any manual edits to proguard file in the app.

## Compatibility

| OS      | Implemented |
| ------- | :---------: |
| iOS     |    ❌     |
| Android |    ✅     |

## Checklist


heroic:
@msand @WoLewicki, can you also release with this for `12.*`?

WoLewicki:
@heroic thanks for your contribution! As for releasing it for `12.*`, I think it is no longer necessary since `13.*` should be stable right now. Or am I missing something?

heroic:
> @heroic thanks for your contribution! As for releasing it for `12.*`, I think it is no longer necessary since `13.*` should be stable right now. Or am I missing something?

not sure what's the status of expo with 13.*

WoLewicki:
Expo SDK 47 features version `13.4.0` of the library so it should be ok.

PR #1847, opened 2022-08-23
feat: use codegenNativeComponent to import native views

WoLewicki:
Changed `requireNativeComponent` to `codegenNativeComponent` so that upcoming changes (Static View Configs, Bridgeless Mode and idk what more) in `react-native` are available in the library. Also, types and native components are now taken directly from `fabric` folder to make sure the values passed to the native components are the ones defined in props. It should work on all supported versions since `codegenNativeComponent` function exists from RN v. 0.61.0. Suggested by @RSNara and @cipolleschi

Reason for [`5394bbb` (#1847)](https://github.com/software-mansion/react-native-svg/pull/1847/commits/5394bbbced6c1838e67240997cf4621d7f783197): 
- on `Paper`, `Animated` uses `setNativeProps` method when we set `useNativeDriver`  to `false`, and does not rerender the component. Therefore, new transform lands only in `SvgView` and is parsed in `RCTViewManager.m` .
- on `Fabric`, the same code makes the components rerender. Due to this, information about new transform is passed to the `SvgView` child: `G` , making it apply translations from the transform in its `updateProps` method.
- other than `Animated` use-case, on both archs, if we just passed `transform` prop to `Svg` component, it would end up in double transformations now as well. All of those changes are due to https://github.com/software-mansion/react-native-svg/pull/1895, which added proper parsing of RN style `transform` prop (array of transformations objects) therefore making `G` properly handle `transform` prop passed from `Svg`.

Reason for [`19bcb24` (#1847)](https://github.com/software-mansion/react-native-svg/pull/1847/commits/19bcb2464b3a40a47e4396ea66f49d6a95bfc9bc): Same as https://github.com/software-mansion/react-native-screens/pull/1624


tomekzaw (approved):

tomekzaw (approved):
I can't see `any`thing wrong with this PR.

src/elements/NativeComponents.tsx
```diff
@@ -1,31 +1,29 @@
...
+export const RNSVGPattern = codegenNativeComponent<any>('RNSVGPattern');
+export const RNSVGRadialGradient = codegenNativeComponent<any>(
+  'RNSVGRadialGradient',
+);
+export const RNSVGRect = codegenNativeComponent<any>('RNSVGRect');
+export const RNSVGSymbol = codegenNativeComponent<any>('RNSVGSymbol');
+export const RNSVGText = codegenNativeComponent<any>('RNSVGText');
+export const RNSVGTextPath = codegenNativeComponent<any>('RNSVGTextPath');
+export const RNSVGTSpan = codegenNativeComponent<any>('RNSVGTSpan');
+export const RNSVGUse = codegenNativeComponent<any>('RNSVGUse');
```
RSNara, line 29: @WoLewicki, this won't work unfortunately.

The type passed into codegenNativeComponent needs to be declared in the same file as the call to codegenNativeComponent. Since we're passing any to all these codegenNativeComponent calls, the babel plugin will not transform these into NativeComponetRegistry.get calls. 

So, ideally, the setup should be like this:
1. We delete this NativeComponents.ts file.
2. All TypeScript files use the spec files defined in src/fabric

We might have to translate the files inside src/fabric into typescript.

src/fabric/CircleNativeComponent.ts
```diff
@@ -0,0 +1,56 @@
+import codegenNativeComponent from 'react-native/Libraries/Utilities/codegenNativeComponent';
+import type { ViewProps, ColorValue } from 'react-native';
+import type {
+  Float,
+  Int32,
+  WithDefault,
+  // @ts-ignore TODO: remove once there is a .d.ts file with definitions
```
RSNara, line 7: DefinitelyTyped has annotations for react-native/Libraries/Types/CodegenTypes, here:

https://github.com/DefinitelyTyped/DefinitelyTyped/blob/ffc894660966251e61928e62fffddb4705059ae4/types/react-native/Codegen.d.ts#L31-L61

Do we need this ts-ignore?

apple/ViewManagers/RNSVGImageManager.mm
```diff
@@ -33,7 +33,7 @@ - (RNSVGRenderable *)node
 {
   view.imageheight = [RCTConvert RNSVGLength:json];
 }
-RCT_EXPORT_VIEW_PROPERTY(src, id)
+RCT_EXPORT_VIEW_PROPERTY(src, UIImage)
```
RSNara, line 36: @WoLewicki, did RCTImageSource not work here?

android/src/main/java/com/horcrux/svg/SvgPackage.java
```diff
@@ -10,50 +10,134 @@
...
-        new PatternManager(),
-        new MaskManager(),
-        new ForeignObjectManager(),
-        new MarkerManager(),
-        new SvgViewManager());
+  public @Nullable ViewManager createViewManager(
+    ReactApplicationContext reactContext, String viewManagerName) {
+    ModuleSpec spec = getViewManagersMap(reactContext).get(viewManagerName);
+    return spec != null ? (ViewManager) spec.getProvider().get() : null;
   }
```
RSNara, line 89: Let's implement one more method:

```
  @Override
  public List<ModuleSpec> getViewManagers(ReactApplicationContext reactContext) {
    return new ArrayList<>(getViewManagersMap(reactContext).values());
  }
```

I think this should fix our `requireNativeComponent: "RNSVGCircle" was not found in the UIManager` problem.

apple/RNSVGSvgViewModule.mm
```diff
@@ -0,0 +1,97 @@
...
+RCT_EXPORT_VIEW_PROPERTY(bbWidth, RNSVGLength *)
+RCT_EXPORT_VIEW_PROPERTY(bbHeight, RNSVGLength *)
+RCT_EXPORT_VIEW_PROPERTY(minX, CGFloat)
+RCT_EXPORT_VIEW_PROPERTY(minY, CGFloat)
+RCT_EXPORT_VIEW_PROPERTY(vbWidth, CGFloat)
+RCT_EXPORT_VIEW_PROPERTY(vbHeight, CGFloat)
+RCT_EXPORT_VIEW_PROPERTY(align, NSString)
+RCT_EXPORT_VIEW_PROPERTY(meetOrSlice, RNSVGVBMOS)
+RCT_EXPORT_VIEW_PROPERTY(tintColor, UIColor)
+RCT_REMAP_VIEW_PROPERTY(color, tintColor, UIColor)
```
RSNara, line 31: @WoLewicki, was this a typo? We don't need these, right?

Exporting ViewProperties on TurboModules doesn't make much sense. So, we should get rid of these macros.

skurgansky-sugarcrm:
i guess RN 0.70.1 projects can't use svg without this done, are they?

WoLewicki:
@skurgansky-sugarcrm `react-native-svg` should work on both architectures in the newest version of RN. Does it not work for you?

FabricExample/src/examples/Svg.tsx
```diff
@@ -133,22 +133,22 @@ class SvgNativeMethods extends Component {
...
+    console.log(this.circle?.getCTM());
+    console.log(this.circle?.getScreenCTM());
   };
-  root: any;
-  circle: any;
+  root: Svg | null;
+  circle: Circle | null;
 
   componentDidMount() {
       this.alert();
```
RSNara, line 154: @WoLewicki, should we remove this componentDidMount() method from this example? (Or, we could also comment this out, and document why it doesn't work).

To summarize our discord discussion (for posterity): With Fabric, componentDidMount executes when the shadow tree gets committed (i.e: Fabric won't actually execute componentDidMount **after** the views are mounted on to the screen). So, trying to call a method like this.circle.getCTM(), which performs lookup on the view hierarchy (i.e: calling view.getParent()) inside componentDidMount is a bad idea.

src/fabric/AndroidSvgViewNativeComponent.ts
```diff
@@ -61,4 +61,4 @@ interface NativeProps extends ViewProps {
   // borderTopLeftRadius?: Float;
 }
 
-export default codegenNativeComponent<NativeProps>('RNSVGSvgView');
+export default codegenNativeComponent<NativeProps>('RNSVGSvgViewAndroid');
```
RSNara, line 64: We should be able to re-use the name "RNSVGSvgView" here.

android/src/main/java/com/horcrux/svg/SvgViewManager.java
```diff
@@ -315,32 +315,27 @@ public void setNextFocusLeft(SvgView view, int value) {
...
+  public void setBorderRadius(SvgView view, double value) {
+    super.setBorderRadius(view, 0, (float) value);
   }
 
   @Override
-  @ReactProp(name = "borderBottomRightRadiusAndroid")
-  public void setBorderBottomRightRadiusAndroid(SvgView view, float value) {
-    super.setBorderBottomRightRadius(view, value);
+  public void setBorderTopLeftRadius(SvgView view, double value) {
+    super.setBorderTopLeftRadius(view, (float) value);
```
RSNara, line 324: @WoLewicki, this will call BaseViewManager.setBorderTopLeftRadius().

https://github.com/facebook/react-native/blob/048194849bda980eecf7bb006ca4e71e2d60ff4c/ReactAndroid/src/main/java/com/facebook/react/uimanager/BaseViewManager.java#L629

This will throw, or log errors, when the borderRadius prop is set.

Instead, we want these methods to call ReactViewManager.setBorderRadius(), no?

https://github.com/facebook/react-native/blob/main/ReactAndroid/src/main/java/com/facebook/react/views/view/ReactViewManager.java#L111-L138

That way, React Native will actually assign the borderRadius to the SvgView.

WoLewicki: Yeah, you are right 🚀

PR #1848, opened 2022-08-23
Use angle-bracket imports

tomekzaw:
# Summary

This PR replaces bad double-quotes imports (which break Swift interop) with good angle-bracket imports.

The forbidden pattern is: `#import "RCT`, except for `#import "RCTConvert+RNSVG.h"` which is okay.

See also:
* https://github.com/software-mansion/react-native-reanimated/pull/3150
* https://github.com/software-mansion/react-native-gesture-handler/pull/2180
* https://github.com/software-mansion/react-native-screens/pull/1572

## Test Plan

Check if Example and FabricExample apps compile successfully.

### DIFF
--- a/apple/Elements/RNSVGClipPath.mm
+++ b/apple/Elements/RNSVGClipPath.mm
@@ -9,10 +9,10 @@
 #import "RNSVGClipPath.h"
 
 #ifdef RN_FABRIC_ENABLED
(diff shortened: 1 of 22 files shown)


WoLewicki (approved):
LGTM! Thanks for your "contribution" :tada:

PR #1874, opened 2022-09-16
fix: use `export type` when exporting types

renchap:
# Summary

If you built your app using `esbuild` (for example using [`@rnx-kit/metro-serializer-esbuild`](https://microsoft.github.io/rnx-kit/docs/tools/metro-serializer-esbuild)), then it complains that the imports do not exist for type imports.

Marking them explicitely as types fixes this issue, in addition to removing those import/exports from non-TS builds.

I also transformed them from an import + export statements to an `export … from …` statement.

## Test Plan

This has been tested in my own app which uses the esbuild serializer mentioned above.

It only changes type exports, so no actual behaviour changes are expected

Having an app using the esbuild serializer

## Compatibility

Not relevant

## Checklist

Not relevant

### DIFF
--- a/src/ReactNativeSVG.ts
+++ b/src/ReactNativeSVG.ts
@@ -1,40 +1,30 @@
 import Shape from './elements/Shape';
-import Rect, { RectProps } from './elements/Rect';
-import Circle, { CircleProps } from './elements/Circle';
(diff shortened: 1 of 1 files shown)


tido64:
@WoLewicki, can we backport this to the 12.x branch? If yes, what is the process? I'd be happy to do it myself.

PR #1877, opened 2022-09-21
Fixed RCT_NEW_ARCH_ENABLED in podspec

geroale:
RCT_NEW_ARCH_ENABLED seems to be always true (even when its value is '0') without an explicit check


WoLewicki (approved):

PR #1886, opened 2022-09-29
make reanimated work in web

phryneas:
# Summary

Up until now, trying to use `reanimated` with `react-native-svg` in `react-native-web` resulted in an error.
This adds a `setNativeProps` function to the web implementation to directly modify the `transform` and `style` props on a SVGElement `ref`.

Since there is a need to track the "last merged props" and those need to be reset on every render, the `render` method has been moved into the `WebShape` class and a `tag` string property has been added.

As `g` had some extra handling for `x` and `y`, a `prepareProps` function was added as well.

## Test Plan

Here is a video of me dragging a `<g>` with a nested `<image>` in chome: https://www.loom.com/share/675405537d9d41a792432ebf41636047

Just run any reanimated svg in the browser instead of a device.

Run any animation.

## Compatibility

| OS      | Implemented |
| ------- | :---------: |
| iOS     |    (untouched)     |
| Android |   (untouched)  |
| Web |    ✅     |

## Checklist

- [X] I have tested this on a web browser

### DIFF
--- a/src/ReactNativeSVG.web.ts
+++ b/src/ReactNativeSVG.web.ts
@@ -158,12 +158,30 @@ const prepare = <T extends BaseProps>(
     clean.transform = transform.join(' ');
   }
 
-  clean.ref = (el: SVGElement | null) => {
-    self.elementRef.current = el;
+  clean.ref = (element: SVGElement | null) => {
+    self.elementRef.current = element;
+    /**
+     * If the TouchableMixin is used, this already has been
+     * set to `element` in `componentDidMount`.
+     * We want to override `element.setAttribute`, so we replace
+     *  it with a proxied version (not touching the real DOM node).
+     */
+    self._touchableNode = !element
+      ? element
+      : new Proxy(element, {
+          get(target, p) {
+            if (p === 'setAttribute') {
+              return self.schedulePropUpdate.bind(self);
+            }
+            // @ts-expect-error this is always non-typesafe
+            return target[p];
+          },
+        });
+    // handle `forwardedRef` if it was present
     if (typeof forwardedRef === 'function') {
-      forwardedRef(el);
+      forwardedRef(element);
     } else if (forwardedRef) {
-      forwardedRef.current = el;
+      forwardedRef.current = element;
     }
   };
 
@@ -239,6 +257,13 @@ function remeasure() {
   measureLayout(tag, this._handleQueryLayout);
 }
 
+function dashedToCamelCase(dashedKey: string) {
+  return dashedKey.replace(/-[a-z]/g, (m) => m[1].toUpperCase());
+}
+function camelCaseToDashed(camelCase: string) {
+  return camelCase.replace(/[A-Z]/g, (m) => '-' + m.toLowerCase());
+}
+
 export class WebShape<
   P extends BaseProps = BaseProps,
   C = {},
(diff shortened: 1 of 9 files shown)


necolas:
FYI setNativeProps will not be supported in the new RN architecture and is being removed from RNWeb too

WoLewicki:
Could you add some examples to e.g. `Example` project which would show these changes?

phryneas:
@WoLewicki  I can give it a spin.

By the way, I'm just going at that `_touchableNode` part and stumbled across https://github.com/software-mansion/react-native-reanimated/issues/3321. The solution I'm trying right now might also fix that.

phryneas:
and I added an example :)

phryneas:
Also, maybe I should add some explanation on what this is doing and how it is interacting.

The relevant code on the side of `reanimated` can be found here:

From the users perspective:
* usually a component like an `react-native-svg`-component would be wrapped in a `createReanimatedComponent` HOC that would add an `animatedProps` property
* then the user would use the `useAnimatedProps` hook to create an object to pass into that `animatedProps` property.
* changing shared values mapped that way would now not cause a component rerender, but internally change properties on the rendered component

This is [the example from the docs](https://docs.swmansion.com/react-native-reanimated/docs/api/hooks/useAnimatedProps/#example)
```js
import React from 'react';
import { StyleSheet } from 'react-native';
import Animated, {
  useSharedValue,
  useAnimatedProps,
} from 'react-native-reanimated';
import Svg, { Path } from 'react-native-svg';

const AnimatedPath = Animated.createAnimatedComponent(Path);

function App() {
  const radius = useSharedValue(50);

  const animatedProps = useAnimatedProps(() => {
    // draw a circle
    const path = `
    M 100, 100
    m -${radius.value}, 0
    a ${radius.value},${radius.value} 0 1,0 ${radius.value * 2},0
    a ${radius.value},${radius.value} 0 1,0 ${-radius.value * 2},0
    `;
    return {
      d: path,
    };
  });

  // attach animated props to an SVG path using animatedProps
  return (
    <Svg>
      <AnimatedPath animatedProps={animatedProps} fill="black" />
    </Svg>
  );
}
```

Now, let's look at [the internals of those "off-render-updates"](https://github.com/software-mansion/react-native-reanimated/blob/d04720c82f5941532991b235787285d36d717247/src/reanimated2/js-reanimated/index.ts#L34-L40):

```js
    if (typeof component.setNativeProps === 'function') {
      setNativeProps(component, rawStyles);
    } else if (Object.keys(component.props).length > 0) {
      Object.keys(component.props).forEach((key) => {
        if (!rawStyles[key]) {
          return;
        }
        const dashedKey = key.replace(/[A-Z]/g, (m) => '-' + m.toLowerCase());
        component._touchableNode.setAttribute(dashedKey, rawStyles[key]);
      });
    } else {
      console.warn('It is not possible to manipulate component');
    }
```

The relevant part: 
* if there is a `setNativeProps` function on the component wrapped, that will be called with the whole updated prop object at once. This was my first approach, but it seems `setNativeProps` is deprecated now.
* if that is not present, `reanimated` will instead iterate through all attributes, will convert camelCase to dashed keys and set each attribute individually.

Now, "setting each attribute individually" faces two problems from the side of `react-native-svg`:
* with `react-native-svg`, those cannot be set directly on the DOM element as there is that `prepare` function taking all those properties and converting them to a string `transform` property
* since all of those are combined into one string, they also have to be set at once and not bit-by-bit

Hence, my approach here:
* make sure that there is a `component._touchableNode` property in the first place (up until now, that was only present through the `Touchable` Mixin, which was only added when the user added touch listeners
* wrap `component._touchableNode.setAttribute` to our own logic - I used a proxy on the original DOM node here to override only this one specific property
* "batch" modifications coming in from the outside before building a new `transform` string that will be applied - I do this by scheduling an update with `requestAnimationFrame`, which will make sure that any synchronous calls like the `for` loop of `reanimated` above will have ended before the batch is executed
* if meanwhile a render happens, cancel the batched update and reset the batched properties - the latest render is sure to have the most up-to-date data.

WoLewicki:
After consultation with @kmagiera we came to a conclusion that the first version of this PR using `setNativeProps` seems like a better option for now. `react-native-reanimated` still uses `setNativeProps` on web, so since this PR brings integration with `react-native-reanimated`, it should be fine to use it. Could you revert the changes then?

phryneas:
@WoLewicki sure, I reverted those files.
Should it ever be necessary to go the other route, I will keep a branch around with those changes at https://github.com/phryneas/react-native-svg/tree/reanimated-touchableNode

phryneas:
Seems these CI errors come from the example app, or rather from the dependency to `reanimated`.
Quick solution would be to turn `skipLibCheck` to `false` - would you be okay with that?

WoLewicki:
Yeah, it seems right to not check other libraries' types in PRs for this library 😅

WoLewicki:
I checked the current implementation and the rectangle seems to be stuck in one place, not animating. Also, could you add a `fill` prop to it? On `iOS` and `Android` this rectangle seems to be white by  default so it cannot be seen.

phryneas:
I've changed it to red, so it's visible on Android as well (I don't have an iOS device to check at hand).

For me, the rectangle is animating in both Chrome and Firefox. I'm not really sure what's going on there.. :/ 
Maybe some build cache issues?

WoLewicki:
It is frozen at the beginning for me both on chrome and safari. I tested it on the fresh repo so I am not sure what can be a problem here: 
image

phryneas:
Hmm, I can reproduce that on a different machine. Maybe *I* am the one with the stale cache. I'll look into it.

phryneas:
Got it. The original code only applied changes to `transform` and `style`, both of which were not affected at all by the example. I had fixed that in the `_touchableNode` version of this PR, but not in the initial one.
Now it should animate.

WoLewicki (approved):
LGTM! Thanks for your contribution :tada:

phryneas:
Lol last minute fix, I'm sorry!

phryneas:
I noticed that `style` could also come in as an array, so I handle that case now, too.

WoLewicki:
Ok, could you also run `pod install` in the `Example` app since you added `reanimated` to the project? It should change the `Podfile.lock`.

WoLewicki:
Also, why this is needed: https://github.com/software-mansion/react-native-svg/pull/1886/files#diff-8830d392c035c5b523ac5a9a48b3afbc9a4f7c1b56cce4ad36f7b41960c8042eR289 ?

phryneas:
> Ok, could you also run `pod install` in the `Example` app since you added `reanimated` to the project? It should change the `Podfile.lock`.

I'm on a Linux machine, so I don't have `pod` available, I'm sorry.

> Also, why this is needed: [#1886 (files)](https://github.com/software-mansion/react-native-svg/pull/1886/files#diff-8830d392c035c5b523ac5a9a48b3afbc9a4f7c1b56cce4ad36f7b41960c8042eR289) ?

To assign properties not on `style`, like `transform`, `x` etc. 
When `reanimated` calls these updates directly on a DOM node, they replace camelCase with dashed attribute names first, so I just replicated their logic there.
Relevant code: https://github.com/software-mansion/react-native-reanimated/blob/d04720c82f5941532991b235787285d36d717247/src/reanimated2/js-reanimated/index.ts#L34-L40

WoLewicki:
Ok so could you please add a comment in code with the link to the relevant code you just linked?

src/ReactNativeSVG.web.ts
```diff
@@ -237,6 +248,51 @@ export class WebShape<
...
+            break;
+          case 'style':
+            // style can be an object here or an array, so we convert it to an array and assign each element
+            for (const partialStyle of ([] as {}[]).concat(clean.style ?? [])) {
+              // @ts-expect-error "DOM" is not part of `compilerOptions.lib`
+              Object.assign(current.style, partialStyle);
+            }
+            break;
+          default:
+            // @ts-expect-error "DOM" is not part of `compilerOptions.lib`
```
phryneas, line 288:
```suggestion
            // apply all other incoming prop updates as attributes on the node
            // same logic as in https://github.com/software-mansion/react-native-reanimated/blob/d04720c82f5941532991b235787285d36d717247/src/reanimated2/js-reanimated/index.ts#L38-L39
            // @ts-expect-error "DOM" is not part of `compilerOptions.lib`
```

WoLewicki:
Yeah, looks good now :rocket:

PR #1903, opened 2022-10-27
fix: handle null passed to SvgLength

WoLewicki:
`SvgLength` `from` method should be able to parse `null` since it is a correct value, we then use the empty constructor since it provides default values.


Julesssss (commented):
Thanks, this resolves the issue we're seeing in Expensify/App

Issue #1742, opened 2022-04-09
SvgUri example as given in README.md crashes in Expo Web 


raarts:
# Bug

I created a fresh expo app using `expo init app-simple` using the Typescript template. Added the SvgUri example.
Resulted in the following error in my console:

(I have to add, the standard Expo <Image> component supports svg if you install react-native-svg. If not, it will skip svg files without warning.)

```
index.js:1 Warning: React.jsx: type is invalid -- expected a string (for built-in components) or a class/function (for composite components) but got: undefined. You likely forgot to export your component from the file it's defined in, or you might have mixed up default and named imports.

Check the render method of `App`.
    at App
    at ExpoRootComponent (http://localhost:19007/static/js/bundle.js:8361:83)
    at div
    at http://localhost:19007/static/js/bundle.js:46754:25
    at div
    at http://localhost:19007/static/js/bundle.js:46754:25
    at AppContainer (http://localhost:19007/static/js/bundle.js:38675:24)
console.<computed> @ index.js:1
react-dom.development.js:25058 Uncaught Error: Element type is invalid: expected a string (for built-in components) or a class/function (for composite components) but got: undefined. You likely forgot to export your component from the file it's defined in, or you might have mixed up default and named imports.

Check the render method of `App`.
    at createFiberFromTypeAndProps (react-dom.development.js:25058:1)
    at createFiberFromElement (react-dom.development.js:25086:1)
    at createChild (react-dom.development.js:13446:1)
    at reconcileChildrenArray (react-dom.development.js:13719:1)
    at reconcileChildFibers (react-dom.development.js:14125:1)
    at reconcileChildren (react-dom.development.js:16990:1)
    at updateHostComponent (react-dom.development.js:17632:1)
    at beginWork (react-dom.development.js:19080:1)
    at HTMLUnknownElement.callCallback (react-dom.development.js:3945:1)
    at Object.invokeGuardedCallbackDev (react-dom.development.js:3994:1)
    at invokeGuardedCallback (react-dom.development.js:4056:1)
    at beginWork$1 (react-dom.development.js:23964:1)
    at performUnitOfWork (react-dom.development.js:22776:1)
    at workLoopSync (react-dom.development.js:22707:1)
    at renderRootSync (react-dom.development.js:22670:1)
    at performSyncWorkOnRoot (react-dom.development.js:22293:1)
    at scheduleUpdateOnFiber (react-dom.development.js:21881:1)
[...]

PR #1916, opened 2022-11-09
fix(xml): extract tags map to separate entry point for mobile & web

joshuayoes:
# Summary

**What issues does the pull request solve? Please tag them so that they will get automatically closed once the PR is merged**

When utilizing with `react-native-web`: SvgXml and SvgUri are not exported in the `src/ReactNativeSVG.web.ts` extension. The logic for SvgXml and SvgUri do not have native dependencies, so they are safe to consume on web.

Issues:
- https://github.com/software-mansion/react-native-svg/issues/1279
- https://github.com/software-mansion/react-native-svg/issues/1742

**What is the feature? (if applicable)**

Add SvgXml and SvgUri as consumable exports for `react-native-web`

**How did you implement the solution?**

Extract `tags` to shapes map into separate `tags.tsx` file, where native shape elements and web shape elements can be provided to their respective envs. 

**What areas of the library does it impact?**

`src` directory

## Test Plan

n/a

n/a

## Compatibility

| OS      | Implemented |
| ------- | :---------: |
| iOS     |    ✅     |
| Android |    ✅     |

## Checklist


joshuayoes:
Another way I considered going about this was creating `web.tsx` shadows for each exported element in the `src/elements` directory. 
Screen Shot 2022-11-09 at 1 02 29 PM

This would be nice because it would allow `src/ReactNativeSVG.web.ts` to more closely resemble `src/ReactNativeSVG.ts`. But I figured it would make the PR noisier with lots of smaller file changes.

WoLewicki:
@joshuayoes thanks for the contribution! Do you have knowledge how does it go with the discussion from #1279? Especially security concerns e.g. from https://github.com/software-mansion/react-native-svg/issues/1279#issuecomment-595269518?

joshuayoes:
> @joshuayoes thanks for the contribution! Do you have knowledge how does it go with the discussion from #1279? Especially security concerns e.g. from https://github.com/software-mansion/react-native-svg/issues/1279#issuecomment-595269518?

Thanks for your response. Taking a deeper look at those issues, I think this PR may not be a good solution for maximal security, which it seems the team has decided to prioritize.

However, it would be nice if there was a clearer error when trying to import these on the web. Perhaps there could be a web export that just throws an error with a link to the PRs?

WoLewicki:
Yeah, it seems like a good solution. Maybe even better if there was a link to the #1279 and a note that if you want to use it, then there could be a link to the PRs. Would you mind opening such a PR?

bohdanprog:
Hello @joshuayoes,
Would you be able to resolve the conflicts, please?

src/xml.tsx
```diff
@@ -388,7 +339,7 @@ export function parse(source: string, middleware?: Middleware): JsxAST | null {
       props,
       children: [],
       parent: currentElement,
-      Tag: tags[tag] || missingTag,
+      Tag: getTag(tag as keyof typeof tags),
```
jakex7, line 342: Why do we need this change? Shouldn't tags from `tags.tsx` and `tags.web.tsx` be exactly the same?

src/xml.tsx
```diff
@@ -485,6 +436,16 @@ export function parse(source: string, middleware?: Middleware): JsxAST | null {
     return name;
   }
 
+  function getTag(
+    name: keyof typeof tags
+  ): ComponentType<{ children?: React.ReactNode }> {
```
jakex7, line 441: I don't think that `ComponentType<{ children?: React.ReactNode }>` is the right type for Svg tags, i.e. `Circle` or `Rect`

src/xml.tsx
```diff
@@ -485,6 +436,16 @@ export function parse(source: string, middleware?: Middleware): JsxAST | null {
...
+    type Element = (typeof tags)[keyof typeof tags];
+    const Tag: Element | undefined = tags[name];
+    // eslint-disable-next-line @typescript-eslint/ban-ts-comment
+    // @ts-ignore
+    return Tag || missingTag;
```
jakex7, line 446: That TS error is valid and should not be muted. We should correct return type of `getTag` function instead.

bohdanprog: fixed.

src/xml.tsx
bohdanprog: removed that function.

src/xml.tsx
bohdanprog: changed that, and restored the old version.

jakex7 (approved):
LGTM, thank you! 🚀

Issue #1942, opened 2022-12-15
Path's onLayout works fine?


kr-yeon:
Currently working with svg. I needed the horizontal size, vertical size, and x and y values ​​of the path, so I tried to bring them into onLayout, but it seems to be getting the size of the SVG tag.

'''event.nativeEvent.layout'''

 LOG {"height": 0, "width": 370.28570556640625, "x": 0, "y": 0}
 LOG {"height": 0, "width": 370.28570556640625, "x": 0, "y": 0}
 LOG {"height": 0, "width": 370.28570556640625, "x": 0, "y": 0}
 LOG {"height": 0, "width": 370.28570556640625, "x": 0, "y": 0}
 LOG {"height": 0, "width": 370.28570556640625, "x": 0, "y": 0}
 LOG {"height": 0, "width": 370.28570556640625, "x": 0, "y": 0}
 LOG {"height": 0, "width": 370.28570556640625, "x": 0, "y": 0}
 LOG {"height": 0, "width": 370.28570556640625, "x": 0, "y": 0}
 LOG {"height": 0, "width": 370.28570556640625, "x": 0, "y": 0}
 LOG {"height": 0, "width": 370.28570556640625, "x": 0, "y": 0}
 LOG {"height": 0, "width": 370.28570556640625, "x": 0, "y": 0}
 LOG {"height": 0, "width": 370.28570556640625, "x": 0, "y": 0}
 LOG {"height": 0, "width": 370.28570556640625, "x": 0, "y": 0}
 LOG {"height": 0, "width": 370.28570556640625, "x": 0, "y": 0}

react-native-svg: 13.6.0
platform: android

kr-yeon:
iPhone works fine. 

 LOG  {"height": 0, "width": 372, "x": 0, "y": 0}
 LOG  {"height": 0, "width": 372, "x": 0, "y": 0}
 LOG  {"height": 0, "width": 372, "x": 0, "y": 0}
 LOG  {"height": 0, "width": 372, "x": 0, "y": 0}
 LOG  {"height": 0, "width": 372, "x": 0, "y": 0}
 LOG  {"height": 0, "width": 372, "x": 0, "y": 0}
 LOG  {"height": 0, "width": 372, "x": 0, "y": 0}
 LOG  {"height": 0, "width": 372, "x": 0, "y": 0}
 LOG  {"height": 0, "width": 372, "x": 0, "y": 0}
[...]

PR #1998, opened 2023-03-08
fix onLayout 

toter3:
# Summary

when upgrading from v12.1.1 to latest Iv'e noticed onLayout doesn't receive correct values in Android (wrong value for width and height, x, y are all 0)

`LOG: event.nativeEvent.layout:  {"height": 0, "width": 300, "x": 0, "y": 0}`

also found [this issue](https://github.com/software-mansion/react-native-svg/issues/1942) that seems related 

tracing back it seems like it breaks on v13.0

**the fix**:
when debugging i saw that in the new version it never gets to the code dispatching the event (while in v12.4.4 it does)
because there's this if statement checking if `mOnLayout` set to true

_VirtualView.Java:578_
```
   if (!mOnLayout) {
      return;
    }
    EventDispatcher eventDispatcher =
        mContext.getNativeModule(UIManagerModule.class).getEventDispatcher();
    eventDispatcher.dispatchEvent(OnLayoutEvent.obtain(this.getId(), left, top, width, height));
```

this property is never being set to true because the function for ReactProp onLayout is never being called

_RenderableViewManager.java:473_
```
 @ReactProp(name = "onLayout")
  public void setOnLayout(V node, Boolean onLayout) {
    node.setOnLayout(onLayout);
  }
```

looking at other functions here, these function are now being called from the shapes' classes in `facebook.react.viewmanagers/` but there was no call to onLayout

* it seems like it is new to 13.0 version (following two commits implementing usage of Fabric)

my fix is just adding a call to setOnLayout() for all relevant shapes' classses.

* What areas of the library does it  #impact?
this affect just Android - onLayout event for Circle, Path, Ellipse, G, Line, Text, Image

## Test Plan

for testing i added  a 'onLayout prop' to the first component in each example in the example app
e.g. _examples/Circle.tsx:6_
```
 render() {
    return (
      <Svg height="100" width="140">
        <Circle cx="50%" cy="50%" r="40%" fill="pink" onLayout={(e)=>console.log(e.nativeEvent.layout)}/>
      </Svg>
    );
  }
```

when rendered:
```
 LOG  {"height": 0, "width": 100, "x": 0, "y": 0}
 LOG  {"height": 90.18181610107422, "width": 90.18181610107422, "x": 4.7272725105285645, "y": 4.7272725105285645}
```

and for comparison, before the fix it looked like this 
` LOG  {"height": 0, "width": 100, "x": 0, "y": 0}`

### What's required for testing  #(prerequisites)?
add a 'onLayout' function as prop to any svg component and follow the values in event.nativeEvent.layout

## Compatibility

| OS      | Implemented |
| ------- | :---------: |
| iOS     |    ❌     | (working with and without the fix)
| Android |    ✅     |

## Checklist

### DIFF
--- a/FabricExample/ios/Podfile.lock
+++ b/FabricExample/ios/Podfile.lock
@@ -713,7 +713,7 @@ PODS:
     - React-jsi (= 0.71.0)
     - React-logger (= 0.71.0)
     - React-perflogger (= 0.71.0)
-  - RNSVG (13.7.0):
+  - RNSVG (13.8.0):
     - RCT-Folly
     - RCTRequired
     - RCTTypeSafety
@@ -721,8 +721,8 @@ PODS:
     - React-Codegen
     - React-RCTFabric
     - ReactCommon/turbomodule/core
-    - RNSVG/common (= 13.7.0)
-  - RNSVG/common (13.7.0):
+    - RNSVG/common (= 13.8.0)
+  - RNSVG/common (13.8.0):
     - RCT-Folly
     - RCTRequired
     - RCTTypeSafety
@@ -951,7 +951,7 @@ SPEC CHECKSUMS:
   React-rncore: cfeb5532ec459f562410e8058b8f49e07cd215d4
   React-runtimeexecutor: ac80782d9d76ba2b0f709f4de0c427fe33c352dc
   ReactCommon: 20e38a9be5fe1341b5e422220877cc94034776ba
-  RNSVG: 27a7d7438182ca3323de0c57695e3bdaf1aa5eb4
+  RNSVG: 6b96501d7df80ed8b4fe55d903b4f1476a57e8a5
   SocketRocket: fccef3f9c5cedea1353a9ef6ada904fde10d6608
   Yoga: c618b544ff8bd8865cdca602f00cbcdb92fd6d31
   YogaKit: f782866e155069a2cca2517aafea43200b01fd5a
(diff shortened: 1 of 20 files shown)


maxkomarychev:
awesome, thank you @toter3 !

pbaker0804:
Thanks @toter3!  I just ran into this issue with <Rect ... onLayout={...}/>

toter3:
@pbaker0804 idk baout the merge, depends on the team here.
i hope soon (:

WoLewicki:
Sorry for the late response. I think it would be better to add `onLayout` to the codegen type definitions in specs (here: https://github.com/software-mansion/react-native-svg/tree/main/src/fabric) so it is autogenerated and then add those generated methods to the folder as you did. It will ensure that it works correctly both on old and new arch. Could you do it @toter3 or do you want me to handle this?

toter3:
@WoLewicki Hi thanks for the review, I'm not sure i understand where to add it and what to run in order to autogenerate,
if it wouldn't take too much of your time, i'd be happy if you could do it, or add more details so i could do it myself.

btw, i'll appreciate if you find time for another PR review [here](https://github.com/software-mansion/react-native-svg/pull/2001)
thanks!

WoLewicki:
Ok I will try and handle it. As for the other PR, I'll look at it for sure, thanks!

toter3:
@WoLewicki thanks! I'v added your commit and tested both platforms, seems to be working great

WoLewicki:
@toter3 I added one more commit removing unnecessary method on Android. Could you merge it here and we can merge the PR then 🚀

toter3:
done!

PR #2008, opened 2023-03-18
Don't set `self.opaque = false` on macOS

hsjoberg:
# Summary

Hi, guys. In recent react-native-macos versions, react-native-svg is broken.
The reason for this is because https://github.com/microsoft/react-native-macos/pull/1612 addressed some of the compiler warnings they were having, one of the being not redefining `opaque` for `NSView`.

Now I see that you have a couple of `#ifdefs` to deal with this already, but the check is not everywhere.
This PR fixes this.

## Test Plan

Test latest react-native-macos master with react-native-svg and it will complain that `opaque` is `readonly`.
With this PR, the issue is fixed.

Just using react-native-svg.


WoLewicki (approved):
LGTM :rocket: Thanks for your contribution!

Saadnajmi:
I was about to make this change... and saw it was made already, Thank you! Unexpected fallout from my RN-macOS change 😅

PR #2013, opened 2023-03-23
fix: remove unnecessary methods provided by superclasses for accessibility

WoLewicki:
PR removing methods that seem unnecessary for the accessibility props since they are provided by superclasses. It was necessary to update fabric `updateProps` method though, since we don't call `super` there. Also added proper type definitions and removed unnecessary casts.


WoLewicki:
@toter3 could you check if this PR still gives all the abilities that were provided earlier? Also @sseppola could you do it also? I just checked inside the tools available in the IDEs, but it would be great to check if it works correctly with `appium` too.

toter3:
sure, i'll check right after the weekend

toter3:
sure, i'll check right after the weekend

toter3:
great success!
image

toter3:
btw, do you know to say when can we expect a new version with the changes (onLayout and this / mine (Gal's) accessibility MR)?

Issue #1199, opened 2019-11-20
SvgUri Error. UnexpectedData &#10;&#9;&#9;


troublediehard:
# Bug
I'm using `SvgUri` like
```
<SvgUri
	width="60"
	height="60"
	uri="https://mydomain.com/images/logo.svg"
/>
```
and I'm getting following error
<img width="487" alt="Screen Shot 2019-11-20 at 6 04 10 PM" src="https://user-images.githubusercontent.com/6594232/69257851-65f0d900-0bc4-11ea-8ae0-458bafcb72e9.png">

there is element in SVG file
```
<polygon fill="#FF4422" points="37.16667,56.13386 47.16667,56.13386 47.16667,81.13386 62.16665,81.13386 62.16665,56.13386 &#10;&#9;&#9;72.16666,56.13386 72.16666,41.13386 37.16667,41.13386 &#9;"/>
```

and issue causes following escape format sequences `&#10;`, `&#9;`

I've tried to put this string directly to `Polygon` elements and it unescaped those sequences automatically and render just fine. Also, my SVG renders in a browser without any issue. 
```
<Polygon
  points="37.16667,56.13386 47.16667,56.13386 47.16667,81.13386 62.16665,81.13386 62.16665,56.13386 &#10;&#9;&#9;72.16666,56.13386 72.16666,41.13386 37.16667,41.13386 &#9;"
  fill="#FF4422"
/>
```
I see solution as add 
```
.replace(/&#.+?;/g, '')
```
to unescape fetched string. I think best place is 
```
// src/xml.tsx

function getAttributes(props: {
  [x: string]: Styles | string | number | boolean | undefined;
  style?: string | Styles | undefined;
}) {
  while (i < length) {
    
    // ...
    
    // ----- fix starts
    if (typeof value === "string") {
      value = String(value).replace(/&#.+?;/g, '');
    }
    // ----- fix ends

    props[camelCase(name)] = value;
  }
}
```

I can make PR. Please let me know what do you think.

troublediehard:
For those who are looking for a workaround, you can use the following 
```
import React, { useEffect, useState } from 'react';
import { SvgXml, UriProps } from 'react-native-svg';

async function fetchText(uri: string) {
  const response = await fetch(uri);
  const result = await response.text();
  return result.replace(/&#.+?;/g, '');
}

export function SvgUri(props: UriProps) {
  const { uri } = props;
  const [xml, setXml] = useState(null);
  useEffect(() => {
[...]

PR #2049, opened 2023-05-02
fix: svg with special characters leading to UnexpectedData

kibolho:
# Summary

Some SVG have especial characters that causes the following error: 
Screenshot 2023-05-02 at 20 26 34

https://github.com/software-mansion/react-native-svg/issues/1199

No

Add a replace to escape some special characters

Components which use fetchText to fetch the svg: SvgWithCssUri, LocalSvg SvgFromUri and SvgUri

## Test Plan

## Compatibility

| OS      | Implemented |
| ------- | :---------: |
| iOS     |    ✅❌     |
| Android |    ✅❌     |

## Checklist

### DIFF
--- a/src/xml.tsx
+++ b/src/xml.tsx
@@ -123,7 +123,8 @@ export function SvgXml(props: XmlProps) {
 export async function fetchText(uri: string) {
   const response = await fetch(uri);
   if (response.ok || (response.status === 0 && uri.startsWith('file://'))) {
-    return await response.text();
+    const result = await response.text();
+    return result.replace(/&#.+?;/g, '');
   }
   throw new Error(`Fetching ${uri} failed with status ${response.status}`);
 }


mschmid:
Hi @kibolho this fix could also help me with crashes in production, i am very happy you tackled this.

WoLewicki:
How do those special characters end up in the svg? And does the same code work correctly on web? If so, I think we can merge the PR then.

RyanHirsch-Phantom:
> How do those special characters end up in the svg? And does the same code work correctly on web? If so, I think we can merge the PR then.

@WoLewicki, I've just run into this same issue. We have an SVG with a CR/LF in it that after passing through a Cloudflare proxy to sanitize potentially malicious content gets the escape sequence injected rather than the original CR/LF. The web displays both images fine, but react-native-svg displays an empty (black) square for the original with CR/LF and crashes with the escape sequences. (I don't seem to be able to leverage error boundaries to gracefully handle the presence of the escape sequence)

example svgs: https://gist.github.com/RyanHirsch-Phantom/6a563603aefe1a3ca17955588159edfd

bohdanprog:
Hello @RyanHirsch-Phantom,
I've just tested the solution from that PR, but unfortunately, it hasn't solved a problem.

bohdanprog:
Hey @RyanHirsch-Phantom,
I've just checked that [gist.github.com/RyanHirsch-Phantom/6a563603aefe1a3ca17955588159edfd](https://gist.github.com/RyanHirsch-Phantom/6a563603aefe1a3ca17955588159edfd)  and everything worked well, I believed we can close PR and issue, but if you still faced some problems with that feel free to open a new issue.
Thank you.

PR #2052, opened 2023-05-10
[Windows] Port to Direct2D to remove win2d dependency

marlenecota:
# Summary

Explain the **motivation** for making this change: 
This change removes the win2d (Direct2D wrapper) dependency by using D2D directly. This removes the manual step of adding the win2d <PackageReference> to any new react-native-windows projects that want to use react-native-svg. It is also a stepping stone to an easier Fabric implementation for windows. 

## Test Plan

Ran the example app and verified the functionality is the same to the current windows implementation. 

## Compatibility

| OS      | Implemented |
| ------- | :---------: |
| iOS     |    ❌     |
| Android |    ❌     |
| Windows |    ✅     |

## Checklist


windows/RNSVG/GroupView.cpp
```diff
@@ -97,21 +96,40 @@ void GroupView::UpdateProperties(IJSValueReader const &reader, bool forceUpdate,
...
+    SvgRoot().Invalidate();
   }
 }
 
-void GroupView::CreateGeometry(UI::Xaml::CanvasControl const &canvas) {
-  auto const &resourceCreator{canvas.try_as<ICanvasResourceCreator>()};
-  std::vector<Geometry::CanvasGeometry> geometries;
-  for (auto const &child : Children()) {
+void GroupView::CreateGeometry() {
+  std::vector<ID2D1Geometry*> geometries;
```
chrisglein, line 104: Clarifiying question on how this works. Is this taking a ref for each item, and if it's not is that okay?

windows/RNSVG/ImageView.cpp
```diff
@@ -77,16 +78,33 @@ void ImageView::UpdateProperties(IJSValueReader const &reader, bool forceUpdate,
...
-void ImageView::Render(UI::Xaml::CanvasControl const &canvas, CanvasDrawingSession const &session) {
+void ImageView::Draw(IInspectable const &context, Size size) {
+  if (!m_wicbitmap) {
+    return;
+  }
+
+  com_ptr<ID2D1DeviceContext1> deviceContext;
+  copy_to_abi(context, *deviceContext.put_void());
+
+  uint32_t imgWidth, imgHeight = 0;
```
chrisglein, line 89: initializing one but not the other, also immediate written to

windows/RNSVG/LineView.cpp
```diff
@@ -30,18 +29,34 @@ void LineView::UpdateProperties(IJSValueReader const &reader, bool forceUpdate,
...
+  com_ptr<ID2D1Factory> factory;
+  deviceContext->GetFactory(factory.put());
+
+  com_ptr<ID2D1PathGeometry> geometry;
+  check_hresult(factory->CreatePathGeometry(geometry.put()));
+
+  com_ptr<ID2D1GeometrySink> sink;
+  check_hresult(geometry->Open(sink.put()));
+
+  // hollow instead of filled??
```
chrisglein, line 52: What's this comment about?

windows/RNSVG/LinearGradientView.cpp
```diff
@@ -49,33 +48,51 @@ void LinearGradientView::Unload() {
...
-void LinearGradientView::SetPoints(Brushes::CanvasLinearGradientBrush brush, Windows::Foundation::Rect const &bounds) {
-  float x1{Utils::GetAbsoluteLength(m_x1, bounds.Width) + bounds.X};
-  float y1{Utils::GetAbsoluteLength(m_y1, bounds.Height) + bounds.Y};
-  float x2{Utils::GetAbsoluteLength(m_x2, bounds.Width) + bounds.X};
-  float y2{Utils::GetAbsoluteLength(m_y2, bounds.Height) + bounds.Y};
+void LinearGradientView::SetPoints(ID2D1LinearGradientBrush * brush, D2D1_RECT_F bounds) {
+  float width{bounds.right - bounds.left};
+  float height{bounds.bottom - bounds.top};
+
+  float x1{Utils::GetAbsoluteLength(m_x1, width) + bounds.left}; // just right?
```
chrisglein, line 89: > just right?

?

windows/RNSVG/PatternView.cpp
```diff
@@ -55,77 +54,117 @@ void PatternView::UpdateProperties(IJSValueReader const &reader, bool forceUpdat
...
+  com_ptr<ID2D1CommandList> cmdList;
+  check_hresult(deviceContext->CreateCommandList(cmdList.put()));
 
-    if (m_transformSet) {
-      transform = transform * m_transform;
-    }
+  deviceContext->SetTarget(cmdList.get());
+
+  deviceContext->BeginDraw();
+  deviceContext->Clear(D2D1::ColorF(D2D1::ColorF::Orange, 0.0f));
```
chrisglein, line 128: Is the orange clear color for debugging purposes?

windows/RNSVG/PatternView.cpp
```diff
@@ -55,77 +54,117 @@ void PatternView::UpdateProperties(IJSValueReader const &reader, bool forceUpdat
...
+        (m_vbWidth + m_minX),
+        (m_vbHeight + m_minY)};
+
+    auto viewboxTransform{Utils::GetViewBoxTransform(
+        vbRect,
+        {
+            rect.left,
+            rect.top,
+            rect.left + rect.right,
+            rect.top + rect.bottom
```
chrisglein, line 145: All the different rect types are making my head spin. Some LTRB, some XYWH. And here... why would you need left+rigth / top+bottom?

windows/RNSVG/PatternView.cpp
```diff
@@ -55,77 +54,117 @@ void PatternView::UpdateProperties(IJSValueReader const &reader, bool forceUpdat
...
+        },
+        m_align,
+        m_meetOrSlice)};
+
+    transform = D2DHelpers::AsD2DTransform(viewboxTransform) * transform;
   }
 
+  deviceContext->SetTransform(transform);
+
+  Size size{rect.right - rect.left, rect.bottom - rect.top};
```
chrisglein, line 155: Great candidate for a standard transform function (Size from D2D1_RECT_F)

windows/RNSVG/RenderableView.cpp
```diff
@@ -96,40 +95,16 @@ void RenderableView::UpdateProperties(IJSValueReader const &reader, bool forceUp
...
-              break;
-            case 1:
-              m_strokeLineCap = Geometry::CanvasCapStyle::Round;
-              break;
-            case 0:
-            default:
-              m_strokeLineCap = Geometry::CanvasCapStyle::Flat;
-              break;
-          }
+          m_strokeLineCap = propertyValue.AsInt32();
```
chrisglein, line 98: Is it appropriate here to ensure the number value is within the expected valid range? Do the input validation here or later?

windows/RNSVG/RenderableView.cpp
```diff
@@ -345,7 +352,7 @@ Geometry::CanvasGeometry RenderableView::ClipPathGeometry() {
 
 void RenderableView::Unload() {
   if (m_geometry) {
-    m_geometry.Close();
+    //m_geometry.Close();
```
chrisglein, line 355: Delete?

windows/RNSVG/RenderableView.h
```diff
@@ -87,15 +83,13 @@ struct RenderableView : RenderableViewT<RenderableView> {
...
       winrt::single_threaded_vector<RNSVG::SVGLength>()};
-  Microsoft::Graphics::Canvas::Geometry::CanvasCapStyle m_strokeLineCap{
-      Microsoft::Graphics::Canvas::Geometry::CanvasCapStyle::Flat};
-  Microsoft::Graphics::Canvas::Geometry::CanvasLineJoin m_strokeLineJoin{
-      Microsoft::Graphics::Canvas::Geometry::CanvasLineJoin::Miter};
-  Microsoft::Graphics::Canvas::Geometry::CanvasFilledRegionDetermination m_fillRule{
-      Microsoft::Graphics::Canvas::Geometry::CanvasFilledRegionDetermination::Winding};
+  int32_t m_strokeLineCap{0};
+  int32_t m_strokeLineJoin{0};
+  int32_t m_fillRule{1};
```
chrisglein, line 92: Feels like you want to map these to a constant and not a magic value, especially since they're not all 0. Or at a bare minimum comment what the default is

windows/RNSVG/PatternView.cpp
chrisglein: Often these sort of things are done in debug builds but not in release builds in order to save on redundant rendering

windows/RNSVG/Utils.h
```diff
@@ -299,45 +229,97 @@ struct Utils {
...
+
     if (root && brushId != L"") {
       if (brushId == L"currentColor") {
-        brush = Brushes::CanvasSolidColorBrush(resourceCreator, root.CurrentColor());
+        try {
+          com_ptr<ID2D1SolidColorBrush> scb;
+          deviceContext->CreateSolidColorBrush(D2DHelpers::AsD2DColor(root.CurrentColor()), scb.put());
+          brush = scb.as<ID2D1Brush>();
+        } catch (hresult_error const &e) {
+          hresult hr = e.code();
```
chrisglein, line 282: Not doing anything with this hr

windows/RNSVG/Utils.h
```diff
@@ -299,45 +229,97 @@ struct Utils {
...
+
+          if (geometry) {
+            D2D1_RECT_F bounds;
+            geometry->GetBounds(nullptr, &bounds);
+            brushView.SetBounds(D2DHelpers::FromD2DRect(bounds));
+          }
+
+          copy_to_abi(brushView.Brush(), *brush.put_void());
+        } catch (hresult_error const &e) {
+          hresult hr = e.code();
```
chrisglein, line 296: Not doing anything with this hr

windows/RNSVG/Utils.h
```diff
@@ -299,45 +229,97 @@ struct Utils {
...
     }
 
     if (!brush) {
-      brush = Brushes::CanvasSolidColorBrush(resourceCreator, color);
+      try {
+      com_ptr<ID2D1SolidColorBrush> scb;
+      deviceContext->CreateSolidColorBrush(D2DHelpers::AsD2DColor(color), scb.put());
+      brush = scb.as<ID2D1Brush>();
+      } catch (hresult_error const &e) {
+        hresult hr = e.code();
```
chrisglein, line 307: Not doing anything with this hr

windows/RNSVG/Utils.h
```diff
@@ -299,45 +229,97 @@ struct Utils {
...
+        }
       }
     }
 
     if (!brush) {
-      brush = Brushes::CanvasSolidColorBrush(resourceCreator, color);
+      try {
+      com_ptr<ID2D1SolidColorBrush> scb;
+      deviceContext->CreateSolidColorBrush(D2DHelpers::AsD2DColor(color), scb.put());
+      brush = scb.as<ID2D1Brush>();
```
chrisglein, line 305:
```suggestion
        com_ptr<ID2D1SolidColorBrush> scb;
        deviceContext->CreateSolidColorBrush(D2DHelpers::AsD2DColor(color), scb.put());
        brush = scb.as<ID2D1Brush>();
```

windows/RNSVG/D2DHelpers.h
```diff
@@ -0,0 +1,209 @@
...
+    deviceContext->PushLayer(layerParams, opacityLayer.get());
+  }
+
+   static D2D1_CAP_STYLE GetLineCap(int32_t const lineCap) {
+     switch (lineCap) {
+       case 2:
+         return D2D1_CAP_STYLE_SQUARE;
+       case 1:
+         return D2D1_CAP_STYLE_ROUND;
+       case 0:
```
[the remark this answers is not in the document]
marlenecota: https://github.com/software-mansion/react-native-svg/blob/main/src/lib/extract/extractStroke.ts#L12-L16

windows/RNSVG/CircleView.cpp
```diff
@@ -29,13 +28,25 @@ void CircleView::UpdateProperties(IJSValueReader const &reader, bool forceUpdate
...
 
-  Geometry(Geometry::CanvasGeometry::CreateCircle(resourceCreator, cx, cy, r));
+  com_ptr<ID2D1Factory> factory;
+  deviceContext->GetFactory(factory.put());
+
+  com_ptr<ID2D1EllipseGeometry> geometry;
+  check_hresult(factory->CreateEllipseGeometry(D2D1::Ellipse({cx, cy}, r, r), geometry.put()));
+
+  IInspectable asInspectable;
+  copy_from_abi(asInspectable, geometry.get());
```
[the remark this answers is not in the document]
asklar: it's because these types are not WinRT types, so they won't implement IInspectable.
copy_from_abi is just a reinterpret cast, so there's no safety. This operation is in fact undefined behavior :) so instead recommend making m_geometry something other than IInspectable. Either the right base [COM] interface, or IUnknown.

windows/RNSVG/D2DHelpers.h
```diff
@@ -0,0 +1,209 @@
...
+
+    if (clipPathGeometry) {
+      layerParams.geometricMask = clipPathGeometry;
+    }
+
+    deviceContext->PushLayer(layerParams, opacityLayer.get());
+  }
+
+   static D2D1_CAP_STYLE GetLineCap(int32_t const lineCap) {
+     switch (lineCap) {
```
chrisglein, line 48:
```suggestion
     // Enum values defined here: https://github.com/software-mansion/react-native-svg/blob/main/src/lib/extract/extractStroke.ts#L6-L10
     switch (lineCap) {
```

windows/RNSVG/D2DHelpers.h
```diff
@@ -0,0 +1,209 @@
...
+       case 1:
+         return D2D1_CAP_STYLE_ROUND;
+       case 0:
+       default:
+         return D2D1_CAP_STYLE_FLAT;
+     }
+   }
+
+   static D2D1_LINE_JOIN GetLineJoin(int32_t const lineJoin) {
+     switch (lineJoin) {
```
chrisglein, line 60:
```suggestion
     // Enum values defined here: https://github.com/software-mansion/react-native-svg/blob/main/src/lib/extract/extractStroke.ts#L12-L16
     switch (lineJoin) {
```

windows/RNSVG/D2DHelpers.h
```diff
@@ -0,0 +1,209 @@
...
+       case 1:
+         return D2D1_LINE_JOIN_ROUND;
+       case 0:
+       default:
+         return D2D1_LINE_JOIN_MITER;
+     }
+   }
+
+   static D2D1_FILL_MODE GetFillRule(int32_t const fillRule) {
+     switch (fillRule) {
```
chrisglein, line 72: Find and add a link to the matching source for this enum
```
     // Enum values defined here: https://github.com/software-mansion/react-native-svg/blob/main/src/<ACTUAL LINK HERE>
     switch (fillRule) {
```

windows/RNSVG/GroupView.cpp
chrisglein: Looks like this still needs addressing. When we looked at it last you had a matching code comment you were going to duplicate here?

windows/RNSVG/GroupView.cpp
```diff
@@ -130,36 +143,43 @@ void GroupView::MergeProperties(RNSVG::RenderableView const &other) {
...
+
+  D2D1_MATRIX_3X2_F transform{D2DHelpers::GetTransform(deviceContext.get())};
 
   if (m_propSetMap[RNSVG::BaseProp::Matrix]) {
-    session.Transform(transform * SvgTransform());
+    deviceContext->SetTransform(D2DHelpers::AsD2DTransform(SvgTransform()) * transform);
   }
 
-  auto const &clipPathGeometry{ClipPathGeometry()};
+  com_ptr<ID2D1Geometry> clipPathGeometry{nullptr};
```
chrisglein, line 155: Does it not already initialize to nullptr?

windows/RNSVG/ImageView.cpp
```diff
@@ -260,10 +284,57 @@ IAsyncOperation<InMemoryRandomAccessStream> ImageView::GetImageInlineDataAsync(I
...
 
   co_return nullptr;
 }
+
+com_ptr<IWICBitmapSource> ImageView::wicBitmapSourceFromStream(InMemoryRandomAccessStream const &results) {
+  if (!results) {
+    return nullptr;
+  }
+
+  com_ptr<IWICImagingFactory> imagingFactory{nullptr};
```
chrisglein, line 299: Doesn't need nullptr init? (seeing this happen inconsistently, wondering if there's some reason why)

windows/RNSVG/PatternView.cpp
```diff
@@ -55,77 +55,112 @@ void PatternView::UpdateProperties(IJSValueReader const &reader, bool forceUpdat
...
-    brush.ExtendX(Microsoft::Graphics::Canvas::CanvasEdgeBehavior::Wrap);
-    brush.ExtendY(Microsoft::Graphics::Canvas::CanvasEdgeBehavior::Wrap);
+    com_ptr<ID2D1DeviceContext> deviceContext{get_self<D2DDeviceContext>(root.DeviceContext())->Get()};
 
-    cmdList.Close();
+    com_ptr<ID2D1ImageBrush> imageBrush;
+    check_hresult(deviceContext->CreateImageBrush(cmdList.get(), brushProperties, imageBrush.put()));
+
+    auto transform{D2D1::Matrix3x2F::Translation({rect.left, rect.top})};
+    if (m_transformSet) {
```
chrisglein, line 94: if `m_transformSet` used to essentially treat `m_transform` as a nullable type? Is `std::optional` possible instead?

windows/RNSVG/RadialGradientView.cpp
```diff
@@ -53,40 +52,52 @@ void RadialGradientView::Unload() {
...
-  Brushes::CanvasRadialGradientBrush brush{resourceCreator, m_stops};
+  auto const root{SvgRoot()};
 
-  SetPoints(brush, {0, 0, canvas.Size().Width, canvas.Size().Height});
+  com_ptr<ID2D1DeviceContext> deviceContext{get_self<D2DDeviceContext>(root.DeviceContext())->Get()};
+
+  winrt::com_ptr<ID2D1GradientStopCollection> stopCollection;
+  winrt::check_hresult(deviceContext->CreateGradientStopCollection(&m_stops[0], static_cast<uint32_t>(m_stops.size()), stopCollection.put()));
+
+  D2D1_RADIAL_GRADIENT_BRUSH_PROPERTIES brushProperties;
```
chrisglein, line 62: Is this struct being passed in completely uninitialized?

windows/RNSVG/RenderableView.cpp
```diff
@@ -358,20 +364,29 @@ void RenderableView::Unload() {
 
 RNSVG::IRenderable RenderableView::HitTest(Point const &point) {
   if (m_geometry) {
-    bool strokeContainsPoint{false};
+    BOOL strokeContainsPoint{false};
```
chrisglein, line 367: I guess technically both of these `BOOL` types should init to `FALSE` instead of `false`, but 🤢

windows/RNSVG/RenderableView.h
```diff
@@ -33,22 +35,18 @@ struct RenderableView : RenderableViewT<RenderableView> {
   float StrokeDashOffset() { return m_strokeDashOffset; }
   RNSVG::SVGLength StrokeWidth() { return m_strokeWidth; }
   Windows::Foundation::Collections::IVector<RNSVG::SVGLength> StrokeDashArray() { return m_strokeDashArray; }
-  Microsoft::Graphics::Canvas::Geometry::CanvasCapStyle StrokeLineCap() { return m_strokeLineCap; }
-  Microsoft::Graphics::Canvas::Geometry::CanvasLineJoin StrokeLineJoin() { return m_strokeLineJoin; }
-  Microsoft::Graphics::Canvas::Geometry::CanvasFilledRegionDetermination FillRule() { return m_fillRule; }
-  Microsoft::Graphics::Canvas::Geometry::CanvasGeometry ClipPathGeometry();
+  int32_t StrokeLineCap() { return m_strokeLineCap; }
+  int32_t StrokeLineJoin() { return m_strokeLineJoin; }
+  int32_t FillRule() { return m_fillRule; }
```
chrisglein, line 40: What was the reason these should be ints here instead of their mapped enum values?

windows/RNSVG/RenderableView.h
```diff
@@ -87,15 +85,13 @@ struct RenderableView : RenderableViewT<RenderableView> {
...
       winrt::single_threaded_vector<RNSVG::SVGLength>()};
-  Microsoft::Graphics::Canvas::Geometry::CanvasCapStyle m_strokeLineCap{
-      Microsoft::Graphics::Canvas::Geometry::CanvasCapStyle::Flat};
-  Microsoft::Graphics::Canvas::Geometry::CanvasLineJoin m_strokeLineJoin{
-      Microsoft::Graphics::Canvas::Geometry::CanvasLineJoin::Miter};
-  Microsoft::Graphics::Canvas::Geometry::CanvasFilledRegionDetermination m_fillRule{
-      Microsoft::Graphics::Canvas::Geometry::CanvasFilledRegionDetermination::Winding};
+  int32_t m_strokeLineCap{0};
+  int32_t m_strokeLineJoin{0};
+  int32_t m_fillRule{1};
```
chrisglein, line 94: These non-typed defaults feel like they need a comment at _least_.

windows/RNSVG/SvgView.cpp
```diff
@@ -5,36 +5,66 @@
...
+  // Create the Direct3D device.
+  com_ptr<ID3D11Device> d3dDevice;
+  D3D_FEATURE_LEVEL supportedFeatureLevel;
+  check_hresult(D3D11CreateDevice(
+      nullptr, // default adapter
+      D3D_DRIVER_TYPE_HARDWARE,
+      0,
+      creationFlags,
+      featureLevels,
+      ARRAYSIZE(featureLevels),
```
chrisglein, line 44:
```suggestion
      sizeof(featureLevels),
```

windows/RNSVG/SvgView.cpp
```diff
@@ -189,16 +221,54 @@ void SvgView::Unload() {
...
+      static_cast<int32_t>(size.Width), static_cast<int32_t>(size.Height));
+  com_ptr<ISurfaceImageSourceNativeWithD2D> sisNativeWithD2D{surfaceImageSource.as<ISurfaceImageSourceNativeWithD2D>()};
+
+  // Associate the Direct2D device with the SurfaceImageSource.
+  com_ptr<ID2D1Device> device{get_self<D2DDevice>(Device())->Get()};
+  sisNativeWithD2D->SetDevice(device.get());
+
+  com_ptr<IDXGISurface> dxgiSurface;
+
+  RECT updateRect{0, 0, static_cast<long>(size.Width), static_cast<long>(size.Height)};
```
chrisglein, line 248: Oh the hilarity of yet another rectangle type in the same codebase. I mean, you're constrained by the APIs you're calling, so 🤷‍♂️.
This one had me do a double take because there might have been an error, given that `RECT` is LTRB not LTWH. But... since it's 0 offset it's the same.
For the record, that kind of mistake is why I think it's good to have explicit conversion functions for these. Example `RECTFromSize()`. There are so many casts and conversions in this code and so far I think they're right but also this stuff is error prone.

windows/RNSVG/SvgView.cpp
```diff
@@ -189,16 +221,54 @@ void SvgView::Unload() {
...
+  // Set context's render target.
+  deviceContext->SetTarget(bitmap.get());
+
+  // Draw using Direct2D context
+  deviceContext->BeginDraw();
+
+  auto transform = Numerics::make_float3x2_translation({static_cast<float>(offset.x), static_cast<float>(offset.y)});
+  deviceContext->SetTransform(D2DHelpers::AsD2DTransform(transform));
+
+  deviceContext->Clear(D2D1::ColorF(D2D1::ColorF::Orange, 0.0f));
```
chrisglein, line 265: Clear color is for debug purposes, right? Should this be removed?

windows/RNSVG/PatternView.cpp
marlenecota: using the identity matrix as default instead

windows/RNSVG/ImageView.cpp
marlenecota: no reason. I'm just inconsistent 🙃

windows/RNSVG/RadialGradientView.cpp
marlenecota: those values are set in SetPoints below, but yes. I've now changed it to initialize with default values.

windows/RNSVG/SvgView.cpp
marlenecota: I got this from here: https://learn.microsoft.com/en-us/windows/uwp/gaming/directx-and-xaml-interop#surfaceimagesource

I tried sizeof and it crashed, so I will leave ARRAYSIZE.

windows/RNSVG/SvgView.cpp
marlenecota: Orange was for debug purposes. I'll change it to the more usual Black (though it's transparent), but it does need to clear the devicecontext before drawing.

windows/RNSVG/SvgView.cpp
asklar: ARRAYSIZE is correct (number of elements).
sizeof is number of bytes, which is going to be # of elements * sizeof(each element), so you'll end up reading e.g. 7 * 4 entries instead of 7.

A better alternative to ARRAYSIZE (which is a macro, and non-standard) is std::size (still number of elements, doesn't rely on macros)

chrisglein: Sorry, I mean '_countof'. Brain fart.

windows/RNSVG/BrushView.cpp
```diff
@@ -10,14 +10,13 @@ void BrushView::SaveDefinition() {
...
-void BrushView::SetBounds(Windows::Foundation::Rect const &rect) {
-  m_bounds = rect;
+void BrushView::SetBounds(Rect const &rect) {
+  m_bounds = {rect.X, rect.Y, rect.Width + rect.X, rect.Height + rect.Y};
   UpdateBounds();
 }
 
 void BrushView::Unload() {
   if (m_brush) {
-    m_brush.Close();
```
TatianaKapos, line 20: Just curious, why did we close the brush before if we set it to nullptr in the next line?

windows/RNSVG/ImageView.cpp
```diff
@@ -96,52 +113,61 @@ void ImageView::Render(UI::Xaml::CanvasControl const &canvas, CanvasDrawingSessi
...
 
-      session.Transform(transform);
+    if (m_align != "") {
+      com_ptr<ID2D1Effect> bitmapEffects;
+      check_hresult(deviceContext->CreateEffect(CLSID_D2D1BitmapSource, bitmapEffects.put()));
+      check_hresult(bitmapEffects->SetValue(D2D1_BITMAPSOURCE_PROP_WIC_BITMAP_SOURCE, m_wicbitmap.get()));
+
+      com_ptr<ID2D1Effect> transformEffect;
+      Rect elRect{x, y, width, height};
+      Rect vbRect{0, 0, m_source.width, m_source.height};
```
TatianaKapos, line 137: What are elRect and vbRext standing for?

windows/RNSVG/BrushView.cpp
asklar: m_brush used to be an ICanvasBrush, now it's something else.
The ICanvasBrush interface implements IDisposable (WinRT IClosable), which requires calling Close before releasing the last reference:

- https://microsoft.github.io/Win2D/WinUI3/html/T_Microsoft_Graphics_Canvas_Brushes_ICanvasBrush.htm

- https://learn.microsoft.com/en-us/uwp/api/windows.foundation.iclosable?view=winrt-22621

windows/RNSVG/ImageView.cpp
marlenecota: elRect is for the element and vbRect for the viewbox. 

You can read more about viewbox transforms here:
https://svgwg.org/svg2-draft/coords.html#ComputingAViewportsTransform

windows/RNSVG/LineView.cpp
marlenecota: It was a self-note to double check if I wanted to use D2D1_FIGURE_BEGIN_HOLLOW instead of D2D1_FIGURE_BEGIN_FILLED below. I did not :)

windows/RNSVG/PatternView.cpp
marlenecota: The choice of orange was for debugging reasons (I can change it to Black), but the clear call to transparent seems to be a best practice before you draw.

windows/RNSVG/RenderableView.cpp
marlenecota: It's an enum. If it's out of range, the converter will use the default.

TatianaKapos (approved):

windows/RNSVG/RadialGradientView.cpp
```diff
@@ -33,11 +32,10 @@ void RadialGradientView::UpdateProperties(IJSValueReader const &reader, bool for
     } else if (propertyName == "gradientUnits") {
       m_gradientUnits = Utils::JSValueAsBrushUnits(propertyValue);
     } else if (propertyName == "gradientTransform") {
-      m_transformSet = true;
-      m_transform = Utils::JSValueAsTransform(propertyValue);
+      m_transform = Utils::JSValueAsD2DTransform(propertyValue);
 
       if (propertyValue.IsNull()) {
-        m_transformSet = false;
+       m_transform = D2D1::Matrix3x2F::Identity();
```
jonthysell, line 38: Performance nit? I've seen use this elsewhere - does this mean a transform is always applied?

windows/RNSVG/TextView.cpp
```diff
@@ -45,17 +46,20 @@ void TextView::UpdateProperties(IJSValueReader const &reader, bool forceUpdate,
...
+  com_ptr<ID2D1DeviceContext> deviceContext{get_self<D2DDeviceContext>(context)->Get()};
+
+  D2D1_MATRIX_3X2_F transform{D2DHelpers::GetTransform(deviceContext.get())};
+
   bool translateXY{X().Size() > 0 || Y().Size() > 0};
   if (translateXY) {
     float x{X().Size() > 0 ? X().GetAt(0).Value() : 0};
     float y{Y().Size() > 0 ? Y().GetAt(0).Value() : 0};
-    session.Transform(transform * Numerics::make_float3x2_translation(x, y));
+    deviceContext->SetTransform(D2D1::Matrix3x2F::Translation({x,y}) * transform);
```
jonthysell, line 58: Isn't matrix math not commutative? Shouldn't this be:

```suggestion
    deviceContext->SetTransform(transform * D2D1::Matrix3x2F::Translation({x,y}));
```

jonthysell (approved):
Looks good to me, as long as you double-check the transform map. Is there windows build CI / tests that we should be running?

windows/RNSVG/TextView.cpp
marlenecota: Correct that matrix multiplication is not commutative, but this is the correct order.

marlenecota:
> Looks good to me, as long as you double-check the transform map. Is there windows build CI / tests that we should be running?

No window's specific CI/tests yet, but it is on my list of things to add. 

@WoLewicki - any reservations on this PR?

WoLewicki:
I can merge and release it if you think it is ready :rocket:

marlenecota:
@WoLewicki please merge! As far as bumping to 73, I don't expect any issues, but I am happy to jump in and fix if anything pops up.

PR #2058, opened 2023-05-20
Fix setNativeProps type

fauri13:
# Summary

Setting Object to the props type does not allow passing any object to the method. So in TypeScript when using the method for example with Path, you will get an error when passing the d:

```typescript
this.pathComponent?.setNativeProps({ d })
```

```typescript
Argument of type '{ d: string; }' is not assignable to parameter of type 'Object & { matrix?: ColumnMajorTransformMatrix | undefined; fill?: ColorValue | undefined; } & TransformProps'.
  Object literal may only specify known properties, and 'd' does not exist in type 'Object & { matrix?: ColumnMajorTransformMatrix | undefined; fill?: ColorValue | undefined; } & TransformProps'
.ts(2345)
```

Instead ob Object, you should use at least P so it allows the original props of the component you are using.

## Test Plan

Only Type checking is modified, so it does not affect the functionality.

## Compatibility

Not aplicable

## Checklist

- [x ] I updated the typed files (typescript)

### DIFF
--- a/package.json
+++ b/package.json
@@ -1,5 +1,5 @@
 {
-  "version": "13.8.0",
+  "version": "13.9.0",
   "name": "react-native-svg",
   "description": "SVG library for react-native",
   "homepage": "https://github.com/react-native-community/react-native-svg",
--- a/src/elements/G.tsx
+++ b/src/elements/G.tsx
@@ -7,7 +7,6 @@ import type {
   CommonPathProps,
   FontProps,
   NumberProp,
-  TransformProps,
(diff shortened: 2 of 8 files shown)


WoLewicki:
@fauri13 it seems that the job is failing. Could you take a look at it locally and resolve it?

fauri13:
> @fauri13 it seems that the job is failing. Could you take a look at it locally and resolve it?

I've changed the other types as well, now it's compiling.

Issue #2077, opened 2023-06-30
[RN 0.72.1 New Arch] iOS build error: No matching constructor for initialization of 'facebook::react::RNSVGImageState'


yungblud:
# Bug
When I try to build iOS with xcode this error occurs
`No matching constructor for initialization of 'facebook::react::RNSVGImageState'`
![스크린샷 2023-06-30 오후 9 57 40](https://github.com/software-mansion/react-native-svg/assets/36250002/f39e8bf6-c487-4d6a-b0d0-40de23022ada)

#### Unexpected behavior
iOS XCode Build Error

## Environment info

System:
  OS: macOS 13.0
  CPU: (10) arm64 Apple M1 Pro
  Memory: 72.11 MB / 16.00 GB
  Shell:
    version: 5.8.1
    path: /bin/zsh
Binaries:
  Node:
    version: 18.12.1
    path: /var/folders/jf/pnk__5h55vqcsjtj_jchcz780000gn/T/yarn--1688129795348-0.059536718565355695/node
  Yarn:
    version: 1.22.19
    path: /var/folders/jf/pnk__5h55vqcsjtj_jchcz780000gn/T/yarn--1688129795348-0.059536718565355695/yarn
  npm:
    version: 8.19.2
    path: ~/.nvm/versions/node/v18.12.1/bin/npm
  Watchman:
    version: 2023.04.24.00
    path: /opt/homebrew/bin/watchman
Managers:
  CocoaPods:
    version: 1.12.0
    path: /Users/paul/.rvm/gems/ruby-2.7.6/bin/pod
SDKs:
  iOS SDK:
    Platforms:
      - DriverKit 22.1
      - iOS 16.1
      - macOS 13.0
      - tvOS 16.1
      - watchOS 9.1
  Android SDK:
    API Levels:
      - "23"
      - "24"
      - "28"
      - "30"
      - "31"
      - "32"
      - "33"
    Build Tools:
      - 29.0.2
      - 30.0.2
      - 30.0.3
      - 31.0.0
      - 32.0.0
      - 32.1.0
      - 33.0.0
      - 34.0.0
    System Images:
      - android-23 | Google APIs ARM 64 v8a
      - android-24 | Google APIs ARM 64 v8a
      - android-29 | Google Play ARM 64 v8a
      - android-31 | Google Play ARM 64 v8a
      - android-32 | Google APIs ARM 64 v8a
      - android-32 | Google Play ARM 64 v8a
[...]

PR #2079, opened 2023-07-01
Fix syntax error in RNSVGImageShadowNode::initialStateData

hsjoberg:
# Summary

This PR solves #2077.
I'm too bad at C++ to know whether this is idiomatic code, however.

## Checklist

### DIFF
--- a/common/cpp/react/renderer/components/rnsvg/RNSVGImageShadowNode.h
+++ b/common/cpp/react/renderer/components/rnsvg/RNSVGImageShadowNode.h
@@ -48,7 +48,7 @@ class JSI_EXPORT RNSVGImageShadowNode final : public ConcreteViewShadowNode<
       ShadowNodeFamilyFragment const &familyFragment,
       ComponentDescriptor const &componentDescriptor) {
     auto imageSource = ImageSource{ImageSource::Type::Invalid};
-    return {imageSource, {imageSource, nullptr}};
+    return {imageSource, {imageSource, nullptr, {}}};
   }
 
 #pragma mark - LayoutableShadowNode


common/cpp/react/renderer/components/rnsvg/RNSVGImageShadowNode.h
```diff
@@ -48,7 +48,7 @@ class JSI_EXPORT RNSVGImageShadowNode final : public ConcreteViewShadowNode<
       ShadowNodeFamilyFragment const &familyFragment,
       ComponentDescriptor const &componentDescriptor) {
     auto imageSource = ImageSource{ImageSource::Type::Invalid};
-    return {imageSource, {imageSource, nullptr}};
+    return {imageSource, {imageSource, nullptr, SharedFunction<>(nullptr)}};
```
WoLewicki, line 51: Adding this code will result in RN `0.71` not working on new arch, right? Also, is such code enough:

```suggestion
    return {imageSource, {imageSource, nullptr, {}}};
```

?

WoLewicki (commented):
Could you answer the comment? Besides this, the PR looks correct, thanks!

common/cpp/react/renderer/components/rnsvg/RNSVGImageShadowNode.h
Yuniac: Hi, I tested with the suggested change and it was enough to solve the issue (I ran into it today when trying the new arch). As to whether it will affect RN `0.71` or not I don't know 🙂

hsjoberg: On my way testing RN 0.71.

Can also confirm `{}` works (on 0.72)

hsjoberg: @WoLewicki You're correct on both things

RN 0.71 fails as the constructor is different

https://github.com/facebook/react-native/blob/49d16d5aefd703ff20ab5d05544d78d6ce446427/ReactCommon/react/renderer/imagemanager/ImageRequest.h#L26-L33
```CPP
class ImageRequest final {
 public:
  /*
   * The default constructor
   */
  ImageRequest(
      ImageSource imageSource,
      std::shared_ptr<const ImageTelemetry> telemetry);
```

I suppose things could be addressed by using the processor to check for react-native version.
I'm not entirely sure how to do that though.

WoLewicki: I think we will just drop the support for 0.71 then. We keep only the newest version working on new arch until it becomes more stable.

hsjoberg: Sure yeah that's all fine by me. It looks like there is a lot of trouble < 0.72 across the ecosystem anyway.

hsjoberg:
Alright, I've changed to `{}` instead.

khanakia:
Can we merge it soon please? I am unable to upgrade my react-native  to 0.72.1 version because of this issue

hsjoberg:
@khanakia If you need this fix immediately you can change your package.json like this, which will point to the [commit](https://github.com/hsjoberg/react-native-svg/commit/7ac1a17c4bce907be46c349f345a29b439c5e3dd)  in this PR:

```JSON
"react-native-svg": "git+https://github.com/hsjoberg/react-native-svg.git#7ac1a17c4bce907be46c349f345a29b439c5e3dd",
```

But don't forget to change back to an official react-native-svg version once this one gets merged and published.

khanakia:
Thanks

WoLewicki (approved):
Thanks for your contribution :rocket:

Happy-2002:
Please remove {} in this line -- return {imageSource, {imageSource, nullptr,{}}};

PR #2117, opened 2023-08-16
feat: remove UIGraphicsBeginImageContextWithOptions from repo

WoLewicki:
# Summary

Since `UIGraphicsBeginImageContextWithOptions` will be depracated in `iOS 17` (https://developer.apple.com/documentation/uikit/1623912-uigraphicsbeginimagecontextwitho), I changed the implementation to not use it and use [UIGraphicsImageRenderer](https://developer.apple.com/documentation/uikit/uigraphicsimagerenderer) instead.

Also added `Mask` examples to be able to test it.

## Test Plan

`Mask` examples and `Svg` example which for some reason does not show the rendered `base64` image now.

### DIFF
--- a/Example/src/App.tsx
+++ b/Example/src/App.tsx
@@ -115,6 +115,7 @@ const names: (keyof typeof examples)[] = [
   'Reanimated',
   'Transforms',
   'Markers',
+  'Mask',
 ];
 
 const initialState = {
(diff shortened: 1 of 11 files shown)


WoLewicki:
Hmm I think you are right, the library still points that it requires `iOS` 9+, so I'll bump that, thanks.

PR #2132, opened 2023-09-12
feat: SvgFromUri/SvgFromXml add props onError execution

H2-won:
# Summary

The `onError` function of the `SvgFromUri` / `SvgFromXml` component does not work.
According to the current code, there is no way to track if an error occurs in the component rendering.

`uri` fetch fails.

Pass the `onError` function as props to test whether it is executed.

## Compatibility

| OS      | Implemented |
| ------- | :---------: |
| iOS     |    ✅     |
| Android |    ✅❌     |

## Checklist

### DIFF
--- a/src/xml.tsx
+++ b/src/xml.tsx
@@ -174,6 +174,7 @@ export class SvgFromXml extends Component<XmlProps, XmlState> {
     try {
       this.setState({ ast: xml ? parse(xml) : null });
     } catch (e) {
+      this.props.onError?.(e);
       console.error(e);
     }
   }
@@ -204,6 +205,7 @@ export class SvgFromUri extends Component<UriProps, UriState> {
     try {
       this.setState({ xml: uri ? await fetchText(uri) : null });
     } catch (e) {
+      this.props.onError?.(e);
       console.error(e);
     }
   }


src/xml.tsx
```diff
@@ -174,6 +174,7 @@ export class SvgFromXml extends Component<XmlProps, XmlState> {
     try {
       this.setState({ ast: xml ? parse(xml) : null });
     } catch (e) {
+      this.props.onError?.(e);
```
WoLewicki, line 177: Nice catch! Shouldn't we use the same logic as introduced in https://github.com/software-mansion/react-native-svg/commit/3c32a6fdd580dea53ecc271288155117ce040b55, so we use `console.error(e);` as the default value of `onError` if other is not provided?

H2-won: `console.error(e);` does not necessarily need to operate by default.

```suggestion
    this.props.onError ? this.props.onError(e) : console.error(e);
```

Is it correct to say this way? That seems like a good approach!

src/xml.tsx
```diff
@@ -174,6 +174,7 @@ export class SvgFromXml extends Component<XmlProps, XmlState> {
...
       console.error(e);
```
H2-won, line 178: @WoLewicki `console.error(e);` does not necessarily need to operate by default.

```suggestion
    this.props.onError ? this.props.onError(e) : console.error(e);
```

Is it correct to say this way? That seems like a good approach!

WoLewicki: Isn't it the same as the one I mentioned: https://github.com/software-mansion/react-native-svg/commit/3c32a6fdd580dea53ecc271288155117ce040b55#diff-516de011a591b7585d24e2a1ee084a41d09bfc6c9ba70b1320e7e39bebd792f7R131 ? You either have your own implementation or use the `console.error`. Or am I missing something?

H2-won: @WoLewicki 
The commits you mentioned are for `SvgXml` and `SvgUri` components.
The parts I modified are `SvgFromUri` and `SvgFromXml`, so they are different.
And apparently `OnError` doesn't work when I use `SvgFromUri`.

H2-won:
> Could you resolve the failing CI job? It should be ready for merge then 🚀

@WoLewicki 

I am very sorry for the late reply.

I expect TypeScript `compilerOptions` to be stable.
Perhaps the modified code will pass the CI job safely.

Please check new commit.
Also, I hope you compare the new commit with the code below and choose a better solution.

`this.props.onError ? this.props.onError(e as Error) : console.error(e);`

PR #2141, opened 2023-09-21
Build `renderer/components/rnsvg` with `-std=c++20`

NickGerleman:
# Summary

RN 0.73 adds the requirement that ComponentViews, ShadowNodws, etc, build using C++ 20. This adds the flags to do that when building for Android.

## Test Plan

Locally validated this does not prevent us from building earlier RN Fabric (at least 0.72), by changing the flags in CMakeLists.txt, then running `./gradlew assembleDebug` in `FabricExample`.

I think RN on iOS will set C++ version for you [during pod install](https://github.com/facebook/react-native/commit/ca8174e15f77cbeecb7ff7a5a583abb668817777).

## Checklist

### DIFF
--- a/RNSVG.podspec
+++ b/RNSVG.podspec
@@ -20,8 +20,6 @@ Pod::Spec.new do |s|
 
   if fabric_enabled
     s.platforms = { :osx => "10.14", ios: '12.4', tvos: '11.0' }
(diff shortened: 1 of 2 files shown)


WoLewicki (approved):
LGTM :rocket:

PR #2143, opened 2023-09-21
enable buildConfig generation to better support AGP8

pvegh:
So that consumer projects don't have to globally enable this themselves leading to performance impact

# Summary

Since AGP8 this flag is disabled by default. Consumer projects have to enable this flag in them **globally** if the library doesn't do this correctly itself, but uses custom buildConfig fields.

Solves Build Errors with default setup using AGP8+
No errors and better build performance than the workaround in consuming projects would be.

## Test Plan

## Compatibility

| OS      | Implemented |
| ------- | :---------: |
| iOS     |    N/A     |
| Android |    ✅     |

## Checklist

- [ x ] I have tested this on a device and a simulator

### DIFF
--- a/android/build.gradle
+++ b/android/build.gradle
@@ -43,6 +43,11 @@ android {
     if (agpVersion.tokenize('.')[0].toInteger() >= 7) {
         namespace "com.horcrux.svg"
     }
+    if (agpVersion.tokenize('.')[0].toInteger() >= 8) {
+        buildFeatures {
+            buildConfig = true
+        }
+    }
     // Used to override the NDK path/version on internal CI or by allowing
     // users to customize the NDK path/version from their root project (e.g. for M1 support)
     if (rootProject.hasProperty("ndkPath")) {


pvegh:
See the same patch in https://github.com/software-mansion/react-native-gesture-handler/issues/2571 which is already merged.

pvegh:
Can we get a new version with this please, so we can remove the global BuildConfig generation and save some build time?

WoLewicki (approved):
LGTM :rocket: Thanks for your contribution!

mikehardy:
@pvegh thanks for proposing this! I was about to propose the same thing hopefully it is released soon

One question - I notice you wrapped this check in an AGP8+ conditional, any reason for that? Should be supported fine on AGP7+ and similar ones I've proposed do in fact build fine: https://github.com/th3rdwave/react-native-safe-area-context/pull/455

That said, nothing wrong with it technically

One note though, you mention this:

> Can we get a new version with this please, so we can remove the global BuildConfig generation and save some build time?

In my experience at least I needed it at the top-level *and* module level for it to work on AGP8+, where ever buildConfig was used

Cheers

pvegh:
Yeah check for 7+ should work fine. I didn't see the necessity/advantage since only AGP8+ forces us to enable the flag for the whole project in cases like this, when there is at least one module not defining buildConfig for itself.
From 8 on the BuildConfig generation is off by default.

mikehardy:
In my experience, my project wouldn't compile with AGP8 even *with* the global flag, unless the specific modules that needed buildConfig=true had it as well. I needed both, not one or the other 🤔 
Glad I understand the compatibility constraints though (AGP7 vs 8) - I've made PRs to a few repos with this and normally slipstream it in on their existing AGP7 conditional for compatibility

pvegh:
Should work if you put it in the project gradle.properties, but maybe I don't remember every workaround I made for it. AGP8 was a hard nut for sure.

PR #2167, opened 2023-10-26
Deduplicated fabric elements types

michalmaka:
# Summary

I found out that many interfaces in fabric examples are identical so I've deduplicated them and code looks cleaner.

## Test Plan

None

## Compatibility

None needed

## Checklist

None checked


WoLewicki:
If only codegen could import types from other files :(

PR #2177, opened 2023-11-24
fix: apple bounding box (CGPathGetPathBoundingBox)

omeid:
# Summary

As per the spec, bbox should not include any transformations, including scaling. It should also not include any control points. This fixes getBBox to give correct results matching Google Chrome, Firefox as per the spec.

Bug fix.

Like this, see.

getBBox API.

## Test Plan

I build some stuff that works with SVG, the results on web were inconsistent, and digging down I realise it was a react-native-svg.

Some fun fact, there was a surprisingly similar bug related to very confusing named (CGPathGetBoundingBox vs CGPathGetPathBoundingBox) APIs in Firefox some years ago as well.
https://bugzilla.mozilla.org/show_bug.cgi?id=1369904

## Compatibility

| OS      | Implemented |
| ------- | :---------: |
| iOS     |    ✅     |
| Android |   🔘#   |
| Web | 🔘 *|

* Depends on the host platform. But works fine in major browsers.
# will create a follow up PR. Currently getBBox on Android doesn't account for scaling properly.

## Checklist

### DIFF
--- a/apple/Elements/RNSVGForeignObject.mm
+++ b/apple/Elements/RNSVGForeignObject.mm
@@ -150,8 +150,8 @@ using namespace facebook::react;
   [self setHitArea:path];
   if (!CGRectEqualToRect(bounds, CGRectNull)) {
     self.clientRect = bounds;
-    self.fillBounds = CGPathGetBoundingBox(path);
-    self.strokeBounds = CGPathGetBoundingBox(self.strokePath);
+    self.fillBounds = CGPathGetPathBoundingBox(path);
+    self.strokeBounds = CGPathGetPathBoundingBox(self.strokePath);
     self.pathBounds = CGRectUnion(self.fillBounds, self.strokeBounds);
 
     CGAffineTransform current = CGContextGetCTM(context);
--- a/apple/Elements/RNSVGGroup.mm
+++ b/apple/Elements/RNSVGGroup.mm
@@ -129,8 +129,8 @@ using namespace facebook::react;
   [self setHitArea:path];
   if (!CGRectEqualToRect(bounds, CGRectNull)) {
     self.clientRect = bounds;
-    self.fillBounds = CGPathGetBoundingBox(path);
-    self.strokeBounds = CGPathGetBoundingBox(self.strokePath);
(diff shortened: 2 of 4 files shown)


omeid:
cc @msand

Since you have laid out the initial work for this.

omeid:
Pinging this as this is the only reason I need to maintain a fork and build expo apps.

bohdanprog:
@omeid can you rebase your branch to main?

omeid:
@bohdanprog Done!

omeid:
Friendly ping this so that it doesn't go out of sync again. @jakex7

jakex7:
Hey @omeid, thanks for the PR! The code looks good to me, but I'm not entirely sure what issue it's resolving. Could you provide some examples of code that doesn't work currently and how it's improved after the fix? 😄

omeid:
@jakex7 Currently calling `getBBox()` on any shape reference, it returns incorrect in iOS and Android.

On Android, it is wrongly scaled.
On iOS, it uses `CGPathGetBoundingBox` instead of  `CGPathGetPathBoundingBox`. The former includes the control points, while bounding-box is about the visual parts of a shape.

If you want, I can create a snack later.

jakex7:
@omeid sorry for the late response.  I was able to test this PR, and `CGPathGetPathBoundingBox` appears to work well. However, I'm still uncertain about the issue with the Android scaling. Could you provide a reproduction for this problem?

It seems the scaling issue was introduced in this commit: [f3e0b190a917c8f1ddc6a19ccd47563ad6c1ca59](https://github.com/omeid/react-native-svg/commit/f3e0b190a917c8f1ddc6a19ccd47563ad6c1ca59). Should we consider removing the scaling from other related areas as well?

Additionally, could you please split this into two separate PRs since they address different aspects?

omeid:
Thanks for the review @jakex7, I will have to test the other methods as I didn't have a need for them, I will create a separate PR for that along with any other changes that might be required for the scaling on Android.

omeid:
Once this is merged, I will create another PR for the Android side.

jakex7 (approved):
LGTM 🚀 Thank you!

PR #2197, opened 2023-12-22
Fix type exports and enable `isolatedModules` setting to catch it in `tsc`

renchap:
# Summary

This is the same issue as the one I fixed last year in #1874

To prevent this from occurring again, I enabled the `isolatedModules` Typescript option which will warn when you export non-value identifiers without using `export type`, which will break transpilation processes what work on a per-file basis. You can find a [more detailed explanation in Typescript's doc](https://www.typescriptlang.org/tsconfig#isolatedModules)

## Test Plan

I used this patch in my own app using [@rnx-kit/metro-serializer-esbuild](https://microsoft.github.io/rnx-kit/docs/tools/metro-serializer-esbuild) and it builds correctly. The preview version was already using `export type` without any reported issues.

## Compatibility

Not relevant

## Checklist

Not relevant

### DIFF
--- a/src/ReactNativeSVG.ts
+++ b/src/ReactNativeSVG.ts
@@ -141,15 +141,6 @@ export {
   camelCase,
   err,
   fetchText,
-  JsxAST,
-  Middleware,
-  Styles,
-  UriProps,
-  UriState,
-  XmlAST,
-  XmlProps,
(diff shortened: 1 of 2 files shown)


WoLewicki (approved):
LGTM 🚀

Issue #2233, opened 2024-02-26
toDataURL only works the second time it is executed in old Arch


imanderson:
Hello,

in iOS and old Arch, I found that the callback on toDataURL method will only be called the second time of asking. This is due to when the queue from UIManager in iOS is flushed. See https://github.com/facebook/react-native/issues/1365 

I traced the problem to the iOS implementation and will do a PR with my fix.

## Environment info

React native info output:

```bash
System:
  OS: macOS 14.3.1
  CPU: (12) arm64 Apple M2 Max
  Memory: 96.39 MB / 32.00 GB
  Shell:
    version: 3.6.0
    path: /opt/homebrew/bin/fish
Binaries:
  Node:
    version: 18.19.1
    path: ~/.nvm/versions/node/v18.19.1/bin/node
  Yarn:
    version: 1.22.21
    path: ~/.nvm/versions/node/v18.19.1/bin/yarn
  npm:
    version: 10.2.4
    path: ~/.nvm/versions/node/v18.19.1/bin/npm
  Watchman:
    version: 2024.01.22.00
    path: /opt/homebrew/bin/watchman
Managers:
  CocoaPods:
    version: 1.14.3
    path: /opt/homebrew/bin/pod
SDKs:
  iOS SDK:
    Platforms:
      - DriverKit 23.2
      - iOS 17.2
      - macOS 14.2
      - tvOS 17.2
      - visionOS 1.0
      - watchOS 10.2
  Android SDK: Not Found
IDEs:
  Android Studio: 2023.1 AI-231.9392.1.2311.11330709
  Xcode:
    version: 15.2/15C500b
    path: /usr/bin/xcodebuild
Languages:
  Java:
    version: 17.0.10
    path: /usr/bin/javac
  Ruby:
    version: 2.6.10
    path: /usr/bin/ruby
npmPackages:
  "@react-native-community/cli": Not Found
  react:
    installed: 18.2.0
    wanted: 18.2.0
  react-native:
    installed: 0.73.4
    wanted: 0.73.4
  react-native-macos: Not Found
npmGlobalPackages:
  "*react-native*": Not Found
Android:
  hermesEnabled: true
  newArchEnabled: false
iOS:
  hermesEnabled: true
  newArchEnabled: false
```

Library version: 14.1.0

## Steps To Reproduce

[...]

PR #2234, opened 2024-02-26
fix: toDataIURL not being called on iOS old arch

imanderson:
This fixes https://github.com/software-mansion/react-native-svg/issues/2233

## Compatibility

| OS      | Implemented |
| ------- | :---------: |
| iOS     |    ✅    |
| Android |    ❌     |

## Checklist

### DIFF
--- a/apple/RNSVGSvgViewModule.mm
+++ b/apple/RNSVGSvgViewModule.mm
@@ -26,7 +26,7 @@ RCT_EXPORT_MODULE()
          callback:(RCTResponseSenderBlock)callback
           attempt:(int)attempt
 {
-  void (^block)(void) = ^{
+
 #ifdef RCT_NEW_ARCH_ENABLED
     [self.viewRegistry_DEPRECATED addUIBlock:^(RCTViewRegistry *viewRegistry) {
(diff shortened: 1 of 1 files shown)


bohdanprog:
@WoLewicki Can you take a look at that PR? It seems good to me and resolves the issue.
Thanks.

apple/RNSVGSvgViewModule.mm
```diff
@@ -90,5 +84,13 @@ - (void)toDataURL:(nonnull NSNumber *)reactTag
...
 }
 #endif
+    
+- (dispatch_queue_t)methodQueue {
+    #ifdef RCT_NEW_ARCH_ENABLED
+        return dispatch_get_main_queue();
+    #else
+        return _bridge.uiManager.methodQueue;
+    #endif // RCT_NEW_ARCH_ENABLED
+}
```
bohdanprog, line 94: I checked that and there is some problem when we change `bridgelessEnabled`, I recommend to do that like that:
```suggestion
- (dispatch_queue_t)methodQueue {
    if(self.bridge) {
      return RCTGetUIManagerQueue();
    } else {
      return dispatch_get_main_queue();
    }
}
```

PR #2240, opened 2024-03-07
fix: Fix pod install not working with lower version of Cocoapods due to visionOS

jihoon416:
# Summary
On Cocoapods versions of lower than 1.14.3, pod install fails due to it not understanding the visionos field.
Our company cannot use a higher version of Cocoapods right now, so it would be beneficial if we or other similar people can use the latest version of react-native-svg without having to upgrade the Cocoapods version.

I was able to find this solution from https://github.com/getsentry/sentry-react-native/issues/3547#issuecomment-1907677879

## Compatibility

| OS      | Implemented |
| ------- | :---------: |
| iOS     |    ✅     |
| Android |    ❌     |

## Checklist


Arthurmtro (commented):
Good catch!

jakex7 (approved):
LGTM, thank you for your contribution! 🎉

PR #2270, opened 2024-05-09
fix: imports and remove `err` export

### LINKED ISSUE #2265: Why is `err` exported?
[This](https://github.com/software-mansion/react-native-svg/blob/a256f25de8681b4693b4c41575b2ebfd347eb686/src/xml.tsx#L108) seems like a very library specific reference, what makes this need to be exported? `err` is an extremely common term that, if one is not paying attention, causes more confusion than anything by exporting. I suggest that this, at the very least be renamed something more relevant in the context of this lib.

jakex7:
# Summary

* Removed `err` export from root file (ReactNativeSVG)
* replaced imports from `react-native-svg` with relative imports inside `src` dir

Resolves #2265


src/css/LocalSvg.tsx
```diff
@@ -3,9 +3,9 @@ import { useState, useEffect, Component } from 'react';
 import type { ImageSourcePropType } from 'react-native';
 import { Platform, Image } from 'react-native';
 
-import { fetchText } from 'react-native-svg';
+import type { SvgProps } from '../elements/Svg';
```
WoLewicki, line 6: As for these changes, I am afraid we should leave it as it is due to this PR: https://github.com/software-mansion/react-native-svg/pull/2168. It is meant to be a different package so the imports should behave accordingly.

WoLewicki (changes requested):
Removing the `err` is fine, but other changes are not needed as mentioned in the comment. I think we should revert them and then we can merge the PR 🎉

WoLewicki (approved):
LGTM!

PR #2271, opened 2024-05-10
fix: radialGradient `r={0}`

### LINKED ISSUE #2170: Android crashes if r="0" for radialGradient
# Bug

Android is crashing when using `<radialGradient r="0">` with message:

```
IllegalArgumentException
radius must be > 0
```

#### Demo
https://snack.expo.dev/@kocaagasefa/radial-gradient-demo

Library version: 13.4.0

## Steps To Reproduce
- run https://snack.expo.dev/@kocaagasefa/radial-gradient-demo on android

## Example svg: (from [MDN](https://developer.mozilla.org/en-US/docs/Web/SVG/Attribute/r#example))

```
<svg viewBox="0 0 300 200" xmlns="http://www.w3.org/2000/svg">
  <radialGradient r="0" id="myGradient000">
    <stop offset="0" stop-color="white" />
    <stop offset="100%" stop-color="black" />
  </radialGradient>
  <radialGradient r="50%" id="myGradient050">
    <stop offset="0" stop-color="white" />
    <stop offset="100%" stop-color="black" />
  </radialGradient>
  <radialGradient r="100%" id="myGradient100">
    <stop offset="0" stop-color="white" />
    <stop offset="100%" stop-color="black" />
  </radialGradient>

  <circle cx="50" cy="50" r="0" />
  <circle cx="150" cy="50" r="25" />
  <circle cx="250" cy="50" r="50" />

  <rect x="20" y="120" width="60" height="60" fill="url(#myGradient000)" />
  <rect x="120" y="120" width="60" height="60" fill="url(#myGradient050)" />
  <rect x="220" y="120" width="60" height="60" fill="url(#myGradient100)" />
</svg>
```

#### Describe what you expected to happen:
1. Render example svg given on https://developer.mozilla.org/en-US/docs/Web/SVG/Attribute/r#example

jakex7:
# Summary

* Fixed crash on Android when `<RadialGradient r={0}>`.
* Fixed render issue on Android and iOS when radius is zero. According to [MDN Docs](https://developer.mozilla.org/en-US/docs/Web/SVG/Attribute/r#radialgradient)
> A value of lower or equal to zero will cause the area to be painted as a single color using the color and opacity of the last gradient `<stop>`.

## Test Plan

Test is available in `TestsExample` as `Test2170`

## Compatibility

| OS      | Implemented |
| ------- | :---------: |
| iOS     |    ✅    |
| Android |    ✅    |

Fixes #2170

### DIFF
--- a/apple/Brushes/RNSVGPainter.mm
+++ b/apple/Brushes/RNSVGPainter.mm
@@ -232,6 +232,15 @@ void PatternFunction(void *info, CGContextRef context)
   CGFloat rx = [self getVal:[_points objectAtIndex:2] relative:width];
   CGFloat ry = [self getVal:[_points objectAtIndex:3] relative:height];
 
+  if (rx <= 0 || ry <= 0) {
+    // Gradient with radius = 0 should be rendered as solid color of the last stop
+    rx = width;
+    ry = height;
+    CGGradientRelease(gradient);
+    NSArray<NSNumber *> *gradientArray = @[_colors.firstObject, _colors.lastObject, _colors[_colors.count-2], _colors.lastObject];
+    gradient = CGGradientRetain([RCTConvert RNSVGCGGradient:gradientArray]);
+  }
+    
   double ratio = ry / rx;
 
   CGFloat fx = [self getVal:[_points objectAtIndex:0] relative:width] + offsetX;
--- a/TestsExample/src/Test2170.tsx
+++ b/TestsExample/src/Test2170.tsx
@@ -1,0 +1,65 @@
+import * as React from 'react';
+import {View} from 'react-native';
+import {
+  Circle,
+  RadialGradient,
+  Rect,
+  Stop,
+  Svg,
+  SvgXml,
+} from 'react-native-svg';
+
+// https://developer.mozilla.org/en-US/docs/Web/SVG/Attribute/r#example
+const svgXml = `
+<svg viewBox="0 0 300 200" xmlns="http://www.w3.org/2000/svg">
+  <radialGradient r="0" id="myGradient000">
+    <stop offset="0" stop-color="white" />
+    <stop offset="100%" stop-color="black" />
+  </radialGradient>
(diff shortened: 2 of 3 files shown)


apple/Brushes/RNSVGPainter.mm
```diff
@@ -232,6 +232,15 @@ - (void)paintRadialGradient:(CGContextRef)context bounds:(CGRect)bounds
   CGFloat rx = [self getVal:[_points objectAtIndex:2] relative:width];
   CGFloat ry = [self getVal:[_points objectAtIndex:3] relative:height];
 
+  if (rx <= 0 || ry <= 0) {
+    // Gradient with radius = 0 should be rendered as solid color of the last stop
+    rx = width;
+    ry = height;
+    CGGradientRelease(gradient);
+    NSArray<NSNumber *> *gradientArray = @[_colors.firstObject, _colors.lastObject, _colors[_colors.count-2], _colors.lastObject];
```
WoLewicki, line 240: Could you quickly explain what each element of this array corresponds to? And since we have `_colors.count-2`, is there a case there aren't 3 items in the `_colors` array?

jakex7: Sure! When `radialGradient`'s `r=0` it should be painted as a single color using the color of the last `stop`.

The `_colors` array corresponds to the `Stop` elements inside `RadialGradient`. It consists of `offset` and `stopColor` respectively so the `gradientArray` should look like this

`[offset from first, color from last, offset from last, color from last]`

In addition, above is a check that `_colors.count` is not 0 and given that it should always consists of an even number of elements (offset, color), even for `stop` with no props the default values should be passed.

WoLewicki (approved):
LGTM!

apple/Brushes/RNSVGPainter.mm
WoLewicki: Perfect, thanks for the explanation!

PR #2272, opened 2024-05-13
fix: toDataUrl line breaks

### LINKED ISSUE #1986: toDataURL incorrectly contains line breaks
# Bug

The `toDataURL` method contains line breaks `\r\n`, which break the base64 string.

#### Unexpected behavior
This can lead to corrupted images when e.g. using it in [react-native-share](https://github.com/react-native-share/react-native-share).

## Environment info

Library version: 13.4.0

https://snack.expo.dev/@robrechtme/data-url-bug

jakex7:
# Summary

This PR removes the flag to include custom line breaks on `toDataUrl` method. Some software could not read base64 with additional line break characters, so it could lead to corrupting the image (like in react-native-share). Fixes #1986.

## Test Plan

`TestsExample` -> `Test1986`

## Compatibility

| OS      | Implemented |
| ------- | :---------: |
| iOS     |    ✅     |
| Android |    ✅     |

### DIFF
--- a/TestsExample/App.js
+++ b/TestsExample/App.js
@@ -12,6 +12,7 @@ import Test2071 from './src/Test2071';
 import Test2089 from './src/Test2089';
 import Test2196 from './src/Test2196';
 import Test2266 from './src/Test2266';
+import Test1986 from './src/Test1986';
 
 export default function App() {
   return <ColorTest />;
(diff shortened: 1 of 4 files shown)


WoLewicki (approved):
LGTM 🎉

PR #2274, opened 2024-05-13
fix: Fix static view config validation for RNSVGSvgViewAndroid

fabriziocucci:
# Summary

As per title, when running an app in bridgless mode the following redbox appears:

> StaticViewConfigValidator: Invalid static view config for 'RNSVGSvgViewAndroid'.
> 
> 'validAttributes.borderBlockColor' is missing.
> 'validAttributes.borderBlockEndColor' is missing.
> 'validAttributes.borderBlockStartColor' is missing.
> 'validAttributes.borderEndEndRadius' is missing.
> 'validAttributes.borderEndStartRadius' is missing.
> 'validAttributes.borderStartEndRadius' is missing.
> 'validAttributes.borderStartStartRadius' is missing.

## Test Plan

Tested that the redbox is gone.

## Compatibility

| OS      | Implemented |
| ------- | :---------: |
| iOS     |    N/A     |
| Android |    ✅     |

## Checklist

### DIFF
--- a/android/src/main/java/com/horcrux/svg/SvgViewManager.java
+++ b/android/src/main/java/com/horcrux/svg/SvgViewManager.java
@@ -317,6 +317,21 @@ class SvgViewManager extends ReactViewManager
   }
 
   @Override
+  public void setBorderBlockColor(SvgView view, @Nullable Integer value) {
+    super.setBorderColor(view, 9, value);
+  }
+
+  @Override
+  public void setBorderBlockEndColor(SvgView view, @Nullable Integer value) {
+    super.setBorderColor(view, 10, value);
+  }
+
+  @Override
+  public void setBorderBlockStartColor(SvgView view, @Nullable Integer value) {
+    super.setBorderColor(view, 11, value);
+  }
+
+  @Override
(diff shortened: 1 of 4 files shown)


jakex7:
Hi @fabriziocucci, 
actually, I am not familiar with this error but I noticed that in LogCat after adding your changes there are a few more errors regarding `border`. Is this something we should be concerned about? How could we resolve them all? 

image

fabriziocucci:
@jakex7 thanks for raising the question, these are unfortunately two separate but connected problems.

This diff is a workaround to fix the static view config validation (similarly to what was already done in the past [here](https://github.com/software-mansion/react-native-svg/blob/main/src/fabric/AndroidSvgViewNativeComponent.ts#L67-L69)).

For the pre-existing console errors (+ the additional 7 added by this changes), I'm afraid we'll have to do something at framework level (e.g. push all border props in the BaseViewManger or something else).

Does that clarifies a bit?

jakex7 (approved):

WoLewicki (approved):
LGTM!

PR #2280, opened 2024-05-20
adding the bun option to the README

elmanci2:
adding the bun option to the README


elmanci2:
adding the bun option to the README

PR #2281, opened 2024-05-21
fix: error when building paper after fabric

maciekstosio:
# Summary

We had similar config in react-native-screens. It caused problem when building Paper arch app after Fabric. Those changes help, so I'm moving them here as well.


WoLewicki (approved):
LGTM!

PR #2288, opened 2024-06-04
feat: add task to automatically copy codegen files to paper

maciekstosio:
## Description

Sibling of [similar task from screens](https://github.com/software-mansion/react-native-screens/pull/2168).
When changing native props on Fabric, codegen generates corresponding interfaces and delegates. To make sure both implementations are consistent, we implement those interfaces on Paper too. Currently, after generating interfaces using codegen, developer needs to copy corresponding files for paper manually. This task adds Gradle task, that automates this.

## Changes

Add new task to build Gradle and necessary properties: 
- codegen artifacts dir and paper dir 
- flag in both fabric apps that indicates that copying should be performed (we do want this task to be performed only when developing the library)

## Test code and steps to reproduce

Remove `maskUnits` from `src/fabric/MaskNativeComponent.ts` and run ` ./gradlew generateCodegenArtifactsFromSchema` in `./FabricExample/android`. That should automatically copy regenerated files to paper directory.


maciekstosio:
Closing for now, will be replaces with https://github.com/software-mansion/react-native-screens/pull/2224

PR #2289, opened 2024-06-05
Handle error when pass wrong uri to SvgFromUri component

### LINKED ISSUE #2148: Bad svg uri content is causing app to freeze
# Bug

image

#### Unexpected behavior

Happens on all scenarios

## Environment info

"react-native-svg": "^13.14.0",
"react-native": "0.72.5",
"react": "18.2.0",
"expo": "^49.0.0",

## Steps To Reproduce

Load https://only1.io/like-token.svg using SvgFromUri, I was expecting to be a svg file but looks like the file doesn't exist anymore and the whole page is rendered.

1. If the file doesn't exist just throw the error and that's it but the app freezes, the buttons do not respond anymore
2. App should continue working even if the error was thrown

NA

bohdanprog:
# Summary

We want to add the ability to handle the Error when someone passes the wrong URI to the `SvgFromUri` component.

Closes #2148 

## Test Plan

Manual tests in TestsExample app. (Test2148)

### DIFF
--- a/TestsExample/App.js
+++ b/TestsExample/App.js
@@ -2,20 +2,20 @@
 import React from 'react';
 
 import ColorTest from './src/ColorTest';
+import PointerEventsBoxNone from './src/PointerEventsBoxNone';
 import Test1374 from './src/Test1374';
 import Test1718 from './src/Test1718';
 import Test1813 from './src/Test1813';
 import Test1845 from './src/Test1845';
-import Test2080 from './src/Test2080';
-import PointerEventsBoxNone from './src/PointerEventsBoxNone';
+import Test1986 from './src/Test1986';
 import Test2071 from './src/Test2071';
+import Test2080 from './src/Test2080';
 import Test2089 from './src/Test2089';
+import Test2148 from './src/Test2148';
 import Test2196 from './src/Test2196';
 import Test2266 from './src/Test2266';
-import Test1986 from './src/Test1986';
 import Test2276 from './src/Test2276';
-import Test2148 from './src/Test2148';
 
 export default function App() {
-  return <Test2148 />;
+  return <ColorTest />;
 }
--- a/TestsExample/App.js
+++ b/TestsExample/App.js
@@ -14,7 +14,8 @@ import Test2196 from './src/Test2196';
 import Test2266 from './src/Test2266';
 import Test1986 from './src/Test1986';
 import Test2276 from './src/Test2276';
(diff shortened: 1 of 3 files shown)


src/xml.tsx
```diff
@@ -171,10 +171,15 @@ export class SvgFromXml extends Component<XmlProps, XmlState> {
   }
 
   parse(xml: string | null) {
+    const { onError = err } = this.props;
     try {
       this.setState({ ast: xml ? parse(xml) : null });
     } catch (e) {
```
WoLewicki, line 177: Maybe we could do `e: Error` here to not cast below?

src/xml.tsx
```diff
@@ -171,10 +171,15 @@ export class SvgFromXml extends Component<XmlProps, XmlState> {
...
-      console.error(e);
+      const error = e as Error;
+      onError({
+        ...error,
+        message: `Couldn't parse SVG, case: ${error.message}`,
```
WoLewicki, line 181:
```suggestion
        message: `[RNSVG] Couldn't parse SVG, reason: ${error.message}`,
```

Would be better maybe?

TestsExample/App.js
```diff
@@ -14,7 +14,8 @@ import Test2196 from './src/Test2196';
 import Test2266 from './src/Test2266';
 import Test1986 from './src/Test1986';
 import Test2276 from './src/Test2276';
+import Test2148 from './src/Test2148';
 
 export default function App() {
-  return <ColorTest />;
+  return <Test2148 />;
```
WoLewicki, line 20: I think we want to leave `ColorTest` as the default one. Also, could you sort the imports so its easier to go through them?

WoLewicki (commented):
Please answer the comments, other than that LGTM :tada:

TestsExample/App.js
bohdanprog: Okay, I'm gonna change that :)

src/xml.tsx
bohdanprog: Yep, it looks better, thanks 👍

src/xml.tsx
bohdanprog: As far as I know, we can't catch error types, or is there something I'm missing?

WoLewicki: Oh, didn't know that. Then it's fine if we are sure that the `e` is of type `Error`.

WoLewicki (approved):
LGTM!

PR #2292, opened 2024-06-06
chore: remove macos-build-test.yml

bohdanprog:
# Summary

- remove `macos-build-test.yml`

## Test Plan

merge that PR to the main, and check if `macos-build` action will be run


WoLewicki (approved):
Would be perfect to check from time to time if v74 is supported but for now let's merge it.

Issue #1451, opened 2020-09-10
Animated viewBox - transform scale is blury


harveyconnor:
# Question
My issue is that whenever I animate my svg with a transform function it ends up being blury when I zoom in. The best alternative to this option would be to zoom in via the viewbox which will also offering a panning function.
I'm wondering if anyone has been successful in animating the viewBox?

The closest thing I've gotten to is adding an event listener on the animated value however that doesn't update the viewBox with the new scale adjustments.

Please any suggestions or recommendations are welcome! Thanks!

guiccbr:
Same with me. @harveyconnor , were you ever able to do that?

harveyconnor:
@guiccbr nope I could never get it to work. The support for this repo is dead.

guiccbr:
@harveyconnor , I got something working that was a good alternative to me. Instead of using the transform prop, I used the style prop on a <G>. The style prop is not typed though, but it works. And it's great because it doesn't distort the child paths.

Style also works for <Paths> btw. In my case I'm even using react-native-reanimated to animate stuff, and everything works great.

harveyconnor:
Hi @guiccbr could you please share a repo with us that replicates what you have created :)

meepak:
I was able to resolve this by simply changing width height within SVG to 100%, wrap the SVG around the Animated.View and use useAnimatedStyle to change width and height. This way I get the smooth zoom effect on svg whereas transform scale results in pixelated zoom.

My problem is I am unabe to do same with custom exiting animation, width and height manipulation not seems to be working there and transform seems to be the only option, is it?

jakex7:
> This way I get the smooth zoom effect on svg whereas transform scale results in pixelated zoom.

Hi! 
This issue (pixelated zoom) has likely been resolved in #2299 and merged into the main branch. Please feel free to test it and provide feedback before the release.

Thank you!

PR #2299, opened 2024-06-12
fix: scaling when mask is set

jakex7:
# Summary

This PR resolves an issue raised in #1451. 
Currently, when a mask is used, we render the element as a bitmap (or platform equivalent), but the bitmap's size does not update accordingly with transformations. With these changes, the problem is addressed as follows:
* **Android**: We utilize the original canvas layers to render the mask and element with the appropriate blending mode.
* **iOS**: We create an offscreen context with the size multiplied by the screen scale and apply the original UIGraphics CTM (current transformation matrix) to the offscreen context. This ensures that the same transformations are applied as on the original context.

Additionally, there is a significant performance improvement on Android as we are not creating three new Bitmaps and three new Canvases.

## Test Plan

There are many ways for testing these changes, but the required ones are:
* `TestsExample` app -> `Test1451.tsx`
* `Example` app -> Mask section
* `FabricExample` app -> Mask section

## Compatibility

| OS      | Implemented |
| ------- | :---------: |
| Android |    ✅     |
| iOS     |    ✅     |

## Preview

image
image


android/src/main/java/com/horcrux/svg/RenderableView.java
```diff
@@ -329,71 +330,41 @@ public void setPropList(@Nullable ReadableArray propList) {
...
-      Canvas originalCanvas = new Canvas(original);
-      Canvas maskCanvas = new Canvas(maskBitmap);
-      Canvas resultCanvas = new Canvas(result);
+      // prepare maskPaint with luminanceToAlpha + PorterDuffXfermode and create new layer with it
+      Paint maskPaint = new Paint(Paint.ANTI_ALIAS_FLAG);
+      // luminanceToAlpha conversion TODO: use filters
+      ColorMatrix luminanceToAlpha =
+          new ColorMatrix(
+              new float[] {
+                0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0.2125f, 0.7154f, 0.0721f, 0, 0
```
WoLewicki, line 350: Could we get a comment from where we get those numbers?

apple/RNSVGRenderable.mm
```diff
@@ -257,42 +257,38 @@ - (void)renderTo:(CGContextRef)context rect:(CGRect)rect
...
-    NSUInteger npixels = scaledHeight * scaledWidth;
-    CGRect drawBounds = CGRectMake(0, 0, width, height);
-
-    // Allocate pixel buffer and bitmap context for mask
-    NSUInteger bytesPerPixel = 4;
-    NSUInteger bitsPerComponent = 8;
-    NSUInteger bytesPerRow = bytesPerPixel * scaledWidth;
-    CGColorSpaceRef colorSpace = CGColorSpaceCreateDeviceRGB();
+    // TODO: make it the right way, currently this is a hack to get the correct scale for the mask
+    scale = scale * self.matrix.a * self.svgView.viewBoxTransform.a;
```
WoLewicki, line 261: Can we get more explanation why we do it this way?

apple/RNSVGRenderable.mm
```diff
@@ -257,42 +257,38 @@ - (void)renderTo:(CGContextRef)context rect:(CGRect)rect
...
+    CGColorSpaceRef colorSpace = CGColorSpaceCreateDeviceRGB();
+    CGContextRef offscreenContext = CGBitmapContextCreate(
         pixels,
-        scaledWidth,
-        scaledHeight,
-        bitsPerComponent,
-        bytesPerRow,
+        scaledRect.size.width,
+        scaledRect.size.height,
+        8,
```
WoLewicki, line 272: It would be nice to leave those exact values as variables so they are more descriptive.

WoLewicki (commented):
Please test it on many different examples since it changes the mask drawing logic. I also left some comments, please answer them. Looks good overall 🚀

apple/RNSVGRenderable.mm
jakex7: I believe there is currently no way to directly obtain the end scale of the current element, so we calculate it a bit silly way as: `screen scale * current element scale * viewBox scale`

jakex7:
> Please test it on many different examples since it changes the mask drawing logic. I also left some comments, please answer them. Looks good overall 🚀

I can confirm that all examples from these sites as well as a few random images found on the web are working as expected
https://developer.mozilla.org/en-US/docs/Web/SVG/Element/mask
https://www.w3schools.com/graphics/svg_clipping_masking.asp

WoLewicki (approved):

android/src/main/java/com/horcrux/svg/RenderableView.java
```diff
@@ -329,71 +330,65 @@ public void setPropList(@Nullable ReadableArray propList) {
...
 
-      Canvas originalCanvas = new Canvas(original);
-      Canvas maskCanvas = new Canvas(maskBitmap);
-      Canvas resultCanvas = new Canvas(result);
+      Paint dstInPaint = new Paint();
+      dstInPaint.setXfermode(new PorterDuffXfermode(PorterDuff.Mode.DST_IN));
 
-      // Clip to mask bounds and render the mask
+      // calculate mask bounds
       float maskX = (float) relativeOnWidth(mask.mX);
```
WoLewicki, line 352: Can we move those to a place where they are used, so after the `canvas.saveLayer(null, luminancePaint);` ?

android/src/main/java/com/horcrux/svg/RenderableView.java
```diff
@@ -329,71 +330,65 @@ public void setPropList(@Nullable ReadableArray propList) {
...
       float maskY = (float) relativeOnHeight(mask.mY);
       float maskWidth = (float) relativeOnWidth(mask.mW);
       float maskHeight = (float) relativeOnHeight(mask.mH);
-      maskCanvas.clipRect(maskX, maskY, maskWidth + maskX, maskHeight + maskY);
 
-      Paint maskPaint = new Paint(Paint.ANTI_ALIAS_FLAG);
-      mask.draw(maskCanvas, maskPaint, 1);
+      // step 3 - combined layer
```
WoLewicki, line 357: Why is the step 3 the first step? Maybe we should call it `Preparation for step 3` or something like that?

apple/Filters/LuminanceToAlpha.h
```diff
@@ -0,0 +1,19 @@
+#ifndef LuminanceToAlpha_h
+#define LuminanceToAlpha_h
+
+static CIImage *applyLuminanceToAlphaFilter(CIImage *inputImage)
+{
+  CIFilter *luminanceToAlpha = [CIFilter filterWithName:@"CIColorMatrix"];
+  [luminanceToAlpha setDefaults];
+  CGFloat alpha[4] = {0.2125, 0.7154, 0.0721, 0};
```
WoLewicki, line 8: Same comment as in `java` code would be nice here.

apple/RNSVGRenderable.mm
```diff
@@ -222,9 +223,30 @@ - (void)prepareForRecycle
 }
 #endif // RCT_NEW_ARCH_ENABLED
 
-UInt32 saturate(CGFloat value)
+static CGImageRef renderToImage(RNSVGRenderable *object, CGSize bounds, CGRect rect, CGRect *clip)
```
WoLewicki, line 226:
```suggestion
static CGImageRef renderToImage(RNSVGRenderable *renderable, CGSize bounds, CGRect rect, CGRect *clip)
```

Or `element` or something like this would be better I think

apple/RNSVGRenderable.mm
```diff
@@ -239,134 +261,42 @@ - (void)renderTo:(CGContextRef)context rect:(CGRect)rect
...
-        bytesPerRow,
-        colorSpace,
-        kCGImageAlphaPremultipliedLast | kCGBitmapByteOrder32Big);
-    CGContextScaleCTM(bcontext, iscale, iscale);
+    // Render content of current SVG Renderable to image
+    CGImageRef currentContent = renderToImage(self, boundsSize, rect, nil);
+    CIImage *contentSrcImage = [CIImage imageWithCGImage:currentContent];
 
+    // https://www.w3.org/TR/SVG11/masking.html#MaskElement
+    RNSVGMask *_maskNode = (RNSVGMask *)[self.svgView getDefinedMask:self.mask];
```
WoLewicki, line 275: why `_maskNode` and not just `maskNode`? Is it somehow messing with some other variable?

apple/RNSVGRenderable.mm
```diff
@@ -698,4 +628,48 @@ - (void)resetProperties
   self.merging = false;
 }
 
+static CIImage *transparentImage()
```
WoLewicki, line 631: Can we move those to some `utils` directory so this file does not grow too much? Also you have some static fields here and also at the top of the file.

apple/RNSVGRenderable.mm
```diff
@@ -239,134 +261,42 @@ - (void)renderTo:(CGContextRef)context rect:(CGRect)rect
...
+    CGImageRef maskContent = renderToImage(_maskNode, boundsSize, rect, &maskBounds);
+    CIImage *maskSrcImage = [CIImage imageWithCGImage:maskContent];
+
+    // Create mask with mask element alpha and mask luminance
+    CIImage *alphaMask = transformImageIntoAlphaMask(maskSrcImage);
+      
+    // Compose source image with alpha mask
+    CIImage *composite = applyBlendWithAlphaMask(contentSrcImage, alphaMask);
+
+    // Create masked image and release memory
```
WoLewicki, line 292: We do not release memory in the next line I think?

WoLewicki (approved):
Added some more comments, looks good overall. If you feel like its ready after answering the comments, you can merge it 🚀

PR #2305, opened 2024-06-17
fix: add default transform-origin value for web version

### LINKED ISSUE #1566: Default strokeLinecap value is different between Android and iOS/Web versions
# Bug

In your example project, in the circles section you have the pie example. Looking at the android preview of it, it seemed strange and very un-pie-like. Setting the strokeLinecap on the Circle to 'butt' made it look as it does on the web and iOS previews. It seems that the default value for this on Android is 'round'.

Default Android look
Default Web/iOS look

The example expo snack: https://snack.expo.io/hTb8nT1zw

bohdanprog:
# Summary
Closes #1566
Add default value for transform-origin web application.

## Test Plan
When we run our example, it should appear consistently across all platforms.

## Compatibility

| OS      | Implemented |
| ------- | :---------: |
| iOS     |    ✅     |
| Android |  ✅    |
| Web |  ✅    |

### DIFF
--- a/src/ReactNativeSVG.web.ts
+++ b/src/ReactNativeSVG.web.ts
@@ -228,14 +228,17 @@ const prepare = <T extends BaseProps>(
   const parsedTransform = parseTransformProp(transform, props);
   if (parsedTransform) {
     clean.transform = parsedTransform;
+    clean['transform-origin'] = '50% 50%';
   }
   const parsedGradientTransform = parseTransformProp(gradientTransform);
   if (parsedGradientTransform) {
     clean.gradientTransform = parsedGradientTransform;
+    clean['transform-origin'] = '50% 50%';
   }
   const parsedPatternTransform = parseTransformProp(patternTransform);
   if (parsedPatternTransform) {
     clean.patternTransform = parsedPatternTransform;
+    clean['transform-origin'] = '50% 50%';
   }
 
   clean.ref = (el: SVGElement | null) => {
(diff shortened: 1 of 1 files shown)


WoLewicki (commented):
Could you link a svg specification where it says that this is the default? Won't this change override the values set here: https://github.com/software-mansion/react-native-svg/blob/0596fd9dc56057e5c74fef8865404aad8c26c35f/src/ReactNativeSVG.web.ts#L221-L225 if someone does it?

bohdanprog:
> Could you link a svg specification where it says that this is the default? Won't this change override the values set here:
> 
> https://github.com/software-mansion/react-native-svg/blob/0596fd9dc56057e5c74fef8865404aad8c26c35f/src/ReactNativeSVG.web.ts#L221-L225
> 
> if someone does it?

 [here](https://drafts.csswg.org/css-transforms/#propdef-transform-origin) is svg specification.

WoLewicki:
Ok please add the link to it as a comment, maybe above the code utilizing it.

bohdanprog:
@WoLewicki, I rechecked and found that the changes didn't make sense, so I closed the PR.

bohdanprog:
@WoLewicki I checked different test cases using transform: `rotate`, `translate`, `skewX`, `scale`, and everything worked well.
IDK why I had a problem with that before, my bad, sorry.

adamhari:
I'm seeing the following error in my console on web after updating to `react-native-svg@15.12.0`:

_Invalid DOM property `transform-origin`. Did you mean `transformOrigin`?_

PR #2306, opened 2024-06-18
fix: behaviour of gradient-transform for web version

### LINKED ISSUE #1615: gradientTransform `translate` should be in relative instead of absolute value
# Bug

The translate values for gradientTransform are relative values, meaning give translate x a value of `.5` will move the item half way across the area.
However IOS `CGAffineTransform` api expects the `Tx` and `Ty` to be in absolute value instead of relative value, react-native-svg doesn't convert the percentage into absolute value.
Android seems to be experiencing the same issue.

#### Unexpected behavior

Testing SVG:

```
<svg xmlns="http://www.w3.org/2000/svg" xmlns:xlink="http://www.w3.org/1999/xlink" width="500" height="500" viewBox="0 0 500 500">
  <defs>
    <!-- gradientTransform="translate(-4.161 1.494) rotate(-90) scale(1 4.689)" gradientUnits="objectBoundingBox" -->
    <radialGradient 
    	id="a" cx="0.5" cy="0.5" r="0.25"    	
    	gradientTransform="translate(0.25 0.25)"
    >
      <stop offset="0" stop-color="#1024c4" />
      <stop offset="0.069" stop-color="#172596" />
      <stop offset="0.266" stop-color="#00155d" />
      <stop offset="0.906" stop-color="#080052" />
      <stop offset="1" stop-color="#090124" />
    </radialGradient>
  </defs>
  <rect fill="url(#a)" width="500" height="500" />
</svg>
```

This is how the svg rendered in chrome:

image

This is how the svg rendered in iOS and android
image

Click iOS to see how it rendered in iOS by react-native svg:
https://snack.expo.io/8hNuCxs2o

## Environment info

"react-native-svg": "^12.1.1",
for others, please check above expo snack

## Steps To Reproduce

1. Open that expo snack

1. expect the snack renders the same as in chrome

see above

bohdanprog:
Closes #1615

# Summary
We want to achieve the same behavior in the same platform when we use a gradient-transform prop.

## Test Plan

| OS | Before | After |
|--------|--------|--------|
| IOS |  |  |
| ANDROID |  |  |
| WEB |  |  | 

## Compatibility

| OS      | Implemented |
| ------- | :---------: |
| iOS     |    ✅     |
| Android |    ✅    |
| Web |    ✅    |

### DIFF
--- a/src/ReactNativeSVG.web.ts
+++ b/src/ReactNativeSVG.web.ts
@@ -197,7 +197,7 @@ const prepare = <T extends BaseProps>(
     onResponderTerminationRequest?: (e: GestureResponderEvent) => boolean;
     onClick?: (e: GestureResponderEvent) => void;
     transform?: string;
-    gradientTransform?: string;
+    'gradient-transform'?: string;
     patternTransform?: string;
     'transform-origin'?: string;
     style?: object;
@@ -231,7 +231,7 @@ const prepare = <T extends BaseProps>(
   }
   const parsedGradientTransform = parseTransformProp(gradientTransform);
   if (parsedGradientTransform) {
-    clean.gradientTransform = parsedGradientTransform;
+    clean['gradient-transform'] = parsedGradientTransform;
   }
   const parsedPatternTransform = parseTransformProp(patternTransform);
   if (parsedPatternTransform) {


src/ReactNativeSVG.web.ts
```diff
@@ -197,7 +197,7 @@ const prepare = <T extends BaseProps>(
     onResponderTerminationRequest?: (e: GestureResponderEvent) => boolean;
     onClick?: (e: GestureResponderEvent) => void;
     transform?: string;
-    gradientTransform?: string;
+    'gradient-transform'?: string;
     patternTransform?: string;
```
WoLewicki, line 201: What about `patternTransform`? Should we treat it the same way? Is there a reason why it is not automated somehow? Are there any other props that should be kebabCased?

bohdanprog: I think make it in a separate PR.

WoLewicki: No problem, but are there any other props?

PR #2308, opened 2024-06-20
fix: android PathParser crash app if pass some wrong d prop

### LINKED ISSUE #2086: Invalid d attribute on the Path crashes expo without any information
# Bug

If you create a Path element with d attribute that have invalid data, it crashes whole application.

#### Unexpected behavior

Create SVG with Path that have invalid d, it will crash expo. I expect it to give an error that describe what is the problem. Snack just ignore the invalid part of the d attribute and render only what's valid.

## Environment info

React native info output:

```bash
(node:1150961) Warning: Accessing non-existent property 'padLevels' of module exports inside circular dependency
(Use `node --trace-warnings ...` to show where the warning was created)
System:
    OS: Linux 5.19 Ubuntu 22.04.2 LTS 22.04.2 LTS (Jammy Jellyfish)
    CPU: (8) x64 Intel(R) Core(TM) i5-10210U CPU @ 1.60GHz
    Memory: 19.32 GB / 30.98 GB
    Shell: 5.1.16 - /bin/bash
  Binaries:
    Node: 20.3.0 - ~/.nvm/versions/node/v20.3.0/bin/node
    Yarn: 1.22.19 - ~/.nvm/versions/node/v20.3.0/bin/yarn
    npm: 9.6.7 - ~/.nvm/versions/node/v20.3.0/bin/npm
    Watchman: Not Found
  SDKs:
    Android SDK: Not Found
  IDEs:
    Android Studio: Not Found
  Languages:
    Java: 1.8.0_362 - /usr/bin/javac
  npmPackages:
    @react-native-community/cli: Not Found
    react: 18.2.0 => 18.2.0 
    react-native: 0.71.8 => 0.71.8 
  npmGlobalPackages:
    *react-native*: Not Found
info React Native v0.72.1 is now available (your project is running on v0.71.8).
info Changelog: https://github.com/facebook/react-native/releases/tag/v0.72.1
info Diff: https://react-native-community.github.io/upgrade-helper/?from=0.71.8
info For more info, check out "https://reactnative.dev/docs/upgrading".
```

Library version: 13.4.0

I was forced to use this version with expo
And react-native 0.72 is not yet available with stable expo

If needed I can try to reproduce with raw react-native but it can take more time.

## Steps To Reproduce

Issues without reproduction steps or code are likely to stall.

1. git clone git@github.com:jcubic/rnsvg-path.git
2. cd rnsvg-path
3. npm install
5. npm run android
6. click on the button to change the SVG with an invalid Path d attribute

1. App crashes

bohdanprog:
Closes #2086
# Summary

The application crashes if an error is thrown when something goes wrong during path parse.

## Test Plan
You can easily check in that component `Test2086` how it works after the fix.

## Compatibility

| OS      | Implemented |
| ------- | :---------: |
| iOS     |    ✅     |
| Android |    ✅    |

### DIFF
--- a/android/src/main/java/com/horcrux/svg/PathParser.java
+++ b/android/src/main/java/com/horcrux/svg/PathParser.java
@@ -62,7 +62,7 @@ class PathParser {
 
       if (!has_prev_cmd && first_char != 'M' && first_char != 'm') {
         // The first segment must be a MoveTo.
-        throw new Error(String.format("Unexpected character '%c' (i=%d, s=%s)", first_char, i, s));
+        throw new IllegalArgumentException(String.format("Unexpected character '%c' (i=%d, s=%s)", first_char, i, s));
       }
 
       // TODO: simplify
@@ -75,7 +75,7 @@ class PathParser {
       } else if (is_number_start(first_char) && has_prev_cmd) {
         if (prev_cmd == 'Z' || prev_cmd == 'z') {
           // ClosePath cannot be followed by a number.
-          throw new Error(String.format("Unexpected number after 'z' (s=%s)", s));
+          throw new IllegalArgumentException(String.format("Unexpected number after 'z' (s=%s)", s));
         }
 
         if (prev_cmd == 'M' || prev_cmd == 'm') {
@@ -93,7 +93,7 @@ class PathParser {
           cmd = prev_cmd;
         }
       } else {
-        throw new Error(String.format("Unexpected character '%c' (i=%d, s=%s)", first_char, i, s));
+        throw new IllegalArgumentException(String.format("Unexpected character '%c' (i=%d, s=%s)", first_char, i, s));
       }
 
       boolean absolute = is_absolute(cmd);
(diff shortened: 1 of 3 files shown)


android/src/main/java/com/horcrux/svg/PathParser.java
```diff
@@ -62,7 +62,7 @@ static Path parse(String d) {
 
       if (!has_prev_cmd && first_char != 'M' && first_char != 'm') {
         // The first segment must be a MoveTo.
-        throw new Error(String.format("Unexpected character '%c' (i=%d, s=%s)", first_char, i, s));
+        throw new IllegalArgumentException(String.format("Unexpected character '%c' (i=%d, s=%s)", first_char, i, s));
```
WoLewicki, line 65: Why changing it from `Error` to `IllegalArgumentException` fixes the issue? Do we catch it somewhere and handle properly then?

bohdanprog: That fix displays the error and the reason why it happened without crashing the application.

WoLewicki (approved):
If it is the thing that is expected by the users of the library, then LGTM!

PR #2317, opened 2024-06-28
Fix: image onLoad props

bohdanprog:
# Summary
Change data structure returned from the Image onLoad event.
Add fix in new arch for returned sizes.

## Test Plan

Manually test in both architectures and platforms.

## Compatibility

| OS      | Implemented |
| ------- | :---------: |
| iOS     |      ✅     |
| Android |    ✅    |

### DIFF
--- a/apple/Elements/RNSVGImage.mm
+++ b/apple/Elements/RNSVGImage.mm
@@ -38,6 +38,16 @@
 using namespace facebook::react;
 #endif // RCT_NEW_ARCH_ENABLED
 
+static NSDictionary *onLoadParamsForSource(RCTImageSource *source)
+{
+  NSDictionary *dict = @{
+    @"uri" : source.request.URL.absoluteString,
+    @"width" : @(source.size.width),
+    @"height" : @(source.size.height),
+  };
+  return @{@"source" : dict};
+}
+
 @implementation RNSVGImage {
   CGImageRef _image;
   CGSize _imageSize;
@@ -133,9 +143,10 @@ using namespace facebook::react;
     // See for more info: T46311063.
     return;
   }
-    auto imageSource = _state->getData().getImageSource();
-    imageSource.size = {image.size.width, image.size.height};
-    static_cast<const RNSVGImageEventEmitter &>(*_eventEmitter).onLoad({imageSource.size.width, imageSource.size.height, imageSource.uri});
+  auto imageSource = _state->getData().getImageSource();
+  imageSource.size = {image.size.width, image.size.height};
+  static_cast<const RNSVGImageEventEmitter &>(*_eventEmitter)
+      .onLoad({imageSource.size.width, imageSource.size.height, imageSource.uri});
 
   dispatch_async(dispatch_get_main_queue(), ^{
     self->_image = CGImageRetain(image.CGImage);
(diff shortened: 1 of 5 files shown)


apple/Elements/RNSVGImage.mm
```diff
@@ -133,9 +143,12 @@ - (void)didReceiveImage:(UIImage *)image metadata:(id)metadata fromObserver:(voi
     // See for more info: T46311063.
     return;
   }
-    auto imageSource = _state->getData().getImageSource();
-    imageSource.size = {image.size.width, image.size.height};
-    static_cast<const RNSVGImageEventEmitter &>(*_eventEmitter).onLoad({imageSource.size.width, imageSource.size.height, imageSource.uri});
+  auto imageSource = _state->getData().getImageSource();
+  imageSource.size = {image.size.width, image.size.height};
+    static_cast<const RNSVGImageEventEmitter &>(*_eventEmitter).onLoad({.source = {
+        .width = imageSource.size.width * imageSource.scale,
```
WoLewicki, line 149: Why did you change the logic here to also include the scale?

apple/Elements/RNSVGImage.mm
```diff
@@ -133,9 +143,12 @@ - (void)didReceiveImage:(UIImage *)image metadata:(id)metadata fromObserver:(voi
     // See for more info: T46311063.
     return;
   }
-    auto imageSource = _state->getData().getImageSource();
-    imageSource.size = {image.size.width, image.size.height};
-    static_cast<const RNSVGImageEventEmitter &>(*_eventEmitter).onLoad({imageSource.size.width, imageSource.size.height, imageSource.uri});
+  auto imageSource = _state->getData().getImageSource();
+  imageSource.size = {image.size.width, image.size.height};
+    static_cast<const RNSVGImageEventEmitter &>(*_eventEmitter).onLoad({.source = {
```
WoLewicki, line 148: I think it is nice to add a check for if `_eventEmitter` is nil, same as in RNScreens: https://github.com/software-mansion/react-native-screens/blob/a9bc1118b755c362a981638c0ce0091affd46fe0/ios/RNSScreen.mm#L323.

apple/Elements/RNSVGImage.mm
```diff
@@ -203,18 +216,16 @@ - (void)setSrc:(RCTImageSource *)src
...
-                          });
-                          [self invalidate];
-                        });
-                      }];
+                       dispatch_async(dispatch_get_main_queue(), ^{
+                         self->_image = CGImageRetain(image.CGImage);
+                         self->_imageSize = CGSizeMake(CGImageGetWidth(self->_image), CGImageGetHeight(self->_image));
+                         if (self->_onLoad) {
+                           RCTImageSource *sourceLoaded = [src imageSourceWithSize:image.size scale:image.scale];
+                           self->_onLoad(onLoadParamsForSource(sourceLoaded));
```
WoLewicki, line 224: Does it make sense to add it if we only use it in one place? Or is it potentially going to be reused in other events?

WoLewicki (changes requested):
Looks good! I left some comments, please answer them.

apple/Elements/RNSVGImage.mm
bohdanprog: without that unfortunately the app crashed in old architecture.

apple/Elements/RNSVGImage.mm
bohdanprog: okay, the good point :)

apple/Elements/RNSVGImage.mm
WoLewicki: I meant `onLoadParamsForSource` method.

apple/Elements/RNSVGImage.mm
WoLewicki: Ok let's do it like that then.

apple/Elements/RNSVGImage.mm
bohdanprog: without that check, the app expects us to pass onLoad prop every time.

apple/Elements/RNSVGImage.mm
bohdanprog: done

WoLewicki (approved):
So if it is now aligned between platforms and architectures, I think we are good to go :tada:

PR #2325, opened 2024-07-03
fix: extract opacity use percentage values

jakex7:
# Summary

According to the MDN Docs `opacity` as a `<alpha-value>` can be represented as
> A [<number>](https://developer.mozilla.org/en-US/docs/Web/CSS/number) in the range 0.0 to 1.0, inclusive, or a [<percentage>](https://developer.mozilla.org/en-US/docs/Web/CSS/percentage) in the range 0% to 100%, inclusive, representing the opacity of the channel (that is, the value of its alpha channel). Any value outside the interval, though valid, is clamped to the nearest limit in the range.

https://developer.mozilla.org/en-US/docs/Web/CSS/opacity#alpha-value
https://developer.mozilla.org/en-US/docs/Web/SVG/Attribute/fill-opacity#usage_notes

### DIFF
--- a/src/lib/extract/extractOpacity.ts
+++ b/src/lib/extract/extractOpacity.ts
@@ -1,6 +1,9 @@
 import type { NumberProp } from './types';
 
 export default function extractOpacity(opacity: NumberProp | void) {
+  if (typeof opacity === 'string' && opacity.trim().endsWith('%')) {
+    return +opacity.slice(0, -1) / 100;
+  }
   const value = +opacity;
-  return isNaN(value) ? 1 : value;
+  return isNaN(value) || value > 1 ? 1 : Math.max(value, 0);
 }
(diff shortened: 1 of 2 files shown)


src/lib/extract/extractOpacity.ts
```diff
@@ -1,6 +1,9 @@
 import type { NumberProp } from './types';
 
 export default function extractOpacity(opacity: NumberProp | void) {
+  if (typeof opacity === 'string' && opacity.trim().endsWith('%')) {
+    return +opacity.slice(0, -1) / 100;
```
WoLewicki, line 5: What about some weird % values as `500%`? It probably doesn't make sense but if we handle it with normal values then maybe we could to the same with %? If not, then we can just merge it as it is.

jakex7: Fixed in https://github.com/software-mansion/react-native-svg/pull/2325/commits/f63dcd7f2c192c9eb6b45dc403732ff8c78a5f01

jakex7:
I've also added `extractOpacity` to the `Svg` `opacity` property, as it should also accept opacity as a percentage. It's necessary as we're extracting this to the View wrapper.

PR #2337, opened 2024-07-09
fix: macos mask

jakex7:
# Summary

Fixed regression introduced in #2299. Now `<Mask>` is rendering properly on MacOS.

## Test Plan

* `Test1451` in `TestsExample`
* Masking section in `example`

https://github.com/software-mansion/react-native-svg/assets/39670088/dccc937d-125d-4b60-864b-cc3b9a4179fb

## Compatibility

| OS      | Implemented |
| ------- | :---------: |
| macOS     |    ✅     |

### DIFF
--- a/apple/RNSVGRenderable.mm
+++ b/apple/RNSVGRenderable.mm
@@ -280,6 +280,10 @@ UInt32 saturate(CGFloat value)
         bytesPerRow,
         colorSpace,
         kCGImageAlphaPremultipliedLast | kCGBitmapByteOrder32Big);
+#if TARGET_OS_OSX // [macOS]
+    // on macOS currentCTM is not scaled properly with screen scale so we need to scale it manually
+    CGContextConcatCTM(bcontext, screenScaleCTM);
+#endif // [macOS]
     CGContextConcatCTM(bcontext, currentCTM);
 
     // Clip to mask bounds and render the mask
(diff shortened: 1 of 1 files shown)


apple/RNSVGRenderable.mm
```diff
@@ -348,33 +352,35 @@ - (void)renderTo:(CGContextRef)context rect:(CGRect)rect
     CGImageRelease(maskImage);
 #else // [macOS
       // Render content of current SVG Renderable to image
-    UIGraphicsBeginImageContextWithOptions(maskBounds.size, NO, 0.0);
+    UIGraphicsBeginImageContextWithOptions(scaledRect.size, NO, 1.0);
     CGContextRef newContext = UIGraphicsGetCurrentContext();
-    CGContextTranslateCTM(newContext, 0.0, height);
-    CGContextScaleCTM(newContext, 1.0, -1.0);
-    [self renderLayerTo:newContext rect:rect];
+    CGContextConcatCTM(newContext, CGAffineTransformInvert(CGContextGetCTM(newContext)));
```
WoLewicki, line 357: Could we get a comment of what we exactly do in those next lines?

apple/RNSVGRenderable.mm
```diff
@@ -348,33 +352,35 @@ - (void)renderTo:(CGContextRef)context rect:(CGRect)rect
...
     CGImageRelease(contentImage);
 
     CGImageRef blendedImage = CGBitmapContextCreateImage(newContext);
     UIGraphicsEndImageContext();
 
     // Render blended result into current render context
-    CGContextDrawImage(context, maskBounds, blendedImage);
+    CGContextConcatCTM(context, CGAffineTransformInvert(currentCTM));
+    CGContextDrawImage(context, rect, blendedImage);
+    CGContextConcatCTM(context, currentCTM);
```
WoLewicki, line 383: Why do we have to concat at the end?

WoLewicki (approved):
Left some comments. If it works, then I am OK with merging it.

apple/RNSVGRenderable.mm
jakex7: Since it's the "original" `context`, we need to restore it to its original state so that the subsequent elements render correctly.

PR #2338, opened 2024-07-09
feat: e2e snapshot tests

CDFN:
# Summary

This PR adds E2E tests based on view screenshots done via `react-native-view-shot`. It only works with devices that have their [pixel ratio](https://reactnative.dev/docs/pixelratio) equal `3`. If you want to use device with different pixel ratio, you need to adjust it in `e2e/generateReferences.ts` viewport and regenerate reference images (see below).

Steps to run tests:
- Run Metro server for example app via `yarn start` in example app's directory 
- Run `example` app on platform of your choice (currently only Android & iOS are supported) via `yarn android` or `yarn ios` in example app's directory
- Run `yarn e2e` in project's root directory to start Jest server
- Select `E2E` tab in example app
- Wait for tests to finish
- You can see test results, as well as diffs (actual rendered svg vs reference image) in `e2e/diffs` directory

Steps to add new test cases:
- Put SVG of your choice to `e2e/cases` directory
- Run `yarn generateE2eRefrences`, this will open headless chrome browser via `puppeteer` and snapshot all rendered SVGs to .png files and later use them as reference in tests
- You should see new .png files in `e2e/references`
- When you run E2E tests again, it will use new test case(s) you've added

## Test Plan

https://github.com/software-mansion/react-native-svg/assets/41289688/24ee5447-ce9a-43b6-9dde-76229d25a30a

https://github.com/software-mansion/react-native-svg/assets/41289688/71d1873f-8155-4494-80bd-e4c1fa72a065

See Summary

See Summary

## Compatibility

| OS      | Implemented |
| ------- | :---------: |
| iOS     |    ✅     |
| Android |    ✅     |
| Web | ❌ |
## Checklist

- [X] I updated the typed files (typescript)
- [X] I added a test for the API in the `__tests__` folder


CDFN:
Added html reports 🎉 
image

example/babel.config.js
```diff
@@ -3,5 +3,6 @@ module.exports = {
   plugins: [
     '@babel/plugin-proposal-export-namespace-from',
     'react-native-reanimated/plugin',
```
bohdanprog, line 5: [react-native-reanimated/plugin](https://docs.swmansion.com/react-native-reanimated/docs/fundamentals/getting-started/#step-2-add-reanimateds-babel-plugin) has to be listed last.

CDFN: fixed, thanks

e2e/generateReferences.ts
```diff
@@ -0,0 +1,35 @@
+const path = require('path');
+const puppeteer = require('puppeteer')
+const fs = require('fs');
```
bohdanprog, line 3: Why do we use `require` here instead of `import`?

apps/examples/src/e2e/TestingView.tsx
```diff
@@ -0,0 +1,154 @@
...
+        JSON.stringify({
+          os: Platform.OS,
+          version: Platform.Version,
+          arch: isFabric() ? 'fabric' : 'paper',
+          connectionTime: new Date(),
+        }),
+      );
+      setMessage('✅ Connected to Jest server. Waiting for render requests.');
+    };
+    client.onerror = (err: any) => {
```
bohdanprog, line 38: It's just a suggestion, but have you considered using an [Error](https://stackoverflow.com/a/54649623/24106872) type instead of any?

example/package.json
```diff
@@ -18,6 +18,8 @@
     "react-native": "0.74.2",
     "react-native-reanimated": "3.13.0",
     "react-native-svg": "link:../",
+    "react-native-view-shot": "4.0.0-alpha.2",
+    "react-native-web": "^0.19.10",
```
bohdanprog, line 22: could we remove that react-native-web dependency?
We extract our web-example to another directory.

e2e/generateReferences.ts
CDFN: It's ran via `node-ts` which does not work quite well with ES6 from my experience.

apps/examples/src/e2e/TestingView.tsx
CDFN: Typing is tricky in this part of project (whole e2e dir basically). I tried to make it as good as possible, but here it would fail for some reason. `any` is good enough for now.

example/package.json
CDFN: Good point, I had to miss it while merging upstream changes. Will change.

apps/examples/src/e2e/TestingView.tsx
```diff
@@ -0,0 +1,154 @@
...
+
+const address = ['ios', 'web'].includes(Platform.OS) ? 'localhost' : '10.0.2.2';
+const wsUri = `ws://${address}:7123`;
+
+const TestingView = () => {
+  const [renderedContent, setRenderedContent] = useState();
+  const [wsClient, setWsClient] = useState<WebSocket | null>(null);
+  const [readyToSnapshot, setReadyToSnapshot] = useState(false);
+  const wrapperRef = useRef<ViewShot>(null);
+  const [resolution, setResolution] = useState([600, 600]);
```
bohdanprog, line 20: If I understand correctly that `const [resolution, setResolution] = useState([600, 600]);` based on that [comment](https://github.com/software-mansion/react-native-svg/pull/2338/files#diff-c78b493fa10a00e2c85350027da97e6b7bbb15978657add1f56762148fd26c9fR11-R13), like `size * pixelratio`?

apps/examples/src/e2e/TestingView.tsx
```diff
@@ -0,0 +1,154 @@
...
+function isFabric(): boolean {
+  // @ts-expect-error nativeFabricUIManager is not yet included in the RN types
+  return !!global?.nativeFabricUIManager;
+}
+
+export {samples, icon};
+
+const createElementFromObject = (
+  element: keyof typeof RNSVG,
+  props: any,
```
bohdanprog, line 136: Do you know what type this is?

apps/examples/src/e2e/TestingView.tsx
```diff
@@ -0,0 +1,154 @@
...
+): any => {
+  const children: any[] = [];
```
bohdanprog, line 138: and also here?

apps/examples/src/e2e/TestingView.tsx
```diff
@@ -0,0 +1,154 @@
...
+  if (props.children) {
+    if (Array.isArray(props.children)) {
+      props?.children.forEach((child: {type: any; props: any}) =>
+        children.push(createElementFromObject(child.type, child?.props)),
+      );
+      console.log(children);
```
bohdanprog, line 144: Do we need that `console.log`?

apps/examples/src/e2e/TestingView.tsx
```diff
@@ -0,0 +1,154 @@
...
+    } else if (typeof props.children === 'object') {
+      children.push(
+        createElementFromObject(props.children.type, props.children?.props),
+      );
+    } else {
+      children.push(props.children);
+    }
+  }
+  return React.createElement(RNSVG[element] as any, {...props, children});
```
bohdanprog, line 153: May we use that type `keyof React.JSX.IntrinsicElements` instead of `any`?

apps/examples/src/e2e/TestingView.tsx
CDFN: Yes, those are just initial values, they are later replaced with values demanded by a client (jest server): 
https://github.com/software-mansion/react-native-svg/blob/b3ceb8fec9a5e8e92c1c787906e09aea255eb4ba/src/lib/__tests__/e2e/GeneralSvgRenderingTest.spec.tsx#L30-L35
Probably should use [0, 0] to avoid confusion or document it better.

apps/examples/src/e2e/TestingView.tsx
CDFN: it can be literally `any` props, it's passed to React#createElement eventually, so we don't really need to know what's under there. If you have any suggestions on how to type it I'm happy to change it.

e2e/generateReferences.ts
```diff
@@ -0,0 +1,35 @@
...
+
+const main = async () => {
+  const browser = await puppeteer.launch()
+  const page = await browser.newPage()
+  await page.setViewport({
+    height: 200,
+    width: 200,
```
bohdanprog, line 10: Maybe we should get those values from environment variables, or we could specify them as constant values somewhere.

apps/examples/src/e2e/TestingView.tsx
CDFN: I don't think we need typing here, it will add extra effort to correctly type it without real benefits. This method is unlikely to be changed in future (unless React API changes).

e2e/setupJest.ts
```diff
@@ -0,0 +1,35 @@
...
+        console.log(`[react-native-svg] Running E2E test suites...\n`);
+        resolve();
+      });
+    });
+
+    console.log(
+      '\n\n[react-native-svg] E2E WebSocket server is running, waiting for client connection. Run example app and select E2E from examples list.\n'
+    );
+    global.server = wsServer;
+  });
```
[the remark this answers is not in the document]
CDFN: I would apply that suggestion if we didn't use quite old Node which doesn't have this function. It was fully implemented just in March (https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Promise/withResolvers).

e2e/types.ts
```diff
@@ -0,0 +1,22 @@
...
+  data: {
+    os: string;
+    arch: 'paper' | 'fabric';
+    platformVersion: string;
+  };
+}
+
+export interface RenderRequest {
+  type: 'renderRequest';
+  data: any;
```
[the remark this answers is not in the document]
CDFN: Inside `data` there is just object which then is parsed to a component and rendered. I don't think we should touch this type as it literally reflects actual data inside there, it can be anything.

e2e/generateReferences.ts
CDFN: It will be hard to share this value between generateReferences and GeneralSvgRenderingTest, also we don't expect them to change any time soon (it would require regenerating all references).

e2e/types.ts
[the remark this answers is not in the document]
CDFN: Changed to `unknown`

src/lib/__tests__/e2e/GeneralSvgRenderingTest.spec.tsx
```diff
@@ -0,0 +1,93 @@
+import { SvgFromXml } from 'react-native-svg';
```
jakex7, line 1: I think that `__tests__` folder should not be inside `src` folder (or at least it should be excluded from bundle).

apps/examples/src/e2e/TestingView.tsx
```diff
@@ -0,0 +1,157 @@
...
+  // @ts-expect-error nativeFabricUIManager is not yet included in the RN types
+  return !!global?.nativeFabricUIManager;
+}
+
+export {samples, icon};
+
+const createElementFromObject = (
+  element: keyof typeof RNSVG,
+  props: any,
+): any => {
```
jakex7, line 141: Since this function returns `React.createElement` we could at least type it as `ReactElement`
```suggestion
): React.ReactElement => {
```

apps/examples/src/e2e/TestingView.tsx
```diff
@@ -0,0 +1,157 @@
...
+import * as RNSVG from 'react-native-svg';
+import ViewShot from 'react-native-view-shot';
+
+const address = ['ios', 'web'].includes(Platform.OS) ? 'localhost' : '10.0.2.2';
+const wsUri = `ws://${address}:7123`;
+
+const TestingView = () => {
+  const wrapperRef = useRef<ViewShot>(null);
+  const [wsClient, setWsClient] = useState<WebSocket | null>(null);
+  const [renderedContent, setRenderedContent] = useState();
```
jakex7, line 18: > Since this function returns `React.createElement` we could at least type it as `ReactElement`
https://github.com/software-mansion/react-native-svg/pull/2338/files#r1698379189
```suggestion
  const [renderedContent, setRenderedContent] =
    useState<React.ReactElement | null>();
```

.github/workflows/android-e2e.yml
```diff
@@ -23,7 +23,7 @@ jobs:
       SYSTEM_IMAGES: system-images;android-34;google_apis;x86_64
       AVD_NAME: rn-svg-avd
     concurrency:
-      group: android-example-${{ github.ref }}
+      group: android-e2e-example-${{ github.ref }}
```
jakex7, line 26: Should we group all E2E tests together? If something fails, we don't want to ignore it anyway.

apps/examples/src/e2e/TestingView.tsx
bohdanprog: Good point, thanks

apps/examples/src/e2e/TestingView.tsx
bohdanprog: Good point, thanks

.github/workflows/android-e2e.yml
CDFN: This is concurrency group which will just cancel in-progress runs when new one is scheduled. Unless I missed the point.

e2e/cases/1.svg
WoLewicki, line 1: Where is case 2?

e2e/helpers.ts
```diff
@@ -0,0 +1,57 @@
...
+  });
+
+  const pixelDiff = pixelmatch(
+    referencePng.data,
+    responsePng.data,
+    diffPng.data,
+    opts.width * opts.pixelRatio,
+    opts.height * opts.pixelRatio,
+    {
+      diffColor: [95, 0, 160],
```
WoLewicki, line 47: Can we get a comment here or something? They look like magic numbers.

.github/workflows/android-e2e.yml
jakex7: Yes, that was an open question if we want it to cancel the remaining e2e tests if something is wrong.

bohdanprog: Hm, if we place the same concurrency group, our test build never runs.

tsconfig.json
```diff
@@ -21,5 +21,8 @@
     "resolveJsonModule": true,
     "isolatedModules": true
   },
-  "include": ["src"]
+  "include": [
+    "src",
+    "e2e"
```
WoLewicki, line 26: This tsconfig will also end up in user's code and we don't want e2e here I think?

e2e/cases/1.svg
CDFN: That was removed so the android workflow can be cached (it wouldn't get until it completes successfully). It was gradient so Android would fail on it. Just for the background why it was removed. It can be re-added, I'll send it over to @bohdanprog so he can add it back.

e2e/helpers.ts
CDFN: Note for @bohdanprog: they are indeed, it's just random RGB color.

.github/workflows/android-e2e.yml
CDFN: > Yes, that was an open question if we want it to cancel the remaining e2e tests if something is wrong.

They won't get cancelled if something is wrong. Whenever a commit is added to a branch, it will just cancel run for previous commit so resources aren't wasted. It doesn't affect other workflows, neither tests.

tsconfig.json
bohdanprog: @WoLewicki, without including here `e2e` and `__tests__/e2e/GeneralSvgRenderingTest.spec.tsx` we have problems inside those folders with typescript.

e2e/cases/1.svg
bohdanprog: I put case 2 in our cases folder.

src/lib/__tests__/e2e/GeneralSvgRenderingTest.spec.tsx
bohdanprog: I moved that file to `__tests_/e2e` directory

example/patches/react-native-view-shot+4.0.0-alpha.2.patch
```diff
@@ -35,17 +35,4 @@ index bd55b92..6a20e9d 100644
...
-+++ b/node_modules/react-native-view-shot/src/specs/NativeRNViewShot.ts
-@@ -2,7 +2,7 @@ import type { TurboModule } from 'react-native';
- import { TurboModuleRegistry } from 'react-native';
-
- export interface Spec extends TurboModule {
--  releaseCapture: () => string;
-+  releaseCapture: (uri: string) => void;
-   captureRef: (tag: number, options: Object) => Promise<string>
-   captureScreen: (options: Object) => Promise<string>;
- }
```
jakex7: Is this patch no longer needed?

.github/workflows/windows-build-test.yml
```diff
@@ -57,4 +57,4 @@ jobs:
 
       - name: Build app
         working-directory: ${{ matrix.working-directory }}/windows
-        run: npx react-native run-windows --logging --no-packager --no-deploy
\ No newline at end of file
+        run: npx react-native run-windows --logging --no-packager --no-deploy --no-autolink
```
jakex7, line 60: Why we're disabling autolinking on windows CI?

bohdanprog: Without that `--no-autolink` new libraries will be linking on the Windows platform. 
We do not want to install react-native-view-shot on the Windows platform.

example/patches/react-native-view-shot+4.0.0-alpha.2.patch
bohdanprog: Doesn't it fix some issues on the IOS platform?

jakex7: What do you mean? It looks like a codegen fix

.github/workflows/windows-build-test.yml
jakex7: Doesn't it mean no autolinking at all? Shouldn't we link other libraries manually in that case?

bohdanprog: Any of the new libraries aren't needed in the Windows build test.

example/patches/react-native-view-shot+4.0.0-alpha.2.patch
bohdanprog: Okay, maybe I overlooked something. I didn't include that patch.

CDFN: It is 100% needed, it was introduced by me and it was indeed fixing their codegen issue. This function had incorrect signature, so I fixed it with a patch.

jakex7 (approved):
Good job! 🚀
LGTM

bohdanprog:
Updated Podfile.lock across all our example apps.
I believe that the PR is ready to be merged.

bohdanprog (approved):

Issue #2192, opened 2023-12-11
Unable to build for macOS with new arch enabled


ryanlntn:
# Bug

RNSVG fails to build on macOS when `RCT_NEW_ARCH_ENABLED`. 

#### Unexpected behavior

The build fails with `Property 'center' not found on object of type 'RNSVGUse *'` in [RNSVGUse.mm](https://github.com/software-mansion/react-native-svg/blob/main/apple/Elements/RNSVGUse.mm#L170). This doesn't happen with iOS builds or without the new arch enabled.

## Environment info

React native info output:

```bash
System:
  OS: macOS 14.1.2
  CPU: (10) arm64 Apple M1 Pro
  Memory: 198.45 MB / 16.00 GB
  Shell:
    version: "5.9"
    path: /bin/zsh
Binaries:
  Node:
    version: 21.2.0
    path: ~/.asdf/installs/nodejs/21.2.0/bin/node
  Yarn:
    version: 1.22.21
    path: /opt/homebrew/bin/yarn
  npm:
    version: 10.2.3
    path: ~/.asdf/plugins/nodejs/shims/npm
  Watchman: Not Found
Managers:
  CocoaPods:
    version: 1.12.1
    path: /Users/ryan/.asdf/shims/pod
SDKs:
  iOS SDK:
    Platforms:
      - DriverKit 23.0
      - iOS 17.0
      - macOS 14.0
      - tvOS 17.0
      - watchOS 10.0
  Android SDK: Not Found
IDEs:
  Android Studio: 2022.2 AI-222.4459.24.2221.10121639
  Xcode:
    version: 15.0.1/15A507
    path: /usr/bin/xcodebuild
Languages:
  Ruby:
    version: 3.2.2
    path: /Users/ryan/.asdf/shims/ruby
npmPackages:
  "@react-native-community/cli": Not Found
  react:
    installed: 18.2.0
    wanted: 18.2.0
  react-native:
    installed: 0.72.6
    wanted: 0.72.6
  react-native-macos:
    installed: 0.71.33
    wanted: ^0.71.0-0
npmGlobalPackages:
  "*react-native*": Not Found
Android:
  hermesEnabled: true
  newArchEnabled: false
iOS:
  hermesEnabled: true
  newArchEnabled: Not found
```

Library version: main

## Steps To Reproduce

Issues without reproduction steps or code are likely to stall.

1. git clone https://github.com/software-mansion/react-native-svg
2. cd react-native-svg
3. yarn
[...]

PR #2341, opened 2024-07-10
fix: macOS new architecture build

bohdanprog:
# Summary
Based on recommendations from this [proposition](https://github.com/software-mansion/react-native-svg/issues/2192#issuecomment-2177330499) and added some changes now we can build macOS using the new architecture.

## Compatibility

| OS      | Implemented |
| ------- | :---------: |
| macOS     |    ✅     |


apple/RNSVGUIKit.h
```diff
@@ -8,7 +11,11 @@
 #define RNSVGColor UIColor
 #define RNSVGPlatformView UIView
 #define RNSVGTextView UILabel
+#ifdef RCT_NEW_ARCH_ENABLED
+#define RNSVGView RCTViewComponentView
```
WoLewicki, line 15: Did you test it on new arch on `iOS`? What about e.g. such code: https://github.com/software-mansion/react-native-svg/blob/3b360a2b58a6bbc9541f81ee6c7150ca316d583f/apple/ViewManagers/RNSVGSvgViewManager.mm#L16, should we expect `RCTViewComponentView` there? Shouldn't it be rather `UIView` ? Can we narrow the class this way?

bohdanprog: As I understand we need _props value from RCTViewComponentView, because it does not exist in UIView, and we use that in RNSVGPattern.mm and RNSVGPath.mm.

WoLewicki: Why do we need `_props` in the file I linked?

WoLewicki: I mean there was always just `UIView` passed there: https://github.com/software-mansion/react-native-screens/blob/2cd05a8e39b81ec0458aba9e733c5a8a95a2d1d6/ios/RNSScreen.mm#L1489 and right now it will get `RCTViewComponentView` if I understand the code correctly, yeah?

bohdanprog: [That](https://github.com/software-mansion/react-native-svg/pull/2341/files#diff-f69c74ee636ed35a8d1db786a7f2446e211e3566a7155bbbedb3b61d1ea31864R16) changes should answer your question. Now we specify the returned type from view.
here:
```
@interface RNSVGView :
#ifdef RCT_NEW_ARCH_ENABLED
    RCTViewComponentView
#else
    RCTUIView
#endif // RCT_NEW_ARCH_ENABLED
```
- macOS:
 Fabric: `RCTViewComponentView` extends from  `RCTUIView` -> `RCTPlatformView` -> `RCTPlatformView NSView`.
Paper: `RCTUIView` extends from `RCTPlatformView` -> -> `RCTPlatformView NSView`.

- IOS:
```
#ifdef RCT_NEW_ARCH_ENABLED
#define RNSVGView RCTViewComponentView
#else
#define RNSVGView UIView
#endif // RCT_NEW_ARCH_ENABLED
```
Fabric: `RCTViewComponentView` extends from `UIView`
Paper: `UIView`

WoLewicki: Here is an article that shows the problem: https://en.wikipedia.org/wiki/Covariance_and_contravariance_(computer_science). E.g. here: https://github.com/software-mansion/react-native-svg/blob/26c4c73400d36513de61f00b7297104c6c7d960d/apple/Elements/RNSVGClipPath.mm#L63 you will now assume on new arch that `self.subviews` has elements of kind `RCTViewComponentView` which is certainly not true and it seems weird that it even compiles.

WoLewicki (approved):
LGTM!

.github/workflows/macos-build-fabric-test.yml
```diff
@@ -0,0 +1,21 @@
...
+      - 'src/fabric/**'
+      - 'package.json'
+  push:
+    branches:
+      - main
+
+jobs:
+  test-build:
+    if: github.repository == 'software-mansion/react-native-svg'
+    uses: software-mansion/react-native-svg/.github/workflows/reusable-macos-build.yml
```
CDFN, line 19:
```suggestion
    uses: software-mansion/react-native-svg/.github/workflows/reusable-macos-build.yml@main
```

WoLewicki (approved):
LGTM!

PR #2345, opened 2024-07-11
fix: mask and filters rendering issues

jakex7:
# Summary

Fixed rendering issues on macOS and iOS after merging Mask and Filter changes together.

## Test Plan

Example app


WoLewicki (approved):
LGTM!

PR #2346, opened 2024-07-12
feat: support filters on web

jakex7:
# Summary

* Export `Filter` and `FeColorMatrix` components on `web`
* Change filter IDs in example to be unique*
* Generate filter ID when using `FilterImage`*
* Hide `FilterImage` example on web, since it's crashing the whole site (see #2334)

\* ID on web has to be unique, otherwise it'll use the first element with that ID, even if they are in the separate SVG elements

## Test Plan

run `web-example` app

## Compatibility

| OS      | Implemented |
| ------- | :---------: |
| Web     |    ✅     |


WoLewicki (commented):
Should we wait for https://github.com/software-mansion/react-native-svg/pull/2334 with merging this one?

jakex7:
> Should we wait for #2334 with merging this one?

@WoLewicki 
We don't have to. It's temporary disabling `FilterImage` example on web so everything else should work.

src/filter-image/FilterImage.tsx
```diff
@@ -19,13 +26,32 @@ const getFilters = (filters: FilterImageProps['filters']) => {
...
+        ';',
+      ''
+    )
+  );
+};
+
 export const FilterImage = (props: FilterImageProps) => {
   const { source, style, ...imageProps } = props;
+  const hasFilter = props.filters.length > 0;
+  const filterId = useMemo(() => getFilterName(props.filters), [props.filters]);
```
WoLewicki, line 46: What if all the filters are the same on both images? Wouldn't it be better to just use some hash and add it?

jakex7: Sure, any suggestions on how to get a relatively unique hash? I'm afraid that `Math.random` won't be sufficient.

jakex7: Done in [26cbf53](https://github.com/software-mansion/react-native-svg/pull/2346/commits/26cbf53ee2f9566c98eb2d8afc729eb4d9a16290)

src/filter-image/FilterImage.tsx
```diff
@@ -20,12 +29,17 @@ const getFilters = (filters: FilterImageProps['filters']) => {
 };
 
 export const FilterImage = (props: FilterImageProps) => {
-  const { source, style, ...imageProps } = props;
+  const { filters, source, style, ...imageProps } = props;
+  const hasFilter = filters.length > 0;
+  const filterId = useMemo(() => `RNSVG-${getRandomNumber()}`, [filters]);
```
WoLewicki, line 34: Do we need to change the `filterId` when the `filters` change?

WoLewicki (approved):
Added one comment. Please answer it and then I think we are good to go 🚀

src/filter-image/FilterImage.tsx
jakex7: After some tests, we don't. Removed in https://github.com/software-mansion/react-native-svg/pull/2346/commits/af12b62af91bc39b87f1b82630eb08233793f657

PR #2372, opened 2024-07-29
fix: add new macos examples

bohdanprog:
# Summary
We want to avoid problems with GitHub workflows.
Create a new Macos example for new and old architecture.

## Test Plan

Run GitHub workflows

## Compatibility

| OS      | Implemented |
| ------- | :---------: |
| macOS     |    ✅     |


src/ReactNativeSVG.web.ts
```diff
@@ -156,7 +158,7 @@ function parseTransformProp(
       transformArray.push(`matrix(${transform.join(' ')})`);
     } else {
       const stringifiedProps = transformsArrayToProps(
-        transform as TransformsStyle['transform']
+        transform as TransformsStyleArray
```
bohdanprog, line 161: The `Test JS build / build (pull_request)` build crashed without altering that type.

src/ReactNativeSVG.web.ts
```diff
@@ -40,7 +39,10 @@ import type {
 } from './lib/extract/types';
 import SvgTouchableMixin from './lib/SvgTouchableMixin';
 import { resolve } from './lib/resolve';
-import { transformsArrayToProps } from './lib/extract/extractTransform';
+import {
+  transformsArrayToProps,
+  TransformsStyleArray,
```
WoLewicki, line 44: If it is a type, please import it separately in `import type` style. Also, if it fails, I guess it also fails on `main` yes? If so, please submit another PR fixing it since it does not seem to be logically part of this PR.

WoLewicki (approved):
If it works, then I'm fine with it for now :D

PR #2381, opened 2024-07-30
fix: circular-dependencies

bohdanprog:
# Summary
- Require cycles are allowed, but can result in uninitialized values. Consider refactoring to remove the need for a cycle.
- extract SVG web components.
- extract web types.
- extract utils function.

## Test Plan
You can test that problem by running a test example `Test1813`

Without those changes, you can occur that problem by running a test example `Test1813`

## Compatibility

| OS      | Implemented |
| ------- | :---------: |
| iOS     |    ✅     |
| Android |    ✅     |


jakex7 (approved):
LGTM

PR #2382, opened 2024-07-30
feat: add  web implementation SvgXml and others component

bohdanprog:
# Summary
We want to make available SvgXml and other components in the web platform.

## Test Plan
We can easily check how that works by opening `Test1813`.

## Compatibility

| OS      | Implemented |
| ------- | :---------: |
| Web     |    ✅     |


src/css/LocalSvg.tsx
```diff
@@ -39,7 +40,7 @@ export async function loadAndroidRawResource(uri: string) {
 
 export function loadLocalRawResourceAndroid(source: ImageSourcePropType) {
   const uri = getUriFromSource(source);
-  if (isUriAnAndroidResourceIdentifier(uri)) {
+  if (uri && isUriAnAndroidResourceIdentifier(uri)) {
     return loadAndroidRawResource(uri);
   } else {
     return fetchText(uri);
```
jakex7, line 46: If there's no URI, `fetchText` should be executed?

src/xml.tsx
```diff
@@ -78,7 +78,10 @@ export function SvgXml(props: XmlProps) {
   }
 }
 
-export async function fetchText(uri: string) {
+export async function fetchText(uri?: string) {
+  if (!uri) {
+    return null;
+  }
```
jakex7, line 84: I believe we should check for that earlier

jakex7 (commented):
Left some comments, please answer them

src/css/LocalSvg.tsx
bohdanprog: In that case, when I updated the getUriFromSource [function](https://github.com/software-mansion/react-native-svg/pull/2382/files#diff-e08238a7634331fb1f865645ff2ff331b68fe550dfeb2ee36971661c92a757a3R8-R13), it is required to add a check on line 43 to ensure that the URI exists for some reason.

src/xml.tsx
bohdanprog: I believe it's better to check the URI only in that specific place, especially if we have multiple areas where we utilize that function.

jakex7 (approved):

PR #2385, opened 2024-07-31
fix: mac os crashes

jakex7:
# Summary

Fix some crashes on MacOS
* patch `reanimated`
* update fabric type
* invert axis in `FeOffset`


WoLewicki (approved):
LGTM

PR #2388, opened 2024-08-01
fix: FeGaussianBlur stdDeviation on Apple

jakex7:
Scale stdDeviation by screenScale (density) on Apple devices in order to achive the same results as on web

### DIFF
--- a/apps/examples/src/examples/Filters/FeGaussianBlur.tsx
+++ b/apps/examples/src/examples/Filters/FeGaussianBlur.tsx
@@ -1,7 +1,7 @@
 import React, {Component} from 'react';
 import {Circle, FeGaussianBlur, Filter, G, Svg} from 'react-native-svg';
 
-class StdDeviation5Example extends Component {
+class StdDeviation3Example extends Component {
   static title = 'stdDeviation="3"';
   render() {
     return (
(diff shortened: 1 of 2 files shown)


apps/examples/src/examples/Filters/FeGaussianBlur.tsx
```diff
@@ -2,12 +2,12 @@ import React, {Component} from 'react';
 import {Circle, FeGaussianBlur, Filter, G, Svg} from 'react-native-svg';
 
 class StdDeviation5Example extends Component {
-  static title = 'stdDeviation="5"';
+  static title = 'stdDeviation="3"';
```
WoLewicki, line 5: Example titles don't seem to be correct right now. Also does `Android` work ootb?

jakex7: Fixed in [ff942ab](https://github.com/software-mansion/react-native-svg/pull/2388/commits/ff942abcb8b4d68baf60d16376d617179674ea5b)

jakex7: Android blur is limited by the platform and is match weaker anyway. With the full support PR it should be the same as on other platforms.

PR #2393, opened 2024-08-01
fix: apple empty filter results

jakex7:
# Summary

When there is no content of CGImage, ref will return `nil` instead of empty image and crash, we should ignore that results and continue applying filters.

## Example 
```tsx
<svg width="200" height="200">
  <filter id="offset">
    <feOffset dx="1000" dy="1000" />
  </filter>
  <rect
    x="0"
    y="0"
    width="100"
    height="100"
    fill="green"
    filter="url(#offset)" />
</svg>
```


WoLewicki (approved):
I am not aware what exactly it is fixing but I trust you it is needed 😅

Issue #2239, opened 2024-03-05
Svg Image pointerEvents="none" not working on Android


lorenzo-ange:
In an SVG with a transparent image I set pointerEvents="none" to allow another Image to receive the gesture events.
It is working fine on Web and on iOS. 
It is not working on Android.

Here is a simple snack working fine on web and iOS and not working on Android that shows the problem.
In the snack if you tap on the smiling face on Android the gesture event is not firing.

[https://snack.expo.dev/@lorenzoangelini/rn-svg-gesture-image-gesture-swallowing](https://snack.expo.dev/@lorenzoangelini/rn-svg-gesture-image-gesture-swallowing)

Thank you, I am at your disposal for any need for clarification.

GrinZero:
My performance here is exactly the opposite of yours. I set pointerEvents to have no effect on IOS, but it works on other devices

bohdanprog:
Hi @lorenzo-ange, @GrinZero,
The issue is related to React Native. My colleague fixed it. Here is the [PR](https://github.com/facebook/react-native/pull/45975) that resolves the problem.

kacperzolkiewski:
Hi @lorenzo-ange,
This issue has been resolved by the React Native fix in https://github.com/facebook/react-native/pull/45975. I can no longer reproduce the problem.
If you still encounter it or notice anything I might have missed feel free to comment here and we can reopen the issue.

PR #2395, opened 2024-08-02
fix: pointer events getter

coado:
# Summary

Regarding this [issue](https://github.com/software-mansion/react-native-gesture-handler/issues/2793), the PR adds a getter for `mPointerEvents` in `VirtualView` class. Thanks to this, the pointer event type can be accessed from the RNSVG components in [react-native-gesture-handler](https://github.com/software-mansion/react-native-gesture-handler/blob/main/android/src/main/java/com/swmansion/gesturehandler/react/RNViewConfigurationHelper.kt#L15-L18).

## Test Plan

You can test it by changing the `pointerEvents` value and checking if the 'tap' is logged when tapping on the red square:
```ts
import React from 'react';
import { SafeAreaView, StyleSheet } from 'react-native';
import Svg, {  Rect } from 'react-native-svg';
import { Gesture, GestureDetector, GestureHandlerRootView } from 'react-native-gesture-handler';

  const tap = Gesture.Tap().onBegin((event) => console.log('tap'));

  return (
    <GestureHandlerRootView>
      <SafeAreaView style={styles.container}>
        <Svg
          xmlns="http://www.w3.org/2000/svg"
          xmlnsXlink="http://www.w3.org/1999/xlink"
          xmlSpace="preserve"
          viewBox="0 0 382.7 382.7"
          id="container">
          <GestureDetector gesture={tap}>
            <Rect
                width={90.3}
                height={90.3}
                x={242.7}
                y={242.7}
                fill="red"
                strokeWidth={1}
            />
          </GestureDetector>

        <Rect
            width={375}
            height={374}
            fill="none"
            stroke="#000"
            strokeWidth={1}
            pointerEvents="none" 
          />
        </Svg>
      </SafeAreaView>
    </GestureHandlerRootView>
  );
}

  container: {
    flex: 1,
    justifyContent: 'center',
    backgroundColor: '#ecf0f1',
    padding: 8,
  },
});
```

### DIFF
--- a/android/src/main/java/com/horcrux/svg/VirtualView.java
+++ b/android/src/main/java/com/horcrux/svg/VirtualView.java
@@ -102,10 +102,14 @@ public abstract class VirtualView extends ReactViewGroup {
   Region mStrokeRegion;
   Region mClipRegion;
   ArrayList<PathElement> elements;
-  PointerEvents mPointerEvents;
+  PointerEvents mPointerEvents = PointerEvents.AUTO;
 
   void setPointerEvents(PointerEvents pointerEvents) {
     mPointerEvents = pointerEvents;
+  }
+
+  public PointerEvents getPointerEvents() {
+    return this.mPointerEvents;
   }
 
   @Override


WoLewicki (approved):
LGTM

jakex7 (approved):
I wonder if we should move `mPointerEvents` to `RenderableView` as `VirtualView` is also extended by `DefinitionView` that shouldn't be pressable?

bohdanprog:
Related [issue](https://github.com/software-mansion/react-native-svg/issues/2239). to that problem.

kacperzolkiewski:
I close this PR due to the fact that it was fixed by @jakex7 from React Native side [here](https://github.com/facebook/react-native/pull/45975)

PR #2396, opened 2024-08-05
feat: add support dataUri for android platform

bohdanprog:
# Summary

Add support data URI in the Android platform. It already works on the iOS platform.

| Before1 | Before2 | After |
|--------|--------|--------|
| |   |  |

## Compatibility

| OS      | Implemented |
| ------- | :---------: |
| Android |    ✅      |
| iOS		 |    ✅      |

### DIFF
--- a/apps/test-examples/src/Test1318.tsx
+++ b/apps/test-examples/src/Test1318.tsx
@@ -1,0 +1,52 @@
+import React from 'react';
+import {SvgUri} from 'react-native-svg';
+import {SvgCssUri} from 'react-native-svg/css';
+
+export default function Test() {
+  return (
+    <>
+      <SvgCssUri
+        height={100}
+        width={100}
+        uri={
+          'data:image/svg+xml;utf8,%3Csvg%20xmlns%3D%22http%3A%2F%2Fwww.w3.org%2F2000%2Fsvg%22%20viewBox%3D%220%200%20100%20100%22%3E%3Cpath%20fill%3D%22%2354b847%22%20d%3D%22M0%200h100v100H0z%22%2F%3E%3Cpath%20fill%3D%22%23fff%22%20d%3D%22M61.6%2069.1h-17V38.3h8.6v25.4h8.2c5.7%200%208.1-3.9%208.1-14.1S66.8%2036.4%2061%2036.4H42.2v32.7h-8.4V36.3H21.4v-5.4h42.3c10.2%200%2015%205.3%2015%2018.6-.1%2017.4-7.3%2019.6-17.1%2019.6%22%2F%3E%3C%2Fsvg%3E'
+        }
+      />
+      <SvgCssUri
+        height={100}
+        width={100}
+        uri={
+          'data:image/svg+xml;utf8,%3Csvg%20width%3D%22300%22%20height%3D%22700%22%20viewBox%3D%2250%2050%20100%20100%22%20preserveAspectRatio%3D%22xMidYMid%20meet%22%3E%3Crect%20x%3D%2250%22%20y%3D%2250%22%20width%3D%22100%22%20height%3D%22100%22%20fill%3D%22yellow%22%3E%3C%2Frect%3E%3C%2Fsvg%3E'
+        }
+      />
+      <SvgUri
+        height={100}
+        width={100}
+        uri={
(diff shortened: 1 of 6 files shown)


apps/test-examples/src/Test1318.tsx
```diff
@@ -0,0 +1,52 @@
...
+          'data:image/svg+xml;utf8,%3Csvg%20width%3D%22300%22%20height%3D%22700%22%20viewBox%3D%2250%2050%20100%20100%22%20preserveAspectRatio%3D%22xMidYMid%20meet%22%3E%3Crect%20x%3D%2250%22%20y%3D%2250%22%20width%3D%22100%22%20height%3D%22100%22%20fill%3D%22yellow%22%3E%3C%2Frect%3E%3C%2Fsvg%3E'
+        }
+      />
+      <SvgUri
+        height={100}
+        width={100}
+        uri={
+          'data:image/svg+xml;utf8,%3Csvg%20xmlns%3D%22http%3A%2F%2Fwww.w3.org%2F2000%2Fsvg%22%20width%3D%22100%22%20height%3D%22100%22%20viewBox%3D%220%200%20100%20100%22%3E%3Cpolygon%20points%3D%2250%2C15%2061%2C40%2088%2C40%2066%2C57%2075%2C84%2050%2C68%2025%2C84%2034%2C57%2012%2C40%2039%2C40%22%20fill%3D%22gold%22%2F%3E%3C%2Fsvg%3E'
+        }
+      />
```
jakex7, line 49: Please, deduplicate examples

src/utils/fetchData.ts
```diff
@@ -0,0 +1,53 @@
...
+
+  // get the encoded data portion and decode URI-encoded chars
+  const data = unescape(uri.substring(firstComma + 1));
+  if (!base64) {
+    return data;
+  }
+  return Buffer.from(data).toString('base64');
+}
+
+async function localFetch(uri?: string) {
```
jakex7, line 44: It's not localFetch, please give it more appropriate name

src/utils/fetchData.ts
```diff
@@ -0,0 +1,53 @@
+import { Buffer } from 'buffer';
+
+export function fetchText(uri?: string): Promise<string | null> {
+  if (uri && uri.startsWith('data:')) {
+    return new Promise((resolve) => {
+      resolve(dataUriToXml(uri));
+    });
+  } else {
+    return localFetch(uri);
+  }
```
jakex7, line 10: Can we use that flow exclusively on Android since it's the only platform that wasn't working?

src/utils/fetchData.ts
```diff
@@ -0,0 +1,53 @@
+import { Buffer } from 'buffer';
+
+export function fetchText(uri?: string): Promise<string | null> {
+  if (uri && uri.startsWith('data:')) {
```
jakex7, line 4: Quick question: Can base64 file be fetched from remote? If yes, then it still would not work since it's base64 or fetch

src/utils/fetchData.ts
bohdanprog: I included a check for the Android platform and enhanced the validation for the URI path.

src/utils/fetchData.ts
bohdanprog: changed.

apps/test-examples/src/Test1318.tsx
bohdanprog: done.

src/utils/fetchData.ts
bohdanprog: It has been updated.

src/utils/fetchData.ts
```diff
@@ -0,0 +1,37 @@
+import { Platform } from 'react-native';
+
+export function fetchText(uri?: string): Promise<string | null> {
+  if (
+    uri &&
```
jakex7, line 5: Could we make a guard here? If both functions `dataUriToXml` and `fetchUriData` expect string, then we can add this here
```suggestion
  if (!uri) {
    return null;
  }
```
and then remove the check from `fetchUriData`

src/utils/fetchData.ts
```diff
@@ -0,0 +1,37 @@
...
+  }
+}
+
+function dataUriToXml(uri: string): string | null {
+  try {
+    // decode and remove data:image/svg+xml;utf8, prefix
+    const xml = decodeURIComponent(uri).split(',').slice(1).join(',');
+    return xml;
+  } catch (error) {
+    console.log('error', error);
```
jakex7, line 23: We probably should throw an `Error` here instead of `log`. Also, it should be more descriptive like here https://github.com/software-mansion/react-native-svg/pull/2396/files#diff-1d505e3c2299d034f80dac76837cf7c866616fd1a8c2361295d141833b104f46R36

src/utils/fetchData.ts
bohdanprog: Good point, thanks.

src/utils/fetchData.ts
bohdanprog: done.

src/utils/fetchData.ts
bohdanprog: done

jakex7 (approved):
LGTM

PR #2397, opened 2024-08-06
feat: add support int32 color

bohdanprog:
# Summary
When we use reanimated it's important to use createAnimatedPropAdapter like in that [example](https://docs.swmansion.com/react-native-reanimated/docs/core/useAnimatedProps/#adapters-), but we didn't have support int32Color on the Web platform.

## Test Plan

When we run that [example](https://snack.expo.dev/@bohdan.artiukhov/animation-example) that should work on all platforms.

## Compatibility

| OS      | Implemented |
| ------- | :---------: |
| iOS       |    	✅      	|
| MacOS |   ✅      	|
| Android |    ✅	  	|
| Web      |    ✅		  	|


apps/test-examples/App.js
```diff
@@ -12,6 +12,7 @@ import Test1790 from './src/Test1790';
 import Test1813 from './src/Test1813';
 import Test1845 from './src/Test1845';
 import Test1986 from './src/Test1986';
+import Test2054 from './src/Test2054';
```
jakex7, line 15: Please, change the test number to this PR or corresponding issue

jakex7 (approved):
LGTM

apps/test-examples/App.js
bohdanprog: Test number and order of imports have been updated.

PR #2399, opened 2024-08-08
feat: calculate width from height

### LINKED ISSUE #2138: SVG doesn't calculate width from height+viewbox (web/native inconsistency)
Library version: 13.4.0

Snack link: https://snack.expo.dev/@chris_whiterosemaths/svgs-natural-aspect-ratio

I have an SVG defined with a `viewbox` (which I think should give it a "natural aspect ratio" (using CSS terminology)).

```tsx
function Shape(props) {
  return <Svg
    viewBox="0 0 24 24"
    fill="black"
    {...props}
  >
    <Path d="...snip..." />
  </Svg>
}
```

Elsewhere, when using it, I provide the `height` but not the `width`.

```tsx
<Shape height={100} />
```

### Expected result

I would like this to mean that the SVG now has a concrete height _and width_, since the width can be calculated from the height and the aspect ratio. This behaviour works on web. See the snack link.

### Actual result

However, on native with this library, the SVG isn't visible. I think this is because the width is being assumed as 0, but I'm not sure. If I give it a width, all is well

### Why this is a problem / Use case

I'm using react-native-svg-transformer to auto-generate the Svg component from a .svg file. The .svg file has the viewbox defined in it, but no concrete width or height. There is no way to access that viewport, so I can't do my own calculations at runtime.

I would like to place several SVGs in a row within a View. I basically just want to fit all the SVGs in. I do that by giving each a height equal to the height of the row, and allowing their width to be whatever it needs to be for that height. As far as I can tell there's no way to allow an SVGs width/height to be based off the viewbox's aspect ratio on native, but this is how it's meant to work (on web)!

preserveAspectRatio, whilst amazing, isn't helpful here because I want each SVG to take up as much width as it needs to. I don't want to put it in a container and have it scale to fit that container.

bohdanprog:
# Summary
Closes #2138 

We want to calculate width and height based on the aspect ratio defined by the `viewBox`.

## Test Plan

We can easily test my solution when we run the `Test2138`

## Compatibility

| OS      | Implemented |
| ------- | :---------: |
| iOS     |    ✅      |
| Android |    ✅      |


src/elements/Svg.tsx
```diff
@@ -125,6 +124,31 @@ export default class Svg extends Shape<SvgProps> {
...
+
+    if (viewBox) {
+      const viewBoxValues = viewBox.split(' ');
+      if (viewBoxValues.length === 4) {
+        const viewBoxWidth = parseFloat(viewBoxValues[2]);
+        const viewBoxHeight = parseFloat(viewBoxValues[3]);
+        if (!isNaN(viewBoxWidth) && !isNaN(viewBoxHeight)) {
+          if (width !== undefined && height === undefined) {
+            height = isWidthPercentage
+              ? `${(parseFloat(width) * viewBoxHeight) / viewBoxWidth}%`
```
jakex7, line 141: I dont think that's how `%` value should be calculated. For example, let's take that screen width is `200` and height is `600`. Now we have SVG with VB proportion `1` (like `100/100`) and width `50%`. Height is expected to be the same as width while in this implementation it will end up as `50%` of the height which is `300`, not `100`

I don't think the `%` value is being calculated correctly. For instance, if the screen width is `200` and the height is `600`, and we have an SVG with a viewBox proportion of `1` (such as `100/100`) and a width set to `50%`, the height should match the width. However, in this implementation, the height ends up being `50%` of the screen height, which is `300`, rather than the expected `100`.

PR #2426, opened 2024-08-23
feat: add un-escape special symbols functionality

### LINKED ISSUE #2411: SvgUri/SvgFromXml components don't un-escape special symbols
### Description 
The `SvgUri` and its underlying `SvgFromXml` component do not un-escape text from the <text> SVG element. It prevents displaying special characters (such as `<`, `>`, `&`, and others) in SVG text elements.

### Example
Source SVG string, retrieved from remote URL:
```svg
<svg>
    <text x="10" y="32" fontSize="32">&amp; &lt; &gt;</text>
</svg>
```

React Native SVG component example:

```jsx
import { SafeAreaView, StyleSheet } from 'react-native';
import { SvgFromXml } from "react-native-svg";

  return (
    <SafeAreaView>
      <SvgFromXml
        xml={`<svg><text x="10" y="32" fontSize="32">&amp; &lt; &gt;</text></svg>`}
      />
    </SafeAreaView>
  );
}
```

### Result

| IOS | Android | Web |
| --- | --- | --- |
| Screenshot 2024-08-14 at 13 06 28 | Screenshot 2024-08-14 at 13 08 13 | Screenshot 2024-08-14 at 13 09 13 |

### Expected behavior

The HTML special characters should be un-escaped inside the <text> tag at the SVG rendering time, as it works on a web browser.

### Steps to reproduce

Can be reproduced using a base React Native or Expo project.

### Snack or a link to a repository

https://snack.expo.dev/@danilademidovich/23ac00?platform=ios

### SVG version

15.2.0

### React Native version

0.74.1

### Platforms

iOS

### JavaScript runtime

Hermes

### Workflow

Expo Go

### Architecture

None

### Build type

Release app & dev bundle

### Device

None

### Device model

iPhone 14 Pro Max

### Acknowledgements

Yes

bohdanprog:
# Summary
Fixes #2411
Motivation: the parse function does not un-escape text from the SVG element. It prevents displaying special characters (such as <, >, &, and others) in SVG text elements.

## Test Plan
The example is available in the test-examples app as Test2411

## Compatibility

| OS      | Implemented |
| ------- | :---------: |
| iOS     |    ✅      |
| MacOS   |    ✅      |
| Android |    ✅     |
| Web     |    ✅      |

### DIFF
--- a/src/xml.tsx
+++ b/src/xml.tsx
@@ -272,6 +272,14 @@ export function parse(source: string, middleware?: Middleware): JsxAST | null {
     );
   }
 
+  // New function to decode HTML entities
+  function decodeEntities(text: string): string {
+    return text
+      .replace(/&amp;/g, '&')
+      .replace(/&lt;/g, '<')
+      .replace(/&gt;/g, '>');
+  }
+
   function metadata() {
     while (
       i + 1 < length &&
@@ -296,7 +304,7 @@ export function parse(source: string, middleware?: Middleware): JsxAST | null {
     }
 
     if (/\S/.test(text)) {
-      children.push(text);
+      children.push(decodeEntities(text)); // Decode entities before pushing text to children
     }
 
     if (source[i] === '<') {
@@ -391,7 +399,7 @@ export function parse(source: string, middleware?: Middleware): JsxAST | null {
       error('expected ]]>');
     }
 
-    children.push(source.slice(i + 7, index));
+    children.push(decodeEntities(source.slice(i + 7, index))); // Decode entities in CDATA section
 
     i = index + 2;
     return neutral;
@@ -457,7 +465,7 @@ export function parse(source: string, middleware?: Middleware): JsxAST | null {
         i += 1;
         allowSpaces();
 
-        value = getAttributeValue();
+        value = decodeEntities(getAttributeValue()); // Decode entities in attribute values
         if (!isNaN(+value) && value.trim() !== '') {
           value = +value;
         }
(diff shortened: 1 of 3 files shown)


src/xml.tsx
```diff
@@ -272,6 +272,13 @@ export function parse(source: string, middleware?: Middleware): JsxAST | null {
     );
   }
 
+  function decodeEntities(text: string): string {
+    return text
+      .replace(/&amp;/g, '&')
+      .replace(/&lt;/g, '<')
+      .replace(/&gt;/g, '>');
```
jakex7, line 279: Making an exception for some character just seems wrong. What about `&apos;` or `&quot;`? There are dozens of characters that could be escaped.

jakex7:
> It prevents displaying special characters (such as <, >, &, and others) in SVG text elements.

It seems to me that it is doing the opposite. With this function we will be able to display characters such as `< / > / &`.

jakex7 (commented):
I left my comment. Maybe it fixes this specific issue, but we should consider the broader context.

src/xml.tsx
jakex7, line 279: Additionally, what about numeric character references? According to the spec, it's also a valid text content. Example:`&lt;` can be represented as `&#60;`, `&gt;` as `&#62;` and `&amp;` as `&#38;`

bohdanprog:
I guess we have to find another solution for that, that's why I closed that PR.

PR #2431, opened 2024-08-28
feat: add format and quality in toDataURL function

bohdanprog:
# Summary
Introduce new parameters, format, and quality, for the toDataURL function.

Allow users to select the compressed format (defaulting to PNG or JPEG) in the toDataURL function. Additionally, include a new parameter called quality for the JPEG format.
Created a new type, ToDataUrlOptions, and exported it externally.

The data structure appears as follows:
```ts
  const optionsWithJPEGFormat: ToDataUrlOptions = {
    format: 'jpeg',
    width,
    height,
    quality: 100,
  };
```

```ts
  const optionsWithPNGFormat: ToDataUrlOptions = {
    format: 'png',
    width,
    height,
  };
```

## Compatibility

| OS      | Implemented |
| ------- | :---------: |
| iOS     |    ✅      |
| MacOS   |    ✅      |
| Android |    ✅      |
| Web     |    ✅❌      |

## Checklist

### DIFF
--- a/android/src/main/java/com/horcrux/svg/SvgView.java
+++ b/android/src/main/java/com/horcrux/svg/SvgView.java
@@ -356,7 +356,7 @@ public class SvgView extends ReactViewGroup implements ReactCompoundView, ReactC
     return Base64.encodeToString(bitmapBytes, Base64.NO_WRAP);
   }
 
-  String toDataURL(int width, int height) {
+  String toDataURL(int width, int height, String format, Integer quality) {
     Bitmap bitmap = Bitmap.createBitmap(width, height, Bitmap.Config.ARGB_8888);
 
     clearChildCache();
@@ -364,7 +364,12 @@ public class SvgView extends ReactViewGroup implements ReactCompoundView, ReactC
     clearChildCache();
     this.invalidate();
     ByteArrayOutputStream stream = new ByteArrayOutputStream();
-    bitmap.compress(Bitmap.CompressFormat.PNG, 100, stream);
+    if (format.equals("jpeg")) {
+      int checkQualityValue = (quality != null) ? quality : 100;
+      bitmap.compress(Bitmap.CompressFormat.JPEG, checkQualityValue, stream);
+    } else {
+      bitmap.compress(Bitmap.CompressFormat.PNG, 100, stream);
+    }
     bitmap.recycle();
     byte[] bitmapBytes = stream.toByteArray();
     return Base64.encodeToString(bitmapBytes, Base64.NO_WRAP);
--- a/android/src/main/java/com/horcrux/svg/SvgViewModule.java
+++ b/android/src/main/java/com/horcrux/svg/SvgViewModule.java
@@ -67,8 +67,10 @@ class SvgViewModule extends NativeSvgViewModuleSpec {
                   });
             } else {
               if (options != null) {
+                  Integer quality = options.hasKey("quality") ? options.getInt("quality") : null;
+                  String format = options.hasKey("format") ? options.getString("format") : null;
                 successCallback.invoke(
-                    svg.toDataURL(options.getInt("width"), options.getInt("height")));
+                    svg.toDataURL(options.getInt("width"), options.getInt("height"), format, quality));
               } else {
                 successCallback.invoke(svg.toDataURL());
               }
--- a/apps/test-examples/src/Test2233.tsx
+++ b/apps/test-examples/src/Test2233.tsx
@@ -5,7 +5,7 @@ import Svg, {Path, type ToDataUrlOptions} from 'react-native-svg';
 const SvgLogoWelcome = () => {
   const ref = React.useRef<Svg | null>(null);
 
-  const {width, height} = Dimensions.get('window');
+  const {width, height} = Dimensions.get('screen');
   const optionsWithJPEGFormat: ToDataUrlOptions = {
     format: 'jpeg',
     width,
@@ -20,7 +20,13 @@ const SvgLogoWelcome = () => {
   };
 
   return (
-    <View>
+    <View
+      style={{
+        width: '100%',
+        height: '100%',
+        justifyContent: 'center',
+        alignItems: 'center',
+      }}>
       <Button
         onPress={() => {
           ref.current?.toDataURL(base64Image => {
@@ -29,7 +35,7 @@ const SvgLogoWelcome = () => {
         }}
         title="log"
       />
-      <Svg ref={ref} viewBox="0 0 24 24" fill="black">
+      <Svg ref={ref} viewBox="0 0 24 24" fill="black" width={250} height={250}>
         <Path d="M12 2c5.514 0 10 4.486 10 10s-4.486 10-10 10S2 17.514 2 12 6.486 2 12 2zm0-2C5.373 0 0 5.373 0 12s5.373 12 12 12 12-5.373 12-12S18.627 0 12 0zm5.507 13.941c-1.512 1.195-3.174 1.931-5.506 1.931-2.334 0-3.996-.736-5.508-1.931L6 14.434C7.127 16.154 9.2 18 12.001 18c2.8 0 4.872-1.846 5.999-3.566l-.493-.493zM8.5 8a1.5 1.5 0 100 3 1.5 1.5 0 000-3zm7 0a1.5 1.5 0 100 3 1.5 1.5 0 000-3z" />
       </Svg>
     </View>
--- a/apps/test-examples/src/Test2233.tsx
+++ b/apps/test-examples/src/Test2233.tsx
@@ -1,16 +1,30 @@
 import * as React from 'react';
 import {Button, SafeAreaView, View} from 'react-native';
-import Svg, {Path} from 'react-native-svg';
+import Svg, {Path, type ToDataUrlOptions} from 'react-native-svg';
 
 const SvgLogoWelcome = () => {
   const ref = React.useRef<Svg | null>(null);
+
+  const optionsWithJPEGFormat: ToDataUrlOptions = {
+    format: 'jpeg',
+    width: 100,
+    height: 100,
+    quality: 1,
+  };
+
+  const optionsWithPNGFormat: ToDataUrlOptions = {
+    format: 'png',
+    width: 100,
+    height: 100,
+  };
+
   return (
     <View>
       <Button
         onPress={() => {
           ref.current?.toDataURL(base64Image => {
             console.log(base64Image, 'data');
-          });
+          }, optionsWithJPEGFormat);
         }}
         title="log"
       />
--- a/apps/test-examples/src/Test2233.tsx
+++ b/apps/test-examples/src/Test2233.tsx
@@ -1,21 +1,22 @@
 import * as React from 'react';
-import {Button, SafeAreaView, View} from 'react-native';
+import {Button, Dimensions, SafeAreaView, View} from 'react-native';
 import Svg, {Path, type ToDataUrlOptions} from 'react-native-svg';
 
 const SvgLogoWelcome = () => {
   const ref = React.useRef<Svg | null>(null);
 
+  const {width, height} = Dimensions.get('window');
   const optionsWithJPEGFormat: ToDataUrlOptions = {
     format: 'jpeg',
-    width: 100,
-    height: 100,
-    quality: 1,
+    width,
+    height,
+    quality: 100,
   };
 
   const optionsWithPNGFormat: ToDataUrlOptions = {
     format: 'png',
-    width: 100,
-    height: 100,
+    width,
+    height,
   };
 
   return (
--- a/src/elements/Svg.tsx
+++ b/src/elements/Svg.tsx
@@ -45,6 +45,21 @@ export interface SvgProps extends GProps, ViewProps, HitSlop {
   title?: string;
 }
 
+export type ToDataUrlOptions = JpegOptions | PngOptions;
+
+interface JpegOptions {
+  format: 'jpeg';
+  quality?: number; // Quality is optional but only available when format is 'jpeg'
+  width?: number;
+  height?: number;
+}
+
+interface PngOptions {
+  format: 'png';
+  width?: number;
+  height?: number;
+}
+
 export default class Svg extends Shape<SvgProps> {
   static displayName = 'Svg';
 
@@ -81,10 +96,29 @@ export default class Svg extends Shape<SvgProps> {
     root && root.setNativeProps(props);
   };
 
-  toDataURL = (callback: (base64: string) => void, options?: object) => {
+  someFunction(options?: ToDataUrlOptions) {
+    if (options) {
+      if (!options.width && !options.height) {
+        throw new Error(
+          'toDataURL: Please provide width and height in options.'
+        );
+      }
+      if (options.width && !options.height) {
+        options.height = options.width;
+      } else if (!options.width && options.height) {
+        options.width = options.height;
+      }
+    }
+  }
+
+  toDataURL = (
+    callback: (base64: string) => void,
+    options?: ToDataUrlOptions
+  ) => {
     if (!callback) {
       return;
     }
+    this.someFunction(options);
     const handle = findNodeHandle(this.root as Component);
     const RNSVGSvgViewModule: Spec =
       // eslint-disable-next-line @typescript-eslint/no-var-requires
--- a/src/elements/Svg.tsx
+++ b/src/elements/Svg.tsx
@@ -60,6 +60,41 @@ interface PngOptions {
   height?: number;
 }
 
+function checkOptions(options?: ToDataUrlOptions) {
+  if (options) {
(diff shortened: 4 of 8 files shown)


src/elements/Svg.tsx
```diff
@@ -45,6 +45,40 @@ export interface SvgProps extends GProps, ViewProps, HitSlop {
   title?: string;
 }
 
+export type ToDataUrlOptions = JpegOptions | PngOptions;
```
jakex7, line 48:
```suggestion
export type DataUrlOptions = JpegOptions | PngOptions;
```

src/elements.web.ts
```diff
@@ -264,6 +261,14 @@ export class Svg extends WebShape<BaseProps & SvgProps> {
 
     const width = Number(options.width) || rect.width;
     const height = Number(options.height) || rect.height;
+    // Set format, defaulting to 'png'
+    const format = options.format === 'jpeg' ? 'image/jpeg' : 'image/png';
+    // Get quality, only applicable for JPEG
+
+    let quality = 0.92; // Default quality is 0.92 for JPEG
```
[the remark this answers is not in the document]
bohdanprog: I relied on that [answer](https://stackoverflow.com/a/15754534), but I'm unsure if it's correct or not.

apps/test-examples/src/Test2233.tsx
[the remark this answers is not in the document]
bohdanprog: It's a good point, I put it there in case I wanted to check how it worked.

src/elements.web.ts
```diff
@@ -278,7 +283,14 @@ export class Svg extends WebShape<BaseProps & SvgProps> {
...
       const context = canvas.getContext('2d');
       context?.drawImage(img, 0, 0);
-      callback(canvas.toDataURL().replace('data:image/png;base64,', ''));
+      // Get the data URL with the specified format and quality
+      const dataURL =
+        format === 'image/jpeg'
+          ? canvas.toDataURL(format, quality)
+          : canvas.toDataURL(format);
+      // Return the data URL, removing the prefix if it's PNG to match your original implementation
+      const base64Data = dataURL.replace(`data:${format};base64,`, '');
```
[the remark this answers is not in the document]
bohdanprog: my bad

src/elements.web.ts
```diff
@@ -278,7 +283,14 @@ export class Svg extends WebShape<BaseProps & SvgProps> {
       canvas.height = height;
       const context = canvas.getContext('2d');
       context?.drawImage(img, 0, 0);
-      callback(canvas.toDataURL().replace('data:image/png;base64,', ''));
+      // Get the data URL with the specified format and quality
+      const dataURL =
+        format === 'image/jpeg'
+          ? canvas.toDataURL(format, quality)
+          : canvas.toDataURL(format);
```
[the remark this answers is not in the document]
bohdanprog: changed.

src/elements.web.ts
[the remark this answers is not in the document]
bohdanprog: removed.

src/elements/Svg.tsx
bohdanprog: renamed.
Thank you for the suggestion.

android/src/main/java/com/horcrux/svg/SvgView.java
```diff
@@ -356,15 +356,20 @@ String toDataURL() {
...
     Bitmap bitmap = Bitmap.createBitmap(width, height, Bitmap.Config.ARGB_8888);
 
     clearChildCache();
     drawChildren(new Canvas(bitmap));
     clearChildCache();
     this.invalidate();
     ByteArrayOutputStream stream = new ByteArrayOutputStream();
-    bitmap.compress(Bitmap.CompressFormat.PNG, 100, stream);
+    if (format.equals("jpeg")) {
+      int checkQualityValue = (quality != null) ? quality : 100;
```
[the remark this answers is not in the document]
bohdanprog: done

src/elements/Svg.tsx
```diff
@@ -45,6 +45,40 @@ export interface SvgProps extends GProps, ViewProps, HitSlop {
...
+function checkOptions(options?: ToDataUrlOptions) {
+  if (options && options?.format === 'jpeg') {
+    if (!validateJpegQualityParameter(options)) {
+      throw new Error('toDataURL: Invalid quality parameter for JPEG format.');
+    }
+  }
+}
+
+function validateJpegQualityParameter(options: JpegOptions): boolean {
+  if (options.quality && (options.quality < 0.1 || options.quality > 1)) {
```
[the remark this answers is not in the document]
bohdanprog: Yeah, you're right. I update that check.

src/elements.web.ts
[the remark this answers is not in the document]
jakex7: I suppose that's true, but why do we need to specify it? We can leave out that value, and the browser should default to its default value.

src/elements/Svg.tsx
```diff
@@ -45,6 +45,40 @@ export interface SvgProps extends GProps, ViewProps, HitSlop {
...
+  quality?: number;
+  width: NumberProp;
+  height: NumberProp;
+}
+
+interface PngOptions {
+  format: 'png';
+  width: NumberProp;
+  height: NumberProp;
+}
```
jakex7, line 61: `width` and `height` should not be mandatory. User should be able to use 
```ts
ref.toDataURL(someCallback, { format: 'jpeg', quality: 1 });
```
without specifying the width and height

src/elements/Svg.tsx
```diff
@@ -45,6 +45,40 @@ export interface SvgProps extends GProps, ViewProps, HitSlop {
...
+    }
+  }
+}
+
+function validateJpegQualityParameter(options: JpegOptions): boolean {
+  if (options.quality && (options.quality < 0 || options.quality > 1)) {
+    return false;
+  }
+  // Android requires quality to be a number between 0 and 100
+  if (Platform.OS === 'android' && options.quality) {
```
jakex7, line 76: Redundant check. If `options.quality` is a falsy value, the function will `return false` in the previous condition

src/elements/Svg.tsx
```diff
@@ -45,6 +45,40 @@ export interface SvgProps extends GProps, ViewProps, HitSlop {
...
+    options.quality = Math.round(options.quality * 100);
+  }
```
jakex7, line 78: We can multiply that value on the native side to maintain consistency in JS values.

jakex7 (changes requested):
Left some more comments

src/elements/Svg.tsx
bohdanprog: In our [implementation](https://github.com/software-mansion/react-native-svg/pull/2431/files#diff-767ae05982e97c6a98efdf996bcad641585b80c35046ecf9c96c77c3dc7d146bL45), an error occurs when you provide any options but don't provide width and height.
Or should we change the whole implementation?

jakex7: Yes, the current implementation allows using `toDataURL(() => {});` without specifying the size, so we shouldn't require specifying it just because someone want to change the format.

jakex7: I know it can be challenging to handle the types correctly (e.g. making `width` required when `height` is specified and vice versa), so we could deprecate the old options for now and introduce a new `size` property.

bohdanprog: @jakex7 Okay, I thought maybe there would be significant changes and decided not to do that :)
But I'll provide those changes ;)

src/elements/Svg.tsx
bohdanprog: You're right, thank you ;)

src/elements/Svg.tsx
bohdanprog: move that multiply on the native side 👍

src/elements.web.ts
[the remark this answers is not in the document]
bohdanprog: You're right, we can leave out that value ;)

src/elements/Svg.tsx
```diff
@@ -45,6 +45,36 @@ export interface SvgProps extends GProps, ViewProps, HitSlop {
...
+  quality?: number;
+  width?: NumberProp;
+  height?: NumberProp;
+}
+
+interface PngOptions {
+  format: 'png';
+  width?: NumberProp;
+  height?: NumberProp;
+}
```
jakex7, line 61: We should probably require width, when height is passed (and vice versa)

src/elements/Svg.tsx
```diff
@@ -45,6 +45,36 @@ export interface SvgProps extends GProps, ViewProps, HitSlop {
...
+
+function checkOptions(options?: DataUrlOptions) {
```
jakex7, line 63: maybe `validateOptions`?

android/src/main/java/com/horcrux/svg/SvgViewModule.java
```diff
@@ -67,8 +67,9 @@ public void run() {
                   });
             } else {
               if (options != null) {
+
```
jakex7, line 70: empty line

jakex7 (commented):
Left some comments, rest looks good to me.

src/elements/Svg.tsx
bohdanprog: Done

src/elements/Svg.tsx
```diff
@@ -45,6 +45,39 @@ export interface SvgProps extends GProps, ViewProps, HitSlop {
...
+    }
+  }
+}
+
+function validateJpegQualityParameter(options: JpegOptions): boolean {
+  if (options.quality && (options.quality < 0 || options.quality > 1)) {
+    return false;
+  }
+  return true;
+}
```
jakex7, line 79: Can we move it to the lib directory? It's not related to the component itself

src/elements.web.ts
```diff
@@ -262,8 +259,14 @@ export class Svg extends WebShape<BaseProps & SvgProps> {
...
     const rect = getBoundingClientRect(ref);
 
-    const width = Number(options.width) || rect.width;
-    const height = Number(options.height) || rect.height;
+    const width = Number(options?.size?.width) || rect.width;
+    const height = Number(options?.size?.height) || rect.height;
+    const format = options.format === 'jpeg' ? 'image/jpeg' : 'image/png';
+
+    let quality: number | undefined;
+    if (options && options.format === 'jpeg' && options.quality) {
```
jakex7, line 267: Is quality=0 a valid value? If yes, then we should check for `options.quality !== undefined` not for a truthy value.

apps/examples/src/examples/Svg.tsx
```diff
@@ -131,14 +139,27 @@ class SvgNativeMethods extends Component {
...
+    format: 'jpeg',
+    quality: 1,
+    size: {width: 150, height: 100},
+  };
+
+  optionsWithPNGFormat: DataUrlOptions = {
+    format: 'png',
+    size: {width: 150, height: 100},
+  };
+
```
jakex7, line 155: unused values?

src/elements/Svg.tsx
```diff
@@ -81,16 +114,30 @@ export default class Svg extends Shape<SvgProps> {
     root && root.setNativeProps(props);
   };
 
-  toDataURL = (callback: (base64: string) => void, options?: object) => {
+  /**
+   * @deprecated use size property instead
+   * @param callback
+   * @param options as { size: { width: number, height: number }, format: 'jpeg' | 'png', quality?:1 }
```
jakex7, line 120: https://jsdoc.app/tags-callback
callback should be typed

https://jsdoc.app/tags-param

https://jsdoc.app/tags-type
quality is not an explicit `1` value but rather a number in range {0, 1}

android/src/main/java/com/horcrux/svg/SvgView.java
```diff
@@ -361,15 +363,25 @@ String toDataURL() {
     return Base64.encodeToString(bitmapBytes, Base64.NO_WRAP);
   }
 
-  String toDataURL(int width, int height) {
+  String toDataURL(final ReadableMap options) {
+    int width = options.hasKey("width") ? options.getInt("width") : getWidth();
+    int height = options.hasKey("height") ? options.getInt("height") : getHeight();
```
jakex7, line 368: Are you sure it's still working after moving the width/height to `size` property?

apps/examples/src/examples/Svg.tsx
bohdanprog: I wanted to leave that as an example, or is it better to remove that?

jakex7: It does not show anything when it's not used. Let's remove that for now and just add a simple inline object to `toDataUri` function execution below.

src/elements.web.ts
bohdanprog: @jakex7 Don't we need to check the null variable?

android/src/main/java/com/horcrux/svg/SvgView.java
bohdanprog: Already fixed, that exception.

src/elements.web.ts
bohdanprog: updated that check value.

src/elements/Svg.tsx
bohdanprog: updated JSdoc

src/elements/Svg.tsx
```diff
@@ -81,16 +86,46 @@ export default class Svg extends Shape<SvgProps> {
     root && root.setNativeProps(props);
   };
 
-  toDataURL = (callback: (base64: string) => void, options?: object) => {
+  /**
+   * @deprecated use size property instead.
+   * @description Converts the SVG component to a data URL in either JPEG or PNG format.
+   * @param {function(string): void} callback - A function that receives the base64-encoded data URL string.
+   * @param {DataUrlOptions} options - The options to specify the format (JPEG or PNG) and additional parameters.
```
jakex7, line 93: It's a different function. Above JSDoc description is related to deprecated signature:
```ts
toDataURL(callback: (base64: string) => void, options: Size): void;
```

src/elements/Svg.tsx
```diff
@@ -81,16 +86,46 @@ export default class Svg extends Shape<SvgProps> {
...
     const handle = findNodeHandle(this.root as Component);
     const RNSVGSvgViewModule: Spec =
       // eslint-disable-next-line @typescript-eslint/no-var-requires
       require('../fabric/NativeSvgViewModule').default;
     RNSVGSvgViewModule.toDataURL(handle, options, callback);
-  };
+  }
+
+  /**
+   * Options specific to JPEG format.
```
jakex7, line 116: > to JPEG format

Doesn't this apply to PNG as well? I think this description is redundant and could be removed.

src/elements/Svg.tsx
jakex7, line 93:
```suggestion
   * @deprecated use size property instead.
```

src/elements/Svg.tsx
```diff
@@ -81,16 +86,46 @@ export default class Svg extends Shape<SvgProps> {
...
-  toDataURL = (callback: (base64: string) => void, options?: object) => {
+  /**
+   * @deprecated use size property instead.
+   * @description Converts the SVG component to a data URL in either JPEG or PNG format.
+   * @param {function(string): void} callback - A function that receives the base64-encoded data URL string.
+   * @param {DataUrlOptions} options - The options to specify the format (JPEG or PNG) and additional parameters.
+   */
+  toDataURL(callback: (base64: string) => void, options: Size): void;
+  toDataURL(callback: (base64: string) => void): void;
+  toDataURL(callback: (base64: string) => void, options: DataUrlOptions): void;
```
jakex7, line 97:
```suggestion
  /**
   * @description Converts the SVG to a data URL image.
   * @param {function(string): void} callback - A callback function that receives the base64-encoded data URL
   * @param {DataUrlOptions} options - The options to specify the size, format (JPEG or PNG) and quality.
   */
  toDataURL(callback: (base64: string) => void, options: DataUrlOptions): void;
```

src/elements/Svg.tsx
```diff
@@ -81,16 +86,46 @@ export default class Svg extends Shape<SvgProps> {
...
 
-  toDataURL = (callback: (base64: string) => void, options?: object) => {
+  /**
+   * @deprecated use size property instead.
+   * @description Converts the SVG component to a data URL in either JPEG or PNG format.
+   * @param {function(string): void} callback - A function that receives the base64-encoded data URL string.
+   * @param {DataUrlOptions} options - The options to specify the format (JPEG or PNG) and additional parameters.
+   */
+  toDataURL(callback: (base64: string) => void, options: Size): void;
+  toDataURL(callback: (base64: string) => void): void;
```
jakex7, line 96:
```suggestion
  /**
   * @description Converts the SVG to a data URL image.
   */
  toDataURL(callback: (base64: string) => void): void;
```

PR #2434, opened 2024-08-30
docs: introduce about webpack and metro bundler

bohdanprog:
# Summary
Short explanation on how to configure the webpack bundler and a brief rationale on why it's preferable to use the Metro bundler.

## Checklist

### DIFF
--- a/USAGE.md
+++ b/USAGE.md
@@ -145,9 +145,10 @@ export default () => {
   );
 };
 ```
+# Web configuration
 
 ## Metro bundler
-You can set up your mobile and web applications using the Metro bundler, and you shouldn't face any issues using our library. 
+No additional steps are required when using Metro bundler.
 
 ### Why Use Metro and Expo for Native and Web Development?
 
@@ -163,7 +164,7 @@ Using Metro and Expo provides several advantages when developing for both native
 
 5. **Simplified Configuration**: With Metro and Expo, there’s less overhead in terms of configuration compared to Webpack. This can make the setup process faster and less error-prone, especially for developers who are not as familiar with Webpack.
 
-## Webpack bundler
+## Webpack
 > [!WARNING] `expo/webpack-config`
 > 
 > `@expo/webpack-config` is deprecated and will not receive any new feature updates.
@@ -172,7 +173,7 @@ Using Metro and Expo provides several advantages when developing for both native
 > 
 > If you are still using the Webpack bundler, we recommend migrating to Metro. This [documentation](https://docs.expo.dev/router/migrate/from-expo-webpack/) provides a good guide on how to migrate a website from Expo Webpack to Expo Metro.
 
-If, for some reason, you are using the Webpack bundler, the following steps are important:
+If you are using the Webpack bundler, the following steps are needed:
 
 - Install `@react-native/assets-registry/registry` as a project dependency.
 - Ensure that you have the proper configuration to parse flow files from `@react-native/assets-registry/registry`.
--- a/USAGE.md
+++ b/USAGE.md
@@ -146,6 +146,62 @@ export default () => {
 };
 ```
 
+## Metro bundler
+You can set up your mobile and web applications using the Metro bundler, and you shouldn't face any issues using our library. 
+
+### Why Use Metro and Expo for Native and Web Development?
+
+Using Metro and Expo provides several advantages when developing for both native and web platforms:
+
+1. **Unified Development Experience**: Metro is the default bundler for React Native, providing a consistent experience across iOS, Android, and web platforms. This eliminates the need for managing separate build configurations for different platforms.
+
+2. **Optimized Performance**: Metro is designed specifically for React Native, offering faster build times and optimized performance, especially when dealing with large codebases.
+
+3. **Seamless Integration**: Expo offers out-of-the-box integration with Metro, simplifying the development process and reducing the need for complex configuration setups, which can be particularly beneficial when targeting multiple platforms.
+
+4. **Active Maintenance**: Unlike Webpack, Metro is actively maintained and continuously improved to support the latest React Native features. This ensures better support and access to new functionalities as they are released.
+
+5. **Simplified Configuration**: With Metro and Expo, there’s less overhead in terms of configuration compared to Webpack. This can make the setup process faster and less error-prone, especially for developers who are not as familiar with Webpack.
+
+## Webpack bundler
+> [!WARNING] `expo/webpack-config`
+> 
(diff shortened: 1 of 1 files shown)


USAGE.md
```diff
@@ -146,6 +146,62 @@ export default () => {
...
+>  ...
+>};
+>
+>module.exports = {
+>  ...config,
+>  module: {
+>      rules: [babelLoaderConfiguration],
+>    },
+>};
+> ```
```
jakex7, line 201: Why is it quote?

USAGE.md
```diff
@@ -146,6 +146,62 @@ export default () => {
...
+1. **Unified Development Experience**: Metro is the default bundler for React Native, providing a consistent experience across iOS, Android, and web platforms. This eliminates the need for managing separate build configurations for different platforms.
+
+2. **Optimized Performance**: Metro is designed specifically for React Native, offering faster build times and optimized performance, especially when dealing with large codebases.
+
+3. **Seamless Integration**: Expo offers out-of-the-box integration with Metro, simplifying the development process and reducing the need for complex configuration setups, which can be particularly beneficial when targeting multiple platforms.
+
+4. **Active Maintenance**: Unlike Webpack, Metro is actively maintained and continuously improved to support the latest React Native features. This ensures better support and access to new functionalities as they are released.
+
+5. **Simplified Configuration**: With Metro and Expo, there’s less overhead in terms of configuration compared to Webpack. This can make the setup process faster and less error-prone, especially for developers who are not as familiar with Webpack.
+
```
jakex7, line 165: Do we need this section at all? It seems to be completely irrelevant to SVG library

USAGE.md
```diff
@@ -146,6 +146,62 @@ export default () => {
...
+## Webpack bundler
+> [!WARNING] `expo/webpack-config`
+> 
+> `@expo/webpack-config` is deprecated and will not receive any new feature updates.
+
+> [!TIP] Migrate from Expo Webpack
+> 
+> If you are still using the Webpack bundler, we recommend migrating to Metro. This [documentation](https://docs.expo.dev/router/migrate/from-expo-webpack/) provides a good guide on how to migrate a website from Expo Webpack to Expo Metro.
```
jakex7, line 173: Quote?
image

USAGE.md
```diff
@@ -146,6 +146,62 @@ export default () => {
 };
 ```
 
+## Metro bundler
```
jakex7, line 149: `##` makes it sub-header to 
> Use with content loaded from uri

jakex7: We could probably make a `# Web configuration` header and a
`## Metro bundler` + `## Webpack` subheaders

USAGE.md
jakex7: Also, is it needed at all?

USAGE.md
```diff
@@ -146,6 +146,62 @@ export default () => {
...
+You can set up your mobile and web applications using the Metro bundler, and you shouldn't face any issues using our library.
```
jakex7, line 150:
```suggestion
No additional steps are required when using Metro bundler.
```

USAGE.md
```diff
@@ -146,6 +146,62 @@ export default () => {
...
+## Webpack bundler
+> [!WARNING] `expo/webpack-config`
+> 
+> `@expo/webpack-config` is deprecated and will not receive any new feature updates.
+
+> [!TIP] Migrate from Expo Webpack
+> 
+> If you are still using the Webpack bundler, we recommend migrating to Metro. This [documentation](https://docs.expo.dev/router/migrate/from-expo-webpack/) provides a good guide on how to migrate a website from Expo Webpack to Expo Metro.
+
+If, for some reason, you are using the Webpack bundler, the following steps are important:
```
jakex7, line 175:
```suggestion
If you are using the Webpack bundler, the following steps are needed:
```

jakex7 (changes requested):

USAGE.md
bohdanprog: I added a short motivation on why using Metro instead of Webpack is better. We can remove that if you think it doesn't make any sense.

USAGE.md
bohdanprog: Why quote? I want to highlight that part.
In my opinion, it has to be here. However, I don't object if you believe it should be removed. I attempted to clarify it for the user and consolidate some brief tips in one location.

USAGE.md
bohdanprog: I want to highlight that part.

USAGE.md
jakex7: In that case, let's remove it completely

USAGE.md
jakex7: It's not highlighted, somehow Alerts are not rendered properly so it's displayed as a normal quote which makes it **less** visible
image

jakex7: Also, don't put code snippets in quote blocks, please

USAGE.md
jakex7: Same as here https://github.com/software-mansion/react-native-svg/pull/2434/files#r1741596785
Also, this is not relevant to react-native-svg, so I don't see the point of posting it here.

USAGE.md
bohdanprog: something like this it's better?
Screenshot 2024-09-03 at 10 07 46

USAGE.md
bohdanprog: remoed.

USAGE.md
bohdanprog: removed.

USAGE.md
```diff
@@ -145,6 +145,40 @@ export default () => {
...
+No additional steps are required when using Metro bundler.
+
+## Webpack
+If you are using the Webpack bundler, the following steps are needed:
+
+- Install `@react-native/assets-registry/registry` as a project dependency.
+- Ensure that you have the proper configuration to parse flow files from `@react-native/assets-registry/registry`.
+- Configure the Webpack `module -> rules` section and include an important rule for `node_modules/@react-native/assets-registry/registry`.
+
+#### Webpack Configuration
```
jakex7, line 160: Useless header, you can add the filename instead.
It's `babel.config.js` right? Then simply
```suggestion
babel.config.js
```

jakex7 (changes requested):

USAGE.md
bohdanprog: it's a webpack.config.js, I changed. that, thank you

jakex7 (approved):
LGTM

PR #2463, opened 2024-09-20
feat: Support `fontFeatureSettings` on web

qwertychouskie:
# Summary

Support is just a matter of making sure we pass through the value properly.  Hopefully the property is supported on iOS soon, but for now, 2 (Android + web) out of 3 is better than 1 out of 3.

## Test Plan

Ran change locally, worked fine.  It's a pretty trivial change.

An SVG with `fontFeatureSettings` set on a Text element

1. Open browser
2. See that `fontFeatureSettings` is not applied
3. Apply this change
4. See that `fontFeatureSettings` is applied properly

## Compatibility

| OS      | Implemented |
| ------- | :---------: |
| iOS     |    ❌      |
| MacOS   |    ❓      |
| Android |    ✅      |
| Web     |    ✅      |

## Checklist

### DIFF
--- a/src/web/types.ts
+++ b/src/web/types.ts
@@ -55,6 +55,7 @@ export interface BaseProps {
   fontWeight?: NumberProp;
   fontSize?: NumberProp;
   fontFamily?: string;
+  fontFeatureSettings?: string;
   forwardedRef?:
     | React.RefCallback<SVGElement>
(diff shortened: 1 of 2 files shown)


bohdanprog:
Hello @qwertychouskie,
Can you provide an example of how we can check if it works?

PR #2464, opened 2024-09-20
feat: update Gemfile.lock file, start using bundle exec pod install

bohdanprog:
# Summary
Update Gemfile with bundle install && bundle exec pod install
["Pod install" phase should run bundle exec pod install instead of pod install ](https://github.com/microsoft/appcenter/issues/2540)


fabric-macos-example/macos/Podfile.lock
```diff
@@ -969,32 +969,6 @@ PODS:
...
-  - React-RuntimeCore (0.73.34):
-    - glog
-    - RCT-Folly/Fabric (= 2022.05.16.00)
-    - React-cxxreact
-    - React-jsc
-    - React-jserrorhandler
-    - React-jsi
-    - React-jsiexecutor
-    - React-runtimeexecutor
-    - React-runtimescheduler
```
jakex7, line 997: It's **fabric**-macos-example so it should be installed with `RCT_NEW_ARCH_ENABLED` flag

PR #2467, opened 2024-09-27
chore: update iOS workflows

bohdanprog:
# Summary

Update iOS workflows.

### DIFF
--- a/.github/workflows/ios-e2e.yml
+++ b/.github/workflows/ios-e2e.yml
@@ -1,5 +1,25 @@
 name: Test iOS e2e
 on:
+  workflow_dispatch:
+    inputs:
+      logLevel:
+        description: 'Log level'
+        required: true
+        default: 'warning'
+        type: choice
+        options:
+          - info
+          - warning
+          - debug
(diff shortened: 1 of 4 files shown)


.github/workflows/ios-e2e.yml
```diff
@@ -19,7 +19,7 @@ jobs:
     timeout-minutes: 60
     env:
       WORKING_DIRECTORY: example
-      DEVICE: iPhone 14 Pro
+      DEVICE: iPhone 16 Pro
```
jakex7, line 22: In E2E tests, `iPhone 14 Pro` was chosen intentionally, as it has the exact pixel ratio needed

jakex7 (approved):
Left one comment, otherwise LGTM

.github/workflows/ios-e2e.yml
bohdanprog: Don't they have the same pixel ratio?

bohdanprog: @jakex7

PR #2541, opened 2024-11-18
feat: update SvgPackage to supported stable API's

javache:
Address some deprecated API's in SVGPackage

# Summary

This improves compatibility with future React Native releases, which will remove TurboReactPackage and the ReactModuleInfo constructor used.

## Test Plan

CI


jakex7 (approved):
Thanks for the heads-up and contribution!

PR #2542, opened 2024-11-19
fix: resolve transforms on new architecture

jakex7:
# Summary

Add `resolveTransforms` on `updateProps` to get correct matrix. It will improve animating transformations, as Animated/Reanimated skips JS `processTransform` and passes transformations directly. 

## Test Plan

Animate `transform` prop in react-native style.

## Compatibility

| OS      | Implemented |
| ------- | :---------: |
| iOS     |    ✅      |
| MacOS   |    ✅      |

### DIFF
--- a/apple/RNSVGNode.mm
+++ b/apple/RNSVGNode.mm
@@ -248,6 +248,16 @@ CGFloat const RNSVG_DEFAULT_FONT_SIZE = 12;
   }
 }
 
+- (void)setTransforms:(CGAffineTransform)transforms
+{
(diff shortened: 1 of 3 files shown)


noldidrita:
After quite a bit of searching around, I found this PR that actually fixed my issues. Thanks man! I applied it as a patch since the release is not out yet.
Any estimates on when this patch will be released?
Thanks!

jakex7:
I'm happy to hear it's working for you! 
The release will likely be early next week

PR #2562, opened 2024-12-03
refactor: use old style switch statement in FeCompositeView.java

mrtawil:
Use old-style switch statement in `FeCompositeView.java`.


jakex7 (approved):
LGTM 🚀 Thanks!

Issue #2566, opened 2024-12-05
currentColor not using correct color sometimes


believer:
### Description

We have SVG icons that use `currentColor` for strokes and fills. These are styled using [NativeWind](https://www.nativewind.dev) which sets the color on the SVG.

In the screenshots below, both SVGs display that the `color` property is defined as `rgba(171, 171, 171, 1)` in light mode and `rgba(94, 94, 94, 1)` in dark mode. But, the left icon is clearly using some other color. They should both look like the right icon.

Reloading the app solves the problem temporarily.

We don't remember that this was happening before we upgraded to the New Architecture, so there might be something related with that?

| Light | Dark |
| ------ | ----- |
| ![Screenshot 2024-12-05 at 14 25 49](https://github.com/user-attachments/assets/2a09776c-a9f3-43af-8649-e3d6a58e0ad7) | ![Screenshot 2024-12-05 at 14 24 51](https://github.com/user-attachments/assets/4f20d46d-8f22-41ee-9050-d00e985119fe) |

### Steps to reproduce

Unfortunately, we don't have a clear reproduction.

1. Use `react-native-svg-transformer` to convert imported SVGs
2. Set `color` property of imported SVGs
3. Navigate around and some icons might loose their connection to the defined color

### Snack or a link to a repository

-

### SVG version

15.10.0

### React Native version

0.76.3

### Platforms

iOS

### JavaScript runtime

Hermes

### Workflow

React Native

### Architecture

Fabric (New Architecture)

### Build type

Debug app & dev bundle

### Device

iOS simulator

### Acknowledgements

Yes

Dunedubby:
Not sure if this is the same issue, but we've also recently noticed that navigating around (particularly when doing a `navigation.reset`) causes `getCurrentColor` to completely crash out the app. When you say navigate, are you using `react-navigation`?

Dunedubby:
As clarification: we are also occasionally seeing the same issue where SVGs revert back to some default color that isn't the one specified in the component tree. It also happens pretty randomly.

believer:
Yes, we are using `react-navigation` and we've also seen crashes related to `getCurrentColor` so that could definitely be something.

jakex7:
Hey, I've never come across that bug before, so please provide a minimal reproduction (can't be random), otherwise I won't be able to debug it

believer:
Will try to create a reproduction of it next week!

gituser8796:
[...]

PR #2573, opened 2024-12-09
fix: currentColor stale value

boazarad88:
# Summary
This is another fix for issue https://github.com/software-mansion/react-native-svg/issues/2566 that also resets the color on prepareForRecycle (in the same manner of this fix https://github.com/software-mansion/react-native-svg/pull/2570). I also don't have a sandboxed reproduce, but was able to verify the fix on my company's code which I unfortunately can't share.

## Compatibility
OS | Implemented
-- | --
iOS | ✅
MacOS | ✅


jakex7 (approved):
Thank you! LGTM

PR #2577, opened 2024-12-13
feat!: Svg intrinsic size, auto and custom shadow nodes for RNSVGSvg

jakex7:
# Summary

BREAKING CHANGE / NEW ARCH ONLY

Implement intrinsic and `auto` value for `width` and `height` on `Svg` component based on the [specification](https://svgwg.org/specs/integration/#svg-css-sizing).

> Initial ‘[svg](https://svgwg.org/svg2-draft/struct.html#SVGElement)’ element [width](https://svgwg.org/svg2-draft/geometry.html#Sizing) and [height](https://svgwg.org/svg2-draft/geometry.html#Sizing) are set to 'auto' [...]
> 
> To resolve 'auto' value on ‘svg’ element if the ‘[viewBox](https://svgwg.org/svg2-draft/coords.html#ViewBoxAttribute)’ attribute is not specified:
> 
> * Use author specified width and height ‘svg’ element attributes.
> * If any of the sizing attributes are missing, resolve the missing ‘svg’ element width to '300px' and missing height to '150px' (using [CSS 2.1 replaced elements size calculation](https://www.w3.org/TR/CSS21/visudet.html#inline-replaced-width)).
> 
> To resolve 'auto' value on ‘svg’ element if the ‘viewBox’ attribute is specified:
> 
> * Use the ‘viewBox’ attribute to calculate the intrinsic aspect ratio of the ‘svg’ element.
> * If the width or height attributes are not specified - keep not specified width or height as 'auto'.

A huge thanks to @janicduplessis for #1720 with a similar concept. However, it needed to be implemented slightly different for some edge cases and the `RNSVGImage` implementation, which requires being a yoga node.

## Test Plan

Example app -> Svg examples

image

## Compatibility

| OS      | Implemented |
| ------- | :---------: |
| iOS     |    ✅     |
| MacOS   |    ✅      |
| Android |    ✅      |
| Web     |    ✅      |

## Checklist

### DIFF
--- a/android/src/main/java/com/horcrux/svg/SVGLength.java
+++ b/android/src/main/java/com/horcrux/svg/SVGLength.java
@@ -43,6 +43,9 @@ class SVGLength {
     } else if (length.codePointAt(percentIndex) == '%') {
       unit = UnitType.PERCENTAGE;
       value = Double.valueOf(length.substring(0, percentIndex));
+    } else if (length.equals("auto")) {
+      unit = UnitType.PERCENTAGE;
+      value = 100;
     } else {
       int twoLetterUnitIndex = stringLength - 2;
       if (twoLetterUnitIndex > 0) {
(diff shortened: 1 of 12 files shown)


jakex7:
Closing it for now as it's not compatible with paper and would be a huge breaking change.

PR #2640, opened 2025-02-27
Fix svgs being rendered blurry when at high dpi

acoates-ms:
# Summary
The windows svg rendering pass was always rendering the svg to a bitmap the size of the svg at 100% display scale factor, which would then be stretched at higher scale factors.  This meant that the svg would be blurry when running in scale factor systems.

With this change the svg will be rendered taking into account the scale factor.

I also removed an invalid color="none" property set, which would cause errors when running against a debug D2D instance. - That property had no effect, as it was an invalid property value. -- Instead we just leave the property unset when there is no color specified.

## Test Plan

Before:
Screenshot 2025-02-27 110318

After:
Screenshot 2025-02-27 151859

## Checklist


jakex7 (approved):
Thank you!

PR #2668, opened 2025-04-23
Remove explicit value for `usesMapBufferForStateData` template parameter

NickGerleman:
# Summary

This wasn't present in older versions of React Native, and is removed from newer versions. It was kind of meant to be hidden, with the default value used, but was fully public. Let's rely on the default `false` instead of specifying an explicit value, for compatibility of versions of RN without it.

## Test Plan

GitHub Actions building the project


jakex7 (approved):
Thank you!

Issue #2615, opened 2025-01-16
Using currentColor inside a Defs element crashes on Android


rickpastoor:
### Description

When using `currentColor` as a color on an element inside a `Defs` section the Svg component will crash.

It will not crash on iOS, although we suspect that at this moment the `currentColor` will incorrectly fall back to black, but that needs a bit more investigation. This issue is primarily about the crash.

### Steps to reproduce

1. Create a Svg component
2. Add a Defs element into it, and use currentColor inside
3. Apply the definition somewhere in your svg

### Snack or a link to a repository

https://github.com/rickpastoor/SvgAndroidCrash

### SVG version

15.11.1

### React Native version

0.76.6

### Platforms

Android

### JavaScript runtime

None

### Workflow

None

### Architecture

Fabric (New Architecture)

### Build type

None

### Device

None

### Acknowledgements

Yes

Issue #2616, opened 2025-01-16
When using currentColor in combination with a Defs pattern, the alpha channel is not picked up


rickpastoor:
### Description

It is possible to apply an alpha channel, which is correctly applied when directly using the color or when using `currentColor`. But this alpha channel seems to be completely skipped when using it inside a Defs section. Instead only the color part is used. This goes for both hex + alpha as well as rgba variants.

Note that currently this whole setup crashes Android, as I described in #2615.

![Image](https://github.com/user-attachments/assets/0527be74-d41a-437c-8c30-70fae2ea0b74)

### Steps to reproduce

I added sample elements to show which ones are correct and which ones aren't.

### Snack or a link to a repository

https://github.com/rickpastoor/SvgAndroidCrash/tree/alpha-sample

### SVG version

15.11.1

### React Native version

0.76.6

### Platforms

iOS

### JavaScript runtime

None

### Workflow

None

### Architecture

None

### Build type

None

### Device

None

### Acknowledgements

Yes

PR #2674, opened 2025-05-02
fix: update extended component for DefsView

bohdanprog:
That problem occurred only on the Android platform because we missed the correct implementation for the Defs component.

# Summary
`Defs` should extend from `RenderableView` and should have base methods. Previously, `DefsView` was extended by `DefinitionView`

Fixed issues:
#2615 
#2616 

## Test Plan

You can easily run an example application with Test2615 and Test2616, which I've created.

The app shouldn't crash.

## Compatibility

| OS      | Implemented |
| ------- | :---------: |
| Android |    ✅      |

### DIFF
--- a/android/src/main/java/com/horcrux/svg/DefsView.java
+++ b/android/src/main/java/com/horcrux/svg/DefsView.java
@@ -9,14 +9,25 @@
 package com.horcrux.svg;
 
 import android.annotation.SuppressLint;
+import android.graphics.Canvas;
+import android.graphics.Paint;
+import android.graphics.Path;
 import android.view.View;
 import com.facebook.react.bridge.ReactContext;
 
 @SuppressLint("ViewConstructor")
-class DefsView extends GroupView {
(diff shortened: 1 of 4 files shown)


android/src/main/java/com/horcrux/svg/DefsView.java
```diff
@@ -11,11 +11,12 @@
 import android.annotation.SuppressLint;
 import android.graphics.Canvas;
 import android.graphics.Paint;
+import android.graphics.Path;
 import android.view.View;
 import com.facebook.react.bridge.ReactContext;
 
 @SuppressLint("ViewConstructor")
-class DefsView extends DefinitionView {
+class DefsView extends RenderableView {
```
jakex7, line 19: DefinitionView is the exact fit for Defs; instead of changing it to Renderable, we should move the necessary functions from Renderable to Definition.

jakex7 (changes requested):

Issue #2609, opened 2025-01-13
isRecycled() on a null object reference


benawad:
### Description

I have some users that get this crash report. (Only happens on Android 10)

```
java.lang.NullPointerException: Attempt to invoke virtual method 'boolean android.graphics.Bitmap.isRecycled()' on a null object reference
    at android.graphics.BaseCanvas.throwIfCannotDraw(BaseCanvas.java:65)
    at android.graphics.BaseCanvas.drawBitmap(BaseCanvas.java:113)
    at android.graphics.Canvas.drawBitmap(Canvas.java:1560)
    at com.horcrux.svg.B.r(SourceFile:252)
    at com.horcrux.svg.B.draw(SourceFile:7)
    at com.horcrux.svg.RenderableView.render(SourceFile:44)
    at com.horcrux.svg.B.r(SourceFile:152)
    at com.horcrux.svg.B.draw(SourceFile:7)
    at com.horcrux.svg.RenderableView.render(SourceFile:44)
    at com.horcrux.svg.SvgView.drawChildren(SourceFile:149)
    at com.horcrux.svg.SvgView.drawOutput(SourceFile:69)
    at com.horcrux.svg.SvgView.onDraw(SourceFile:17)
    at android.view.View.draw(View.java:21618)
    at com.facebook.react.views.view.j.draw(SourceFile:60)
    at android.view.View.updateDisplayListIfDirty(View.java:20483)
    at android.view.View.draw(View.java:21350)
    at android.view.ViewGroup.drawChild(ViewGroup.java:4446)
    at com.facebook.react.views.view.j.drawChild(SourceFile:86)
    at android.view.ViewGroup.dispatchDraw(ViewGroup.java:4205)
    at com.facebook.react.views.view.j.dispatchDraw(SourceFile:24)
    at android.view.View.draw(View.java:21621)
    at com.facebook.react.views.view.j.draw(SourceFile:60)
    at android.view.View.updateDisplayListIfDirty(View.java:20483)
    at android.view.View.draw(View.java:21350)
    at android.view.ViewGroup.drawChild(ViewGroup.java:4446)
    at com.facebook.react.views.view.j.drawChild(SourceFile:86)
    at android.view.ViewGroup.dispatchDraw(ViewGroup.java:4205)
    at com.facebook.react.views.view.j.dispatchDraw(SourceFile:24)
    at android.view.View.draw(View.java:21621)
    at com.facebook.react.views.view.j.draw(SourceFile:60)
[...]

PR #2694, opened 2025-05-27
fix: nullPointerException in GroupView when drawing bitmap on Android 10 and older

OmalPerera:
# Summary

Open Issue: [isRecycled() on a null object reference](https://github.com/software-mansion/react-native-svg/issues/2609) 

This PR addresses a sporadic crash that occurs specifically on Android 10 (API 29) and older when rendering SVGs.

* Although the code already handles the mLayerBitmap being null in an earlier section, the bitmap can still become null on Android 10 and below due to memory pressure or race conditions, ultimately leading to a crash.

* Differences in how Bitmap memory handled and Canvas drawing operations are handled in these versions, make them more susceptible to this issue.)
[Reference: Android Bitmap Memory Management](https://developer.android.com/topic/performance/graphics/manage-memory#recycle)

## Test Plan

Unfortunately, it is bit hard to recreate this behaviour and it is very sporadic.
I have patched this change in my app, and it has produced positive results. both for me and for some other users as well.
(https://github.com/software-mansion/react-native-svg/issues/2609#issuecomment-2648320909)

-

-

## Compatibility

| OS      | Implemented |
| ------- | :---------: |
| Android |    ✅      |

## Checklist

### DIFF
--- a/android/src/main/java/com/horcrux/svg/GroupView.java
+++ b/android/src/main/java/com/horcrux/svg/GroupView.java
@@ -165,7 +165,9 @@ class GroupView extends RenderableView {
       int saveCount = canvas.save();
       canvas.setMatrix(null);
       mLayerPaint.setAlpha((int) (mOpacity * 255));
-      canvas.drawBitmap(mLayerBitmap, 0, 0, mLayerPaint);
+      if (mLayerBitmap != null) {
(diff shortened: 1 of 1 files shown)


OmalPerera:
Since applying this fix(as a patch in my project), the crash hasn’t happened again. This is the 60-day history in Android Vitals.
Screenshot 2025-07-07 at 10 32 42

PR #2725, opened 2025-07-31
chore: Add deprecation comments near separate transform props

MatiPl01:
# Summary

This PR marks separate inline transform properties in favor of the `transform` property that accepts the same format as the RN's `transform` prop.

### DIFF
--- a/src/lib/extract/types.ts
+++ b/src/lib/extract/types.ts
@@ -201,20 +201,65 @@ export type ColumnMajorTransformMatrix = [
 ];
 
 export interface TransformProps {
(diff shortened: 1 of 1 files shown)


GiyoMoon:
Hey, what should I use instead of `origin`? I often use `rotate` with `origin`.

a-klotz-p8:
> Hey, what should I use instead of `origin`? I often use `rotate` with `origin`.

Did you figure this out? Seems pretty weird to mark such a important part of transformations as deprecated without providing the replacement.

PR #2734, opened 2025-08-08
feat: Drop support for React Native <0.78

kacperzolkiewski:
# Summary

This change implements our new maintenance policy of supporting only the two most recent React Native releases. With React Native now at 0.80, the library will support only versions 0.78 and newer. Dropping older versions helps reduce maintenance overhead, remove legacy code paths and ensure we can take advantage of the latest React Native APIs without additional compatibility layers.

## Test Plan

Run test app examples

## Compatibility

| OS      | Implemented |
| ------- | :---------: |
| iOS     |    ✅      |
| MacOS   |    ✅      |
| Android |    ✅      |
| Web     |    ✅      |

## Checklist

### DIFF
--- a/common/cpp/react/renderer/components/rnsvg/RNSVGLayoutableShadowNode.cpp
+++ b/common/cpp/react/renderer/components/rnsvg/RNSVGLayoutableShadowNode.cpp
@@ -27,8 +27,10 @@ void RNSVGLayoutableShadowNode::setZeroDimensions() {
   // the Yoga layout. Setting the dimensions to 0 eliminates randomly positioned
   // views in the layout inspector when Yoga attempts to interpret SVG
   // properties like width when viewBox scale is set.
+  auto style = yogaNode_.style();
   style.setDimension(yoga::Dimension::Width, yoga::StyleSizeLength::points(0));
   style.setDimension(yoga::Dimension::Height, yoga::StyleSizeLength::points(0));
+  yogaNode_.setStyle(style);
 }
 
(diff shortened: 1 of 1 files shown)


common/cpp/react/renderer/components/rnsvg/RNSVGLayoutableShadowNode.cpp
```diff
@@ -27,18 +27,8 @@ void RNSVGLayoutableShadowNode::setZeroDimensions() {
   // the Yoga layout. Setting the dimensions to 0 eliminates randomly positioned
   // views in the layout inspector when Yoga attempts to interpret SVG
   // properties like width when viewBox scale is set.
-  auto style = yogaNode_.style();
-#if REACT_NATIVE_MINOR_VERSION >= 78
   style.setDimension(yoga::Dimension::Width, yoga::StyleSizeLength::points(0));
```
jakex7, line 30:
```suggestion
  auto style = yogaNode_.style();
  style.setDimension(yoga::Dimension::Width, yoga::StyleSizeLength::points(0));
  style.setDimension(yoga::Dimension::Height, yoga::StyleSizeLength::points(0));
  yogaNode_.setStyle(style);
```

jakex7 (changes requested):

common/cpp/react/renderer/components/rnsvg/RNSVGLayoutableShadowNode.cpp
kacperzolkiewski: fixed

jakex7 (approved):
LGTM

Issue #2746, opened 2025-08-20
Strange spike on number 2 with stroke on iOS


juhaelee:
### Description

Using Text component in the react native svg library. The number 2 has a weird spike when there is a stroke on it:

![Image](https://github.com/user-attachments/assets/d82a2466-7351-4b12-ab11-2cd139b042bf)

 

### Steps to reproduce

```
import Svg, { Text as SvgText } from 'react-native-svg';

<Svg height={size} width={size}>
<SvgText
            x="50%"
            y="50%"
            textAnchor="middle"
            alignmentBaseline="middle"
            fontSize="20"
            fontWeight="bold"
            stroke="black"
            strokeWidth="8"
            fill="black">
            22
          </SvgText>
```

### Snack or a link to a repository

https://snack.expo.dev/@junnielabs/bossy-yellow-hummus

### SVG version

15.11.2

### React Native version

0.79.5

### Platforms

iOS

### JavaScript runtime

None

### Workflow

None

### Architecture

None

### Build type

None

### Device

None

### Acknowledgements

Yes

kacperzolkiewski:
Hi @juhaelee,

A workaround for this issue is to set `strokeLinejoin="round"` or `strokeLinejoin="bevel"` as a prop, which will smooth the corner and remove the spike.
Second workaround is to change `fontFamily` for SvgText it also should help.

kacperzolkiewski:
Hey @juhaelee,

I've created PR #2748, which should resolve your issue. Could you test it and let me know if it works for you?

Henrywils0n:
@kacperzolkiewski I am still seeing this issue as of now with 15.15.1. Although Thank you for your work arounds, I was able to play around with font families and find some that have no issues.

PR #2748, opened 2025-08-20
fix: setting strokeMiterLimit prop

kacperzolkiewski:
# Summary

This PR fixes: #2746 

- Previously, the strokeMiterlimit prop had no effect on iOS because it wasn’t being applied to the drawing context.
- Applied `strokeMiterlimit` correctly by calling `CGContextSetMiterLimit`.
- Ensures the prop now works as expected when using strokeLinejoin="miter".
- Structured default behavior on all architectures so that when strokeMiterlimit is not specified, it falls back to 4.0 (matching old arch implementation).

## Test Plan

1. Run the example from issue #2746.
2. Add the `strokeMiterlimit` prop to <SvgText> (e.g. strokeMiterlimit={2}).
3. Verify that the number 2 renders without spikes when using strokeLinejoin="miter" (miter is strokeLinejoin default value).
4. Remove the prop and confirm the default value is applied (strokeMiterlimit=4).
5. Test also with strokeLinejoin="round" and "bevel" to confirm consistent behavior.

## Compatibility

| OS      | Implemented |
| ------- | :---------: |
| iOS     |    ✅      |
| MacOS   |   ✅       |
| Android |    ✅      |
| Web     |    ❌      |

## Checklist

### DIFF
--- a/android/src/main/java/com/horcrux/svg/RenderableView.java
+++ b/android/src/main/java/com/horcrux/svg/RenderableView.java
@@ -604,7 +604,7 @@ public abstract class RenderableView extends VirtualView implements ReactHitSlop
     paint.setStyle(Paint.Style.STROKE);
     paint.setStrokeCap(strokeLinecap);
     paint.setStrokeJoin(strokeLinejoin);
-    paint.setStrokeMiter(strokeMiterlimit * mScale);
+    paint.setStrokeMiter(strokeMiterlimit);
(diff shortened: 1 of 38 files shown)


lafest:
is there any updates? I also have a trouble with stroke in '2' digit.

lafest:
I applied this change to my project using patch-package, and worked well. thanks 😄

Issue #2478, opened 2024-10-08
fontData prop not working


DanyKrk:
### Description

The fontData property does not function as expected when applied to the Text component in react-native-svg. Despite being set within the component, the fontData property appears to have no discernible effect on the font rendering or characteristics of the text. 

### Steps to reproduce

1. Open the provided code in Expo Snack or in a local development environment.
2. Run the application on an Android device
3. Run the application on an iOS device

### Snack or a link to a repository

https://snack.expo.dev/Qwv-wC_hozihAeAItoL_j

### SVG version

15.3.0

### React Native version

0.73.1

### Platforms

Android, iOS

### JavaScript runtime

None

### Workflow

None

### Architecture

None

### Build type

None

### Device

None

### Acknowledgements

Yes

kacperzolkiewski:
Hi @DanyKrk,

The `fontData` property was created for internal usage within the library and was never intended to be exposed or used by external users. Its presence on the` <Text>` component does not have any effect on font rendering for external props which is why it appears non-functional.
As part of the upcoming PR this prop will be removed to avoid confusion and ensure the public API remains clean and aligned with SVG standards.

PR #2753, opened 2025-08-26
fix: deprecate unused fontData prop

kacperzolkiewski:
# Summary
Fixes: #2478 

The `fontData` property was created for internal usage within the library and was never intended to be exposed or used by external users.  PR deprecate this prop to avoid confusion and ensure the public API remains clean and aligned with SVG standards.

## Test Plan
Run example from issue: #2478, you should see a warning:
image

## Compatibility

| OS      | Implemented |
| ------- | :---------: |
| iOS     |    ✅      |
| MacOS   |    ✅      |
| Android |    ✅      |
| Web     |    ✅      |

## Checklist

### DIFF
--- a/src/lib/extract/extractText.tsx
+++ b/src/lib/extract/extractText.tsx
@@ -53,7 +53,6 @@ function parseFontString(font: string) {
 }
 
 interface fontProps {
-  fontData?: unknown;
   fontStyle?: string;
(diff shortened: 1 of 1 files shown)


jakex7 (changes requested):
Actually, we may want to deprecate it instead

Issue #2784, opened 2025-09-26
SVG onPress handling with pointerEvents Expo SDK 54


thame:
### Description

Absolutely positioned parent elements containing an SVG with onPress allowed press handling in Expo SDK 51 (react-native 0.74.5, react-native-svg 15.2.0).

When upgrading to SDK 52-54 (react-native 0.81.4, react-native-svg 15.12.1) press events are only detected for the last rendered SVG element.

A minimal example is provided. "red pressed" is logged for Expo SDK 51 but not in later versions.

### Steps to reproduce

```
import { View } from 'react-native'
import Svg, { Path } from 'react-native-svg'

export default function App() {
  return (
    <View
      style={{
        position: 'relative',
      }}
    >
      <View
        style={{ position: 'absolute', top: 0, left: 0, bottom: 0, right: 0 }}
        pointerEvents={'box-none'}
      >
        <Svg
          width={'100'}
          height={'100'}
          pointerEvents={'box-none'}
        >
          <Path
            d={'M 0 0 L 100 0 L 100 100 L 0 100 Z'}
            fill={'red'}
            onPress={() => console.log('red pressed')}
          />
        </Svg>
      </View>
      <View
        style={{ position: 'absolute', top: 0, left: 0, bottom: 0, right: 0 }}
        pointerEvents={'box-none'}
      >
        <Svg
          width={'100'}
          height={'100'}
          pointerEvents={'box-none'}
        >
          <Path
            d={'M 50 0 L 100 0 L 100 100 L 50 100 Z'}
            fill={'blue'}
            onPress={() => console.log('blue pressed')}
          />
        </Svg>
      </View>
    </View>
  )
}
```

#### Expo SDK 51
`bun install expo@51`
`bunx expo install --fix`
`bun run ios`

Tap Red shape and "red pressed" is logged
Tap Blue shape and "blue pressed is logged

#### Expo SDK 52+
`bun install expo@52`
`bunx expo install --fix`
`bun run ios`

Tap Red shape and nothing is logged
Tap Blue shape and "blue pressed is logged

### Snack or a link to a repository

https://snack.expo.dev/@thame/svg-onpress

### SVG version

15.12.1

### React Native version

0.81.4

### Platforms

iOS

### JavaScript runtime

Hermes

### Workflow

Expo Go

### Architecture

Fabric (New Architecture)

### Build type

Debug app & dev bundle

### Device

iOS simulator

### Acknowledgements

Yes

[...]

Issue #2690, opened 2025-05-22
Svg bounding box blocks touches despite parent View style={{pointerEvents: 'box-none'}} and Svg pointerEvents="none"


msageryd:
### Description

When an `<Svg>` component is nested within a standard React Native `<View style={{pointerEvents: 'box-none'}}>`, and the `<Svg>` itself has the `pointerEvents="none"` prop, its entire bounding box still blocks touch events from reaching elements rendered underneath the parent View. This blocking behavior occurs if the `<Svg>` contains any rendered child element (e.g., a purely visual `<Circle fill="orange" />`), even if that child element has no onPress handlers and no pointerEvents attributes (or `pointerEvents="none"`).

The expected behavior is that the "empty" or non-painted areas of the `<Svg>` should allow touches to pass through to the underlying elements. This would respect the `pointerEvents="box-none"` behavior of its parent View (which correctly allows pass-through for its own empty space) and the `pointerEvents="none"` on the `<Svg>` tag itself (which should make the SVG canvas non-interactive and transparent to touches).

However, it appears that the presence of any rendered child content within the `<Svg>` causes the native SVG view to register its entire bounds as an active/occupied area with React Native's gesture system. This prevents the desired pass-through effect for the SVG's empty regions, effectively making the box-none strategy on the parent View ineffective for the area covered by the Svg's bounds.

### Steps to reproduce

Create a React Native `<View>` and apply `style={{ pointerEvents: 'box-none' }}`. This View will act as a container for an overlay.

Place an `<Svg>` element as a child of this View. Set the `pointerEvents="none"` prop on the `<Svg>` element.
Inside the `<Svg>`, place a single, purely visual, non-interactive child element, for example,`<Circle fill="orange" />` (no onPress handler, no pointerEvents="auto").

Render another interactive element (e.g., a pannable View or a Button) underneath the main View (i.e., rendered before it in the component tree or with a lower zIndex).

**Minimal Code Example:**

```
import React from 'react';
import {View, Alert, Text, PanResponder, StyleSheet} from 'react-native';
import Svg, {Circle} from 'react-native-svg';

const UnderlyingInteractiveElement = () => {
  const panResponder = React.useRef(
    PanResponder.create({
      onStartShouldSetPanResponder: () => true,
      onPanResponderGrant: () => Alert.alert('Underlying Element Touched!'),
      // Add some move handlers to visually confirm panning if desired
      onPanResponderMove: (_, gestureState) => {
        console.log(
[...]

Issue #2639, opened 2025-02-27
Pointer Events on iOS seemingly broken after expo sdk 52 update


jmfp:
### Description

It seems like there is an issue with pointer events for ios and fiber (Could be wrong about it's relation to Fiber). In my project, I'm using react-native-gifted-charts to render a pie chart and then react-native-svg to essentially cut a hole in the middle of the chart. Users should be able to click on the chart (which is behind the svg) and navigate to a drilldown screen of the data it represents. 

Before upgrading to expo sdk 52, everything worked as expected, afterwards, this flow only works on android devices. After removing the svg that covers the chart, clicking is possible again, but due to stakeholder requirements the svg has to be there. I'll attach some code and screens to help illuminate the issue

here is my dependencies in my package.json
`"dependencies": {
    "@bugsnag/expo": "^52.0.0",
    "@expo-google-fonts/inter": "^0.2.3",
    "@expo/vector-icons": "^14.0.2",
    "@hookform/resolvers": "^3.10.0",
    "@react-native-community/datetimepicker": "8.2.0",
    "@react-native-community/masked-view": "^0.1.11",
    "@react-native-community/netinfo": "11.4.1",
    "@react-navigation/bottom-tabs": "^6.5.7",
    "@react-navigation/drawer": "^6.6.2",
    "@react-navigation/native": "^6.1.6",
    "@react-navigation/native-stack": "^6.9.12",
    "@reduxjs/toolkit": "^1.9.3",
    "async-mutex": "^0.4.0",
    "date-fns": "^2.30.0",
    "expo": "~52.0.36",
    "expo-application": "~6.0.2",
    "expo-constants": "~17.0.7",
    "expo-crypto": "~14.0.2",
    "expo-device": "~7.0.2",
    "expo-file-system": "~18.0.11",
    "expo-font": "~13.0.4",
    "expo-linking": "~7.0.5",
    "expo-local-authentication": "~15.0.2",
    "expo-localization": "~16.0.1",
    "expo-secure-store": "~14.0.1",
    "expo-status-bar": "~2.0.1",
    "lodash": "^4.17.21",
    "posthog-react-native": "^3.9.0",
    "react": "18.3.1",
    "react-hook-form": "^7.54.2",
    "react-native": "0.76.7",
    "react-native-calendars": "^1.1294.0",
    "react-native-gesture-handler": "~2.20.2",
    "react-native-gifted-charts": "^1.4.46",
[...]

PR #2787, opened 2025-10-01
fix: pointerEvents handling on iOS

kacperzolkiewski:
# Summary
Fixes: #2784 
Fixes: #2690
Fixes: #2639

Rename `hitTest` to `betterHitTest` for new architecture to handle properly `PointerEvents` in SVG.
Before that change we override `hitTest` from [RCTViewComponentView.mm](https://github.com/facebook/react-native/blob/3766e58b9aa50021bff508e215dcc18efb921435/packages/react-native/React/Fabric/Mounting/ComponentViews/View/RCTViewComponentView.mm#L673) which cause incorrect handling of `pointerEvents`. 
With this change we override `betterHitTest` that `RCTViewComponentView.mm` use in `hitTest` based on `pointerEvents` which fix the issue connected to pointerEvents.

## Test Plan

Run example from issue: #2784 

## Compatibility

| OS      | Implemented |
| ------- | :---------: |
| iOS     |    ✅      |
| MacOS   |    ✅      |
| Android |    ❌      |
| Web     |    ❌      |

## Checklist

### DIFF
--- a/apple/Elements/RNSVGSvgView.mm
+++ b/apple/Elements/RNSVGSvgView.mm
@@ -340,7 +340,11 @@ using namespace facebook::react;
   return CGRectContainsPoint(hitFrame, point);
 }
 
+#ifdef RCT_NEW_ARCH_ENABLED
+- (RNSVGPlatformView *)betterHitTest:(CGPoint)point withEvent:(UIEvent *)event
+#else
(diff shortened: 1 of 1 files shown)


jakex7 (approved):
LGTM

PR #2801, opened 2025-10-16
fix: web types

tjzel:
# Summary

RNSVG doesn't export types in its web implementation.

I fixed and cleanup up `ReactNativeSVG.ts` and `ReactNativeSVG.web.ts` so that the common exports are matching.

## Test Plan

`yarn tsc`


tsconfig.json
```diff
@@ -19,7 +19,9 @@
     "strict": true,
     "forceConsistentCasingInFileNames": true,
     "resolveJsonModule": true,
-    "isolatedModules": true
+    "isolatedModules": true,
+    "moduleSuffixes": [".native", ".ios", ".android", ""]
```
tjzel: This makes so that only `.ts, .tsx, .native.ts, .native.tsx` etc. are considered for type resolution.

Issue #2796, opened 2025-10-09
Path onPress prop not triggered since upgrading to expo 54


EveJulliard:
### Description

Since we update our app to expo 54, the path of my Svg are not pressable anymore.
On real android device, the onPress function is never called but surprisingly it is on android emulator. I reckon it is because user presses are not handled the same wether it is on a real device or on an emulator.

### Steps to reproduce

1. Add an SVG with paths on your app
2. Add an onPress callback on the paths
3. Presses on on path

**Expected result** : the onPress callback is triggered
**Result** : nothing append

### Snack or a link to a repository

https://snack.expo.dev/@healico/react-native-svg-path-not-pressable

### SVG version

15.13.0

### React Native version

0.81

### Platforms

Android

### JavaScript runtime

Hermes

### Workflow

Expo Dev Client

### Architecture

Fabric (New Architecture)

### Build type

Debug app & dev bundle

### Device

Real device

### Device model

Huawei P30 Lite / Google Pixel 9A

### Acknowledgements

Yes

kacperzolkiewski:
Hi @EveJulliard,

Thanks for reporting the issue!

There's indeed an issue with the new architecture when it comes to handling `onPress` events on Android devices. 
As a temporary workaround you can use `onPressIn` instead of `onPress` for your `Path` component like this:
```
        <Path
          d={rightHalf}
          fill={rightActive ? '#E74C3C' : '#F5B7B1'}
          stroke="#000"
          onPressIn={() => {
            console.log('Right half pressed');
            setRightActive(v => !v);
          }}
        />
```

EveJulliard:
Amazing, thanks @kacperzolkiewski !
Do you happen to know where this issue comes from? I tried debugging it, but I couldn’t quite figure out the root cause.

kacperzolkiewski:
@EveJulliard I think that might be some error inside React Native: https://github.com/facebook/react-native/issues/48387#issuecomment-2674162470. Because currently when I debug code inside `SVG`, `hitTest` returns proper id of `Path` component during press action.

The difference between pressing with a finger (touch) and clicking with a mouse is in how Android reports the input events.
Finger touch involves three actions:
1. ACTION_DOWN – when your finger first touches the screen
2. ACTION_MOVE – as your finger moves slightly while still touching (this action somehow breaks press event)
3. ACTION_UP – when you lift your finger off the screen

Mouse click has only two actions:
1. ACTION_DOWN – when the mouse button is pressed
[...]

Issue #2083, opened 2023-07-05
Can't render View and Text inside ForeignObject


jcubic:
# Bug

You can't use this code with the library:

```html
<ForeignObject>
  <View>
    <Text>Fe</Text>
  </View>
</ForeignObject>
```

it works in Snack but fails to render the content of View in the Android emulator.

Here are the screenshots of them:

![Przechwycenie obrazu ekranu_2023-07-05_15-27-54](https://github.com/software-mansion/react-native-svg/assets/280241/bead2831-6e22-4c73-b4ee-3901e7dce0d5)
![Przechwycenie obrazu ekranu_2023-07-05_15-27-36](https://github.com/software-mansion/react-native-svg/assets/280241/8871ca2f-7d88-4fb5-8aec-ee4634abd984)

#### Unexpected behavior

I expect it work on Android Emulator the same as on expo playground.

## Environment info

I use Pixel 6 Emulator.

```bash
info Fetching system and libraries information...
(node:1076929) Warning: Accessing non-existent property 'padLevels' of module exports inside circular dependency
(Use `node --trace-warnings ...` to show where the warning was created)
System:
    OS: Linux 5.19 Ubuntu 22.04.2 LTS 22.04.2 LTS (Jammy Jellyfish)
    CPU: (8) x64 Intel(R) Core(TM) i5-10210U CPU @ 1.60GHz
    Memory: 17.37 GB / 30.98 GB
    Shell: 5.1.16 - /bin/bash
  Binaries:
    Node: 20.3.0 - ~/.nvm/versions/node/v20.3.0/bin/node
    Yarn: 1.22.19 - ~/.nvm/versions/node/v20.3.0/bin/yarn
    npm: 9.6.7 - ~/.nvm/versions/node/v20.3.0/bin/npm
    Watchman: Not Found
  SDKs:
    Android SDK: Not Found
  IDEs:
    Android Studio: Not Found
  Languages:
    Java: 1.8.0_362 - /usr/bin/javac
  npmPackages:
    @react-native-community/cli: Not Found
    react: 18.2.0 => 18.2.0 
    react-native: 0.71.8 => 0.71.8 
  npmGlobalPackages:
    *react-native*: Not Found
info React Native v0.72.1 is now available (your project is running on v0.71.8).
info Changelog: https://github.com/facebook/react-native/releases/tag/v0.72.1
info Diff: https://react-native-community.github.io/upgrade-helper/?from=0.71.8
[...]

PR #2802, opened 2025-10-17
Fix: propagate elements metrics to their shadow nodes

kacperzolkiewski:
# Summary

Fixes: #2796 
Fixes: #2083

Previously, in the new architecture, shadow nodes for elements such as `Circle`, `Rect` and other `SVG` components were initialized with zero dimensions.
This caused issues with press interactions during move action. Yoga was unable to locate the corresponding shadow node resulting in the press event being canceled.
This PR updates the implementation to propagate each element’s actual native layout metrics to its shadow node.
By ensuring the shadow nodes reflect accurate geometry, Yoga can now correctly resolve touch targets fixing the issue where press actions were being incorrectly canceled.

## Test Plan

- Examples -> TouchEvents
- run example from issue: #2796 

## Compatibility

| OS      | Implemented |
| ------- | :---------: |
| iOS     |    ✅      |
| MacOS   |    ✅      |
| Android |    ✅      |
| Web     |    ❌      |

## Checklist

### DIFF
--- a/common/cpp/react/renderer/components/rnsvg/RNSVGComponentDescriptor.h
+++ b/common/cpp/react/renderer/components/rnsvg/RNSVGComponentDescriptor.h
@@ -14,11 +14,12 @@ public:
     react_native_assert(dynamic_cast<RNSVGLayoutableShadowNode *>(&shadowNode));
 
     auto layoutableShadowNode = dynamic_cast<RNSVGLayoutableShadowNode *>(&shadowNode);    
-    auto state = layoutableShadowNode->getStateData();
+    auto state = std::static_pointer_cast<const typename ShadowNodeT::ConcreteState>(shadowNode.getState());
+    auto stateData = state->getData();
 
-    if (state.getWidth() != 0 && state.getHeight() != 0) {
-      layoutableShadowNode->setSize(Size{state.getWidth(), state.getHeight()});
-      layoutableShadowNode->setShadowNodePosition(state.getX(), state.getY());
+    if (stateData.getWidth() != 0 && stateData.getHeight() != 0) {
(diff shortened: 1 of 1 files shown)


apple/RNSVGRenderable.h
```diff
@@ -36,6 +42,10 @@
 @property (nonatomic, assign) CGPathRef hitArea;
 @property (nonatomic, strong) NSString *filter;
 
+#ifdef RCT_NEW_ARCH_ENABLED
+- (facebook::react::RNSVGRenderableShadowNode::ConcreteState::Shared)state;
+#endif
+
```
jakex7: It's only internal property, we don't need to define it here

common/cpp/react/renderer/components/rnsvg/RNSVGComponentDescriptor.h
```diff
@@ -0,0 +1,30 @@
...
+#pragma once
+#include <react/debug/react_native_assert.h>
+#include <react/renderer/core/ConcreteComponentDescriptor.h>
+#include "RNSVGShadowNodes.h"
+
+namespace facebook::react {
+
+template <typename ShadowNodeT>
+class RNSVGComponentDescriptor final
+: public ConcreteComponentDescriptor<ShadowNodeT> {
```
jakex7:
```suggestion
class RNSVGComponentDescriptor final : public ConcreteComponentDescriptor<ShadowNodeT> {
```

apple/RNSVGRenderable.h
kacperzolkiewski: It is also used inside `RNSVGGroup.mm`

PR #2850, opened 2026-01-15
fix: ios breaking changes for RCTImage use-after-free crash

ryanfrawley:
# Summary

Adopts the [breaking change bug fix](https://github.com/facebook/react-native/commit/3b46755b3b2f4a53b79d0828ef4c0cfe319a34ac) in React Native RCTImage for iOS to fix a rare use-after-free race condition in the image progress observer. (See React Native PR for more details)

## Test Plan

Rerun nightly test suite and verify no more breakages.

## Compatibility

| OS      | Implemented |
| ------- | :---------: |
| iOS     |    ✅       |
| MacOS   |    n/a      |
| Android |    n/a      |
| Web     |    n/a      |

## Checklist

### DIFF
--- a/apple/Elements/RNSVGImage.mm
+++ b/apple/Elements/RNSVGImage.mm
@@ -27,6 +27,8 @@
 #import "RNSVGViewBox.h"
 
 #ifdef RCT_NEW_ARCH_ENABLED
+#import <memory>
+
 #import <React/RCTConversions.h>
 #import <React/RCTFabricComponentsPlugins.h>
 #import <React/RCTImageResponseObserverProxy.h>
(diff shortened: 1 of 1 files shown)


ryanfrawley:
> Thank you for the PR! Just to make sure, this isn't breaking change for `react-native-svg` and will still work with older `react-native` versions, right?

That's correct. Only newer react-native versions would be incompatible.

vokhuyetOz:
`/node_modules/react-native-svg/apple/Elements/RNSVGImage.mm:125:40 No viable conversion from 'facebook::react::RCTImageResponseObserverProxy' to 'const std::shared_ptr<const ImageResponseObserver>'
`
got this build error with `react-native-svg@15.15.3`

PR #2944, opened 2026-05-06
Add RNRepo to ci builds

radoslawrolka:
# Summary

Adds RNRepo build-tools integration to `fabric-example` so that CI builds can consume pre-built native dependencies instead of compiling them from scratch, speeding up CI.

The integration is guarded by `CI=true` / `ENV['CI']` environment variables, so local development is unaffected. `react-native-svg` itself is excluded from prebuilds via `rnrepo.config.json` so that CI always builds and tests the library from source.

## Test Plan

Builds in CI using RNRepo should speed up the CI process. Android sees ~38% improvement across the board. iOS builds are ~15% faster.

- https://github.com/software-mansion/react-native-svg/actions/runs/25438112937: 5min 
- without: 6-12min

## Compatibility

| OS      | Implemented |
| ------- | :---------: |
| iOS     |    ✅      |
| MacOS   |    ❌      |
| Android |    ✅      |
| Web     |    ❌      |

## Checklist

### DIFF
--- a/.github/workflows/android-build-test.yml
+++ b/.github/workflows/android-build-test.yml
@@ -68,4 +68,4 @@ jobs:
 
       - name: Build app
         working-directory: apps/${{ matrix.working-directory }}/android
-        run: ./gradlew assembleDebug --build-cache --console=plain -PreactNativeArchitectures=arm64-v8a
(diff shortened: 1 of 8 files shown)


jakex7 (approved):
LGTM, thank you!

PR #2995, opened 2026-07-11
fix: don't clobber inherited fill on stroke-only shapes (emit fill only when set)

FaithfulAudio:
## Problem
A stroke-only element (e.g. `<Circle stroke="..." strokeWidth={2}/>` with no `fill` prop) under a root `<Svg fill="none">` renders with a solid **black** fill on react-native-windows instead of being unfilled.

## Root cause (JS side)
`extractFill` always emits the black spec-default `fill` for any element without an explicit `fill`, but does **not** add `'fill'` to `propList`. `propList` drives inherited-property resolution natively; because `'fill'` isn't in it, the element should inherit fill from its parent. On renderers that apply the emitted `o.fill` regardless of `propList`, the stray black default overrides the inherited `none`.

## Fix (this PR)
Only emit `o.fill` when `fill` is actually set. When unset, emit nothing and keep `'fill'` out of `propList` so the native side inherits. A genuinely-unset root still defaults to black via the native node's own default fill, preserving spec behavior. Behavior-neutral on iOS/Android (they already override the stray default via inheritance).

## Validation — including an important NEGATIVE result on react-native-windows
Applied to react-native-svg **15.15.3** in a production RNW 0.83.2 new-arch app (Facilitron FIT, Windows 11 ARM64, 250% scale, RNSVG native built from source) and verified live at three levels (patched source on disk, Metro-served transform, and in-app introspection of the running Hermes instance).

**Finding: this JS fix alone does NOT cure the symptom on react-native-windows.**

| before (unpatched) | after (JS patch verified live) |
|---|---|
| before | after — still black |

Mechanism (verified in `windows/RNSVG/RenderableView.cpp`, Fabric `UpdateProperties`, ~line 217): when `fill` is absent and `'fill'` is not in `propList`, the code falls back to `parent.Fill()` with a Black default — and the root `<Svg fill="none">` value doesn't reach the child through that path. So the complete cure is two-part: this JS hygiene change **plus** a `windows/` inheritance fix (we're happy to follow up with the native patch if maintainers want it here or as a separate PR).

We're aware a Jest snapshot or two asserting the explicit black `fill` on unset elements may need updating — will adjust on review feedback.

### DIFF
--- a/src/lib/extract/extractFill.ts
+++ b/src/lib/extract/extractFill.ts
@@ -20,10 +20,15 @@ export default function extractFill(
     inherited.push('fill');
     o.fill =
       !fill && typeof fill !== 'number' ? defaultFill : extractBrush(fill);
-  } else {
-    // we want the default value of fill to be black to match the spec
-    o.fill = defaultFill;
   }
+  // When `fill` is not set on this element we deliberately emit no `fill` value
+  // and do not add it to `propList`, so the native side inherits fill from the
+  // parent (e.g. a root <Svg fill="none">). Previously we emitted the black
+  // spec-default here even though 'fill' was absent from propList; on platforms
+  // that apply the emitted value unconditionally (react-native-windows) this
+  // clobbered inheritance and painted a black fill on stroke-only shapes. The
+  // SVG black default is still honored for a genuinely unset root because the
+  // native node's own default fill is black.
   if (fillOpacity != null) {
     inherited.push('fillOpacity');
     o.fillOpacity = extractOpacity(fillOpacity);
--- a/windows/RNSVG/Fabric/RenderableView.cpp
+++ b/windows/RNSVG/Fabric/RenderableView.cpp
@@ -158,6 +158,49 @@ void RenderableView::FinalizeUpates(
 ID2D1SvgElement &RenderableView::Render(const SvgView &svgView, ID2D1SvgDocument& document, ID2D1SvgElement &svgElement) noexcept {
   svgElement.CreateChild(GetSvgElementName(), m_spD2DSvgElement.put());
   OnRender(svgView, document, *m_spD2DSvgElement);
+
+  // extractFill.ts (JS) intentionally leaves `fill` unset and out of `propList` when an
+  // element never declares `fill`, so that native inherits it from the parent (e.g. a root
+  // <Svg fill="none">). SetCommonSvgProps only calls SetAttributeValue for attributes present
+  // in propList, so OnRender above leaves this freshly created D2D SVG element's `fill`
+  // attribute completely unset in that case. Because this tree is rebuilt from scratch via
+  // CreateChild()/SetAttributeValue() every frame (SvgView::Draw creates a new
+  // ID2D1SvgDocument on every Invalidate) rather than parsed from a full <svg> document, D2D
+  // does not retroactively cascade an unset presentation attribute from its parent - it
+  // resolves to the SVG spec's black initial paint instead, reproducing the exact bug the JS
+  // fix was meant to eliminate. Direct2D's paint model (D2D1_SVG_PAINT_TYPE) has no INHERIT
+  // value, so setting the literal string "inherit" via SetAttributeValue is not meaningful for
+  // `fill`/`stroke` - we resolve inheritance ourselves instead, by copying the parent
+  // element's already-resolved `fill` paint onto this element. Parent nodes are always
+  // rendered (and thus resolved by this same logic) before their children in
+  // SvgView::RecurseRenderNode, so by induction this correctly propagates the nearest
+  // explicit ancestor's fill all the way down, and bottoms out at the SVG spec's black
+  // default when no ancestor - including the document root, which never has `fill` set (see
+  // SvgView::Draw) - ever declared one either.
+  if (!m_spD2DSvgElement->IsAttributeSpecified(SvgStrings::fillAttributeName, nullptr)) {
+    winrt::com_ptr<ID2D1SvgPaint> parentPaint;
+    winrt::com_ptr<ID2D1SvgPaint> childPaint;
+    if (SUCCEEDED(svgElement.GetAttributeValue(SvgStrings::fillAttributeName, parentPaint.put())) && parentPaint &&
+        SUCCEEDED(m_spD2DSvgElement->GetAttributeValue(SvgStrings::fillAttributeName, childPaint.put())) && childPaint) {
+      D2D1_SVG_PAINT_TYPE paintType = parentPaint->GetPaintType();
+      childPaint->SetPaintType(paintType);
+      if (paintType == D2D1_SVG_PAINT_TYPE_COLOR) {
+        D2D1_COLOR_F color;
+        parentPaint->GetColor(&color);
+        childPaint->SetColor(color);
+      } else if (
+          paintType == D2D1_SVG_PAINT_TYPE_URI || paintType == D2D1_SVG_PAINT_TYPE_URI_NONE ||
+          paintType == D2D1_SVG_PAINT_TYPE_URI_COLOR || paintType == D2D1_SVG_PAINT_TYPE_URI_CURRENT_COLOR) {
+        UINT32 idLen = parentPaint->GetIdLength();
+        if (idLen > 0) {
+          std::wstring id(idLen, L'\0');
+          parentPaint->GetId(&id[0], idLen + 1);
+          childPaint->SetId(id.c_str());
+        }
+      }
+    }
+  }
+
   return *m_spD2DSvgElement;
 }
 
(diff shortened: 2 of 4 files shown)


FaithfulAudio:
## Part 2: react-native-windows companion fix — now validated end-to-end (ring, not black disk)

Following up on the negative result noted earlier (the JS-only change in this PR was
correct but left the black disk on RNW): I've now written and **validated on-screen** the
companion native fix. A stroke-only `<Circle>` under `<Svg fill="none">` renders as an
unfilled **ring** on react-native-windows, matching an explicit `fill="none"` control,
with no regression on normal filled SVGs.

Root-causing this on RNW turned up **two** distinct native gaps, both in
`windows/RNSVG/Fabric/`. Each is necessary; neither alone fixes the bug.

### Background: how RNW renders SVG

RNW's Fabric renderer doesn't keep a persistent scene graph — on every invalidation
`SvgView::Draw` creates a brand-new `ID2D1SvgDocument` and rebuilds the whole tree from
scratch with `ID2D1SvgElement::CreateChild()` + `SetAttributeValue()`. Presentation
attributes are only pushed onto a child when they appear in that element's `propList`
(`SetCommonSvgProps` in `RenderableView.h`). This PR's JS change deliberately drops `fill`
from `propList` when an element never declared it, intending native to inherit fill from
the parent.

### Gap 1 — the root `<Svg fill="none">` never reached the D2D tree at all

`SvgView::Draw` recursed into the SvgView's own direct children via `RecurseRenderNode`
**without ever calling their `Render()`/`OnRender()`**:

```cpp
for (auto const &child : view.Children()) {
  auto renderable = child.UserData().as<RenderableView>();
  if (renderable->IsSupported()) {
    RecurseRenderNode(this, child, *spSvgDocument, *spRoot);  // child's own element never created
  }
}
```

`Svg.tsx` wraps every `<Svg>`'s children in a `<G>` that carries `<Svg>`'s own presentation
props — including `fill="none"`. Because that wrapper `<G>`'s own D2D `<g>` element was
never materialized, the root `fill="none"` was dropped on the floor: there was nothing in
the D2D SVG DOM for any descendant to inherit `none` from. Fix — call `Render()` for each
direct child before recursing, exactly as `RecurseRenderNode` does at every other level:

```cpp
for (auto const &child : view.Children()) {
  auto renderable = child.UserData().try_as<RenderableView>();
  if (renderable && renderable->IsSupported()) {
    ID2D1SvgElement &newElement = renderable->Render(*this, *spSvgDocument, *spRoot);
    RecurseRenderNode(this, child, *spSvgDocument, newElement);
  }
}
```

### Gap 2 — D2D doesn't cascade an *unset* presentation attribute onto a child

Even with the root fill now in the tree, an element whose `fill` is left unset (because
`fill` isn't in its `propList`) resolves to the SVG spec's **black initial value**, not the
parent's computed value — because the tree is built imperatively per-frame, not parsed from
a document, so D2D never runs the cascade for it. Direct2D's paint model
(`D2D1_SVG_PAINT_TYPE`) has **no `INHERIT` member**, so setting the literal string
`"inherit"` via `SetAttributeValue` is a no-op for paint attributes (I tried it first — the
disk stayed black). Fix — resolve inheritance ourselves in `RenderableView::Render()`, after
`OnRender()` populates the element: if `fill` is unspecified, copy the **parent** element's
already-resolved `ID2D1SvgPaint` onto this element:

```cpp
if (!m_spD2DSvgElement->IsAttributeSpecified(SvgStrings::fillAttributeName, nullptr)) {
  winrt::com_ptr<ID2D1SvgPaint> parentPaint, childPaint;
  if (SUCCEEDED(svgElement.GetAttributeValue(SvgStrings::fillAttributeName, parentPaint.put())) && parentPaint &&
      SUCCEEDED(m_spD2DSvgElement->GetAttributeValue(SvgStrings::fillAttributeName, childPaint.put())) && childPaint) {
    D2D1_SVG_PAINT_TYPE paintType = parentPaint->GetPaintType();
    childPaint->SetPaintType(paintType);
    if (paintType == D2D1_SVG_PAINT_TYPE_COLOR) {
      D2D1_COLOR_F color; parentPaint->GetColor(&color); childPaint->SetColor(color);
    } else if (paintType == D2D1_SVG_PAINT_TYPE_URI || ... /* URI_* variants */) {
      UINT32 idLen = parentPaint->GetIdLength();
      if (idLen > 0) { std::wstring id(idLen, L'\0'); parentPaint->GetId(&id[0], idLen + 1); childPaint->SetId(id.c_str()); }
    }
  }
}
```

Parents are always rendered before their children (`RecurseRenderNode`), so by induction
this propagates the nearest explicit ancestor's fill all the way down, and correctly bottoms
out at the SVG spec's black default when no ancestor — including the document root, which
never has `fill` set — declared one either.

### Validation (real app, on-screen)

Built RNSVG from source into a production RNW 0.83.2 new-arch app (Windows 11 ARM64, Debug,
250% display scale) with this PR's JS fix + both native fixes:

| | no `fill` prop (bug) | explicit `fill="none"` (control) |
|---|---|---|
| **Before** (JS fix only) | solid black disk | ring |
| **After** (JS + both native fixes) | **ring** | ring |

- `evidence/rnsvg-fill-clobber-AFTER-native-fix-250pct.png` — the no-fill circle now renders
  as an unfilled ring, matching the control.
- `evidence/rnsvg-fill-clobber-AFTER-native-fix-regression-dashboard-250pct.png` — app logo
  and header/tab-bar vector icons (normal filled SVGs) unaffected; no regression.

### Files changed (native, this part)

- `windows/RNSVG/Fabric/SvgView.cpp` — render the SvgView's direct children (so the root
  `<Svg>`'s own fill reaches the D2D tree).
- `windows/RNSVG/Fabric/RenderableView.cpp` — copy the parent's resolved paint onto a child
  whose `fill` is unset (manual inheritance; D2D has no `INHERIT` paint type).

Happy to split the native part into its own PR against `windows/` if that's easier to
review, or keep it as part 2 here. The JS change stands on its own for iOS/Android; this is
purely additive for RNW.


---
**Evidence (production app, 250% scale, ARM64):**

| no `fill` prop (was black disk) | explicit `fill="none"` (control) |
|---|---|
| after | (matches) |

Regression check — filled SVGs unaffected: regression

_Companion native commit pushed to this branch: `b2ea9df`._

FaithfulAudio:
⚠️ **Correction / stability regression — please do not act on the native part of my previous comment yet.**

Further testing surfaced a crash caused by the `SvgView::Draw` change (gap 1 — rendering direct children). On a **mount** of a fresh SVG tree (in our app: a responsive layout switch that remounts the view), that render path runs before a `GroupView`'s state is initialized, and `SetCommonSvgProps` dereferences a null `std::optional<Color>` → access violation. Stack:

```
optional<Color>::has_value  (null deref)
operator==<Color>
SetCommonSvgProps<SvgGroupProps>
GroupView::OnRender
RenderableView::Render
SvgView::Draw            <- the gap-1 change
SvgView::OnMounted
```

So my "validated end-to-end" claim was premature: I confirmed the *ring renders* but did not test *stability under mount/resize*. The rendering result stands; the native change as written is **not safe** at mount time. I'm testing a null-guarded version now and will follow up with either a corrected native patch (re-validated for both the ring **and** no crash) or a retraction of the native part (leaving the JS-only fix in this PR, which is correct and safe for iOS/Android regardless). Apologies for the noise — flagging promptly so nobody builds on the unsafe version.

FaithfulAudio:
✅ **Resolved — the stability regression is fixed (forward, not reverted); the ring is intact.**

Root cause of the crash I flagged above: the child-render in `SvgView::Draw` can execute during the `onMounted` cascade **before** Fabric's prop-update pass reaches a freshly-mounted node, so `RenderableView::m_props` is still null and `winrt::get_self<SvgGroupProps>(m_props)` yields an invalid `this` (the +0x1d0 offset in the AV = multiple-inheritance adjustment on a null interface) → `SetCommonSvgProps`'s `Color` compare dereferences it. It reproduced not only on a responsive layout remount but on some cold launches too — any fresh SVG mount races the prop pass.

Fix (pushed to this branch): added `RenderableView::HasProps()` and gated both render sites in `SvgView.cpp` on it, mirroring the existing `IsSupported()` guard. A not-yet-propped node is simply skipped for that `Draw` pass; the in-flight prop update's own `Invalidate()` redraws it a moment later with props present. Arguably this is a pre-existing latent mount-ordering issue that the child-render change merely surfaced — the guard is defensive either way.

**Re-validated in the production app** (RNW 0.83.2, ARM64, 250% scale): 4+ resize cycles across the mobile↔desktop breakpoint plus SVG-heavy screen navigation, same process survives every remount, zero new crash dumps, and the stroke-only circle still renders as an unfilled ring. So the two-part fix (JS `extractFill` + native inheritance + this mount guard) is now stable end-to-end. Happy to squash the three native commits if you'd prefer a cleaner history for review.

PR #3012, opened 2026-08-13
refactor: inline the Touchable press state machine

zoontek:
# Summary

`Shape` handles presses by mixing in `Touchable.Mixin`, a private and deprecated API of `react-native`. react-native cannot delete `Touchable` while this library uses it.

From `packages/react-native/index.js` at `v0.87.0`:

```js
// `Touchable` has been removed from the public API types, but remains
// re-exported at runtime here because of a hanging `react-native-svg` call
// site (fbsource).
// TODO(huntie): Remove this re-export once `react-native-svg` is updated.
```

## How it is implemented

`src/lib/SvgTouchableMixin.ts` now holds its own copy of the press state machine, ported from `Touchable.Mixin`. Same states, same signals, same transition table, same method names, so the callers are unchanged. It only depends on the gesture responder system, which is public API.

Two differences with the original implementation:

- No Android tap sound on press. `SoundManager` has no public export.
- Measurement calls `measure()` on the granted responder node, falling back to the instance `root` ref, instead of `UIManager.measure`.

The mixin is also fully typed now, so the `// @ts-expect-error: Mixin is not typed` is gone.

## Areas impacted

Press props on every shape element: `onPress`, `onPressIn`, `onPressOut`, `onLongPress`, `delayPressIn`, `delayPressOut`, `delayLongPress`, `pressRetentionOffset`, `hitSlop`, `disabled`, and the user supplied `onResponder*` overrides.

Plus the config needed to type check the new test file:

- `tsconfig.json` includes `__tests__` and `e2e`, and lists `types` explicitly
- new `tsconfig.build.json`, narrowed back to `src`, used by the `bob` typescript target so test globals never reach the published types
- devDependency bumps: `@types/jest`, new `@types/react-test-renderer`, `react-native-builder-bob`

`yarn install` at the repo root, then `yarn install` in the example app.

`__tests__/touchable.test.tsx` is new. Its 19 cases cover handler attachment, press order, `delayPressIn` / `delayPressOut` / `delayLongPress`, the 500ms long press threshold, long press cancelling `onPress`, the 10px movement cancel, leaving and re-entering the press rect, the default retention offset, `hitSlop`, `disabled`, user supplied responder handlers, termination, and timer cleanup on unmount.

```sh
yarn jest __tests__/touchable.test.tsx
yarn lint
yarn tsc
```

Then check the `TouchEvents` example screen, which covers `onPress` on `Circle`, `onLongPress` on `Rect`, `onPressIn` / `onPressOut` with `delayPressIn={0}`, and `onPress` on `G` and `Text`:

```sh
cd apps/fabric-example && yarn ios # or yarn android
```

## Compatibility

| OS      | Implemented |
| ------- | :---------: |
| iOS     |     ✅      |
| MacOS   |     ✅      |
| Android |     ✅      |

The change is JavaScript only and shared by every platform. No native code is touched.

## Checklist

## Demos

https://github.com/user-attachments/assets/0000deed-32d4-4fe3-bca0-00c005633bb8

https://github.com/user-attachments/assets/a0798cec-6261-4215-9c30-d40a19588c59

### DIFF
--- a/__tests__/touchable.test.tsx
+++ b/__tests__/touchable.test.tsx
@@ -1,0 +1,363 @@
+import renderer from 'react-test-renderer';
+import Svg, { Rect } from '../src';
+
+// Fake native ref: measure() reports a 100x100 view at page offset (0, 0).
+const createTarget = () => ({
+  measure: (
+    callback: (
+      x: number,
+      y: number,
+      width: number,
+      height: number,
+      pageX: number,
+      pageY: number
+    ) => void
+  ) => callback(0, 0, 100, 100, 0, 0), // eslint-disable-line n/no-callback-literal
+});
+
+const createTouchEvent = (target: unknown, pageX = 50, pageY = 50) => ({
+  persist: jest.fn(),
+  currentTarget: target,
+  nativeEvent: {
+    pageX,
+    pageY,
+    locationX: pageX,
+    locationY: pageY,
+    identifier: 1,
+    target,
+    timestamp: 100,
+    touches: [],
+    changedTouches: [],
+  },
+});
+
+/* eslint-disable @typescript-eslint/no-explicit-any */
+type Handlers = {
+  onStartShouldSetResponder: (e: any) => boolean;
+  onResponderGrant: (e: any) => void;
+  onResponderMove: (e: any) => void;
+  onResponderRelease: (e: any) => void;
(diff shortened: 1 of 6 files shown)


jakex7 (approved):
Looks good to me, thank you! I'll re-introduce Android tap sound on press in a followup PR

PR #3024, opened 2026-08-29
fix: honor an explicit translateX/Y of 0 over the deprecated x/y props

hdimer:
# Summary

`x`/`y` are documented in `src/lib/extract/types.ts` as deprecated aliases for `translateX`/`translateY`:

```ts
/**
 * @deprecated Use translateX in transform prop instead.
 */
x?: NumberArray;
```

So when both are present the explicit `translateX` should win. It doesn't when it is `0`:

```ts
const tr = universal2axis(
  translate,
  translateX || (Array.isArray(x) ? x[0] : x),
  translateY || (Array.isArray(y) ? y[0] : y)
);
```

`||` treats `0` as falsy, so `translateX={0}` is silently dropped and the deprecated `x` prop (or the `translate` array) is used instead. `<G x={100} translateX={0} />` renders translated by 100 rather than by 0. It bites hardest when the value is animated or computed and passes through zero, because the element jumps back to the deprecated coordinate for those frames.

`??` lets an explicit `0` through while `undefined` still falls back to the alias. `universal2axis` already distinguishes the two downstream (`axisX = +axisX; if (!isNaN(axisX))`), so the caller was the only thing destroying the zero. Same shape as `SvgTouchableMixin.ts:364`, where `??` is used for the same reason.

Two things I deliberately left alone, so they don't look like oversights:

- The same falsy-zero pattern exists in `universal2axis`'s own `return [x || defaultValue || 0, ...]`, which turns `scaleX={0}` into `scale(1)` (and the web path does the same at `parseTransform.ts:49`). That is a wider behavioural change with a real blast radius, since scale 0 gives a singular matrix and apps may lean on the current default, so it belongs in its own PR if you want it at all.
- Any *non-numeric* `translateX` (e.g. `""`) now also wins over the deprecated `x` and resolves to `0`, where before it fell back. That is only reachable from JSX, needs both props on one element, and the old behaviour worked only by resurrecting a `@deprecated` alias. I did not pin it in a test, so you keep the option to warn or throw there later.

No issue tagged: I hit this reading the transform extraction code rather than from a report, and I couldn't find an existing one. Happy to open an issue first if you'd rather discuss it before a patch.

## Test Plan

Added `__tests__/extractTransform.test.ts`. Two of the tests fail on `main` and pass with the fix; the third passes on both and guards the fallback, so `x`/`y` keep working when `translateX`/`translateY` are absent (it is what goes red if someone later "simplifies" the fix by dropping the alias).

Nothing beyond `yarn install`. `props2transform` is pure TypeScript with no React Native runtime dependency, and it was already exported.

```ts
import { props2transform } from './src/lib/extract/extractTransform';

props2transform({ x: 10, translateX: 0 }).x; // main: 10, expected: 0
```

`yarn lint`, `yarn tsc` and `yarn bob` are green.

On the new test file: `yarn test` is `lint && tsc`, so jest doesn't run it in CI, and running `yarn jest` directly blocks, because `jest.config.ts` points `globalSetup` at `e2e/setupJest.ts`, which waits on the E2E WebSocket handshake for a device. I ran the file through a local config without that hook (3 passed), and confirmed the first two go red when the two `??` are reverted. `__tests__/css.test.tsx` and `touchable.test.tsx` are in the same position, so I've left the config alone rather than change it in a bug-fix PR.

## Compatibility

| OS      | Implemented |
| ------- | :---------: |
| iOS     |     ✅      |
| MacOS   |     ✅      |
| Android |     ✅      |
| Web     |     ➖      |

Web is unaffected rather than fixed: `src/web/utils/parseTransform.ts` builds its transform string from `translateX`/`translateY` only and never consults `x`/`y`, so this precedence never applied there. Everything else shares this JS path, and the matrix is computed in JS before it reaches native.

## Checklist

- [X] I added a test for the API in the `__tests__` folder

Not device-tested, and no `tests-example` case: this is a pure JS resolution change with unit coverage, and no types or documented API changed. Glad to add a `TestX.tsx` and before/after screenshots if you'd like them.

---

Disclosure: I used an AI assistant while working on this. The bug, the fix and the tests were checked by hand against the repo's own toolchain before submitting.
---
Disclosure: this change was prepared with AI assistance (Claude Code). The repro and tests described above were run before it was opened.


alanAUguy13 (commented):
Run

PR #3025, opened 2026-08-30
Ignore stale SVG URI responses

OskarEichler:
# Summary

`SvgUri` and `SvgFromUri` allow an older slow request to settle after a newer URI and overwrite the already-rendered SVG. The functional component can also fire stale `onLoad`/`onError` callbacks or replace successful output with its fallback, while the class component can update after unmount.

Track each functional effect through its cleanup flag and each class request through a monotonically increasing ID. Only the current request may update state, report callbacks/errors, or render a fallback. The functional effect remains URI-driven only, preserving the existing behavior when callers pass new inline callback identities.

### What is required?

The existing Jest/react-test-renderer environment; network requests are represented by controllable mocked `fetchText` promises.

### Steps and results

- Added stale-success regressions for both `SvgUri` and `SvgFromUri`; exact baseline renders width 111 after the old request wins, while the fix retains the newer width 222.
- Added functional regressions proving stale success cannot call the old `onLoad` and stale rejection cannot call either `onError` or replace newer output with the fallback.
- Focused regressions pass 4/4.
- `yarn test`, `yarn bob`, `yarn check-archs-consistency`, and `git diff --check` pass.
- The pre-existing CSS snapshot suite remains independently stale and is unrelated to this patch.

## Compatibility

| OS      | Implemented |
| ------- | :---------: |
| iOS     |      ✅     |
| MacOS   |      ✅     |
| Android |      ✅     |
| Web     |      ✅     |

## Checklist

- [ ] I have tested this on a device and a simulator — JS lifecycle change covered by deterministic renderer tests
- [ ] I added documentation in `README.md` — no public API or documentation change
- [x] I updated the typed files (TypeScript source implementation)
- [x] I added regression tests in the `__tests__` folder

## Breaking changes

None. Only stale results and callbacks from superseded/unmounted requests are ignored.


OskarEichler:
A second pass found one callback edge case in the hook implementation: removing callbacks from the request effect prevented refetches, but a same-URI prop update left an in-flight request calling the old callback. I aligned this with the CSS URI implementation by keeping current callbacks in refs, added a no-refetch/latest-callback regression, and fixed the deferred helper to satisfy the repository lint rule. Focused Jest passes (5 tests), and targeted ESLint/Prettier checks pass.

